PR #22, opened 2025-06-04
[WIP] - Abstract client server experimentation

pierre-rouanet:
### TODO
- [x] Decide of the default io implementation
     - [x] Add basic client/server io unit testing.
- [x] Zenoh
    - [x] restrict zenoh discovery on localhost by default
    - [x] Add zenoh as a dependency (optional or not)


pierre-rouanet:
@apirrone @cdussieux I'm tempted to set zenoh as the default client/server io. 

- I would say it's not really useful for our current version but could be interesting for the next steps (handling multiple robots, having a dashboard, connection status, router scouting, etc.)
- If we want to use it later, it's probably a good idea to already start using it in more realistic environments and test it more deeply.
- It's only two lines of code to change so we can easily revert this later if we want to.

Sounds goods to you? Feedback?

apirrone:
I'm all for it ! The sooner we switch to the "definitive" client/server io, the sonner we can build the other features on top if it

apirrone (approved):
Tried with mujoco server, real robot, both, client on another computer, localhost_only=True/False 

All works !

zeapsu:
Workflow established and in use (shipped #23); durable steps live in AGENTS.md ("reachy_controller UI gotchas") + agent memory. Closing as a completed reference.

PR #23, opened 2025-06-04
[WIP] 6 handle camera in mujoco

apirrone:


FabienDanieau:
Use `python examples/video_client.py `to see the video stream
Nice to launch `python examples/client_example.py` to see some mouvement

PR #32, opened 2025-06-05
Objects can move around in the mujoco minimal scene

apirrone:
enabling freejoints for objects, adding pure shape collision for the tabletop, enabling multi contact collision detection in mujoco


cdussieux:
lgtm

PR #39, opened 2025-06-05
bug #38: move image grabbing to thread

FabienDanieau:


apirrone:
It works on my machine, but still slows down the simulation a lot. I'll roll back the image grabbing in develop, let's keep it in this branch until we find a good way of doing that

FabienDanieau:
@apirrone can you try again, I've merged back develop with your patch (#159) 

`python examples/camera_viewer.py --udp-camera udp://@127.0.0.1:5005`

Can you fix the camera orientation? Otherwise I'll do it

Issue #8, opened 2025-06-03
Implement goto/traj API


apirrone:
Depends on #3 

- [x] goto()
- [ ] lookat()
- [x] wake_up()
- [x] sleep()
- [ ] reachability() (?)

"Hidden" 
- [x] _ik() 
- [ ] _fk()
- [x] _turn_on()
- [x] _turn_off()
- [x] _set_joints()
- [x] _get_joints()

pierre-rouanet:
Add a dt directly in the command message?

apirrone:
You mean the duration of the move ?

pierre-rouanet:
Yes

Issue #34, opened 2025-06-05
Add basic arguments parsing


pierre-rouanet:
- [ ] Real/Simulated
- [ ] For Client/Server io
  - [ ] localhost only

Issue #31, opened 2025-06-05
Clean client API


pierre-rouanet:
Wrap client/server etc in a clean API.

## Proposition

```
from reachy_mini import ReachyMini

mini = ReachyMini(mujoco=True)
# or
mini = ReachyMini(serial_port="/dev/ttyUSB0")

mini.set_head_pose(np.eye(4))
...
```

apirrone:
Maybe `sim=True` instead of `mujoco=True` ?

pierre-rouanet:
I've iterated a bit on what we could propose. I think it's best to better separate client and daemon API. My above suggestion tend to mix up both, or at least it should not be the "way to go".

## Basic usage
First, we provide basic wrapper function to spawn MuJoCo or the daemon controller. Here, I would mostly make the name clearer to better represent what each part is actually doing.

```python

p = reachy_mini.spawn_mujoco()
# or for a real hardware robot
p = reachy_mini.spawn_control_daemon(serial_port="/dev/ttyACM0")
```

Then, we have the ReachyMini (I would not talk about client anymore) API:
```python
from reachy_mini import ReachyMini

# Connect to any existing daemon - simulated or real.
# If no daemon is found, it will fail.
lemini = ReachyMini()

lemini.set_pose(head=np.eye(4))
lemini.set_pose(antennas=..)

lemini.goto_pose(head=..., antennas=..., duration=2)
```

In my opinion, it's best to keep those options, as you could want to run a daemon once and then keep starting/exiting new clients.
It also allows to hot-swap from a mujoco to a real one, etc.
So, more flexibility.

## All in one, as a specific use

```python
# Spawn a Reachy Mini in a MuJoCo simulation.
# Connect to it.
# What happens if there is already another daemon running?
robert = ReachyMini(spawn_mujoco=True)

# Runs the daemon to control a real hardware Reachy Mini.
# Connect to it.
# What happens if there is already another daemon running?
e_boo = ReachyMini(spawn_control_daemon=True, serial_port="/dev/ttyACM0")
```

## Next steps

If this way seems right with you @apirrone , @cdussieux , I'll start with the API for ReachyMini and "manually" spwaning daemons. So we can quickly work on examples and behaviors with a "frozen" API and then, I can add the convenience start up method which will be easier to integrate. 

Mentioning #8 to remember those as well.

What do you say?

Issue #28, opened 2025-06-05
Add a keep alive message on the server side


pierre-rouanet:
To detect if we are connected and check the connection status on the client side.

PR #45, opened 2025-06-11
31 SDK definition

pierre-rouanet:
Implements a SDK to let users control ReachyMini in the simplest way possible.

Fixes issues #8 #28 #31 #34 

You first need to run a daemon, that deals with the communication with the motors (either the real physical robot or the mujoco version):
```reachy-mini-daemon -p /dev/ttyACM0```
or
```reachy-mini-daemon --sim```

Then, you can use the Python SDK to:

* connect to the robot (it will automatically  run the wake_up procedure at startup and the goto_sleep at exit)
* send pose target (cartesian or joint using a semi-hidden API)
* use goto command with min jerk interpolation

```python
import time

import numpy as np
from scipy.spatial.transform import Rotation as R

with ReachyMini() as mini:
    t0 = time.time()
    while True:
        t = time.time() - t0

        if t > 10:
            break

        target = np.deg2rad(30) * np.sin(2 * np.pi * 0.5 * t)

        yaw = target
        head = np.eye(4)
        head[:3, 3] = [0, 0, 0.177]
        head[:3, :3] = R.from_euler("xyz", [0, 0, yaw], degrees=False).as_matrix()

        mini.set_position(head=head, antennas=np.array([target, -target]))

        time.sleep(0.01)
```

### DIFF
--- a/src/reachy_mini/reachy_mini.py
+++ b/src/reachy_mini/reachy_mini.py
@@ -1,7 +1,7 @@
 import json
 import os
-from pathlib import Path
 import time
+from pathlib import Path
 from typing import List, Optional
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


src/reachy_mini/reachy_mini.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+ROOT_PATH = Path(os.path.dirname(os.path.abspath(__file__))).parent.parent
+
+pygame.mixer.init()
+
+
+class ReachyMini:
+    def __init__(self) -> None:
+        self.client = Client()
```
cdussieux: Are we assuming here Client defaut arg for local only mode ?
In the current state we cant reachy a demon running on network ?

pierre-rouanet: No we should indeed also specify local_only here as well.

That's a bit odd that we need to define it both on the server and client side, but that's the only way I manage to make it work. It's probably more due to my lack of zenoh knowledge than a real limititation imo.

cdussieux:
lgtm
tested on simulation and real robot

PR #58, opened 2025-06-13
basic led ring integration

cdussieux:
Basic ring led integration
Can set every led color at once or specific ones, keeping the previous colors

Could have in later issues : 
* Cross platform auto detection of arduino port 
*  Add some ring simulation in mujoco ?

### DIFF
--- a/src/reachy_mini/reachy_mini.py
+++ b/src/reachy_mini/reachy_mini.py
@@ -21,7 +21,6 @@ pygame.mixer.init()
 
 # Behavior definitions
 INIT_HEAD_POSE = np.eye(4)
-INIT_HEAD_POSE[2, 3] = 0.177
 
 SLEEP_HEAD_JOINT_POSITIONS = [
     0.0,
--- a/apps/hello_app/hello_app/main.py
+++ b/apps/hello_app/hello_app/main.py
@@ -18,7 +18,6 @@ class HelloApp(ReachyMiniApp):
 
             yaw = target
             head = np.eye(4)
-            head[:3, 3] = [0, 0, 0.177]
             head[:3, :3] = R.from_euler("xyz", [0, 0, yaw], degrees=False).as_matrix()
 
             reachy_mini.set_position(head=head, antennas=np.array([target, -target]))
--- a/examples/api_playground.py
+++ b/examples/api_playground.py
@@ -19,7 +19,6 @@ def main():
 
             yaw = target
             head = np.eye(4)
-            head[:3, 3] = [0, 0, 0.177]
             head[:3, :3] = R.from_euler("xyz", [0, 0, yaw], degrees=False).as_matrix()
 
             mini.set_position(head=head, antennas=np.array([target, -target]))
--- a/examples/client_example.py
+++ b/examples/client_example.py
@@ -9,7 +9,7 @@ with ReachyMini(spawn_daemon=True, use_sim=False) as reachy_mini:
     try:
         while True:
             pose = np.eye(4)
-            pose[:3, 3][2] = 0.177 + 0.005 * np.sin(
+            pose[:3, 3][2] = 0.005 * np.sin(
                 2 * np.pi * 0.3 * time.time() + np.pi
             )
             euler_rot = [
--- a/examples/hand_track_demo.py
+++ b/examples/hand_track_demo.py
@@ -46,7 +46,6 @@ cap = cv2.VideoCapture(4)
 
 hand_tracker = HandTracker()
 pose = np.eye(4)
-pose[:3, 3][2] = 0.177  # Set the height of the head
 euler_rot = np.array([0.0, 0.0, 0.0])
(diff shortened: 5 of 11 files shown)


apirrone:
Works !

src/ring_led_arduino/sketch_jun12a.ino
pierre-rouanet, line 1: I find it a bit odd that this file is inside the src folder. But I'm not sure about the standard way to handle this.

src/reachy_mini/reachy_mini.py
```diff
@@ -10,21 +10,28 @@
...
+    def __init__(self, localhost_only: bool = True, spawn_daemon: bool = False, use_sim: bool = True, led_ring_port=None) -> None:
         daemon_check(spawn_daemon, use_sim)
         self.client = Client(localhost_only)
+
         self.client.wait_for_connection()
 
+        if led_ring_port is not None:
+            self.led = NeoPixelRing(led_ring_port)
+        else:
+            self.led = False
```
pierre-rouanet: better to use ```self.led = None```, no?

src/reachy_mini/io/pixel_ring.py
```diff
@@ -0,0 +1,153 @@
...
+        Initialize NeoPixel ring controller
+
+        Args:
+            port: Serial port (e.g., 'COM3' on Windows, '/dev/ttyUSB0' on Linux)
+            baudrate: Serial communication speed (default: 9600)
+            num_pixels: Number of LEDs in ring (default: 12)
+        """
+        self.num_pixels = num_pixels
+        self.serial = serial.Serial(port, baudrate, timeout=2)
+        time.sleep(2)  # Wait for Arduino to initialize
```
pierre-rouanet: This will slow down the entire boot of the robot. Maybe we could try to spawn a thread instead and add a ready event (like in the zenoh_client for instance).
We could just add an issue for the moment and enhance this later.

src/reachy_mini/io/pixel_ring.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # Wait for Arduino ready signal
+        while True:
+            if self.serial.in_waiting:
+                response = self.serial.readline().decode().strip()
+                if response == "READY":
```
pierre-rouanet, line 44: This will only works the first time you connect to the serial, no?

src/reachy_mini/io/pixel_ring.py
```diff
@@ -0,0 +1,153 @@
...
+        self,
+        colors: Union[
+            List[Optional[Tuple[int, int, int]]], Dict[int, Tuple[int, int, int]]
+        ],
+    ):
+        """
+        Set LED colors. LEDs with None values keep their previous state.
+
+        Args:
+            colors: Either a list of (r,g,b) tuples with None for unchanged LEDs,
```
pierre-rouanet, line 87: Do we really need to handle both ways? I would vote for only keeping either the tuple way or changing the set_colors signature to
```def set_colors(self, r: Optional[int] = None, g: Optional[int] = None, b: Optional[int] = None):```

File: src/reachy_mini/io/pixel_ring.py [hunk 5]
```diff
@@ -0,0 +1,153 @@
...
+                if color is not None:
+                    r, g, b = color
+                    command_parts.append(f"{led_id},{r},{g},{b}")
+
+        if command_parts:
+            command = "SET:" + ";".join(command_parts) + "\n"
+            self.serial.write(command.encode())
+
+            # Wait for confirmation
+            response = self.serial.readline().decode().strip()
```
pierre-rouanet, line 121: Here as well, I would add to the issue that we should also run that asynchronously to avoid slowing down the rest of the control if the led serial communication is not really fast.

src/reachy_mini/io/pixel_ring.py
pierre-rouanet: Oh I just got it that you need to handles all led separately! My bad.

Maybe we should rename it to ```set_leds_color``` to make it more obvious?

src/ring_led_arduino/sketch_jun12a.ino
cdussieux: quite right about it feeling odd
more comfortable with the new way

cdussieux:
@pierre-rouanet 

Overall improvements for this branch

- [ ] reworked the blocking parts
- [ ] moved everything to daemon, for future proofing (:crossed_fingers: raspberry ready)
- [ ] added animation management to allow playing led animation without worring of movement and so on
- [ ] tested auto detect arduino feature to avoid hardcode usb path, should be checked on mac and linux

apirrone:
Should we close this PR? We won't have a LED ring in the end right ? @cdussieux @pierre-rouanet

pierre-rouanet:
Not for v1 at least. We can close the PR and keep the branch. Does that work for you @cdussieux?

PR #60, opened 2025-06-13
Wakeup/gotosleep on daemon.

pierre-rouanet:


cdussieux:
tested on real robot
seems ok

PR #61, opened 2025-06-14
Enhance MuJoCo startup initialization

matthieu-lapeyre:
- set the camera in more convenient position by default
- start into the “sleep” pose 

All changes are confined to isolated viewer.lock() blocks and don’t alter the main control loop.

Capture d’écran 2025-06-14 à 13 55 55


matthieu-lapeyre:
BTW I use the same sleep position, but it does not look the same in mujoco.

cdussieux:
no impact seen on real robot

Issue #57, opened 2025-06-12
Add task space goto


apirrone:
Like @matthieu-lapeyre did here : https://github.com/pollen-robotics/reachy_mini/blob/179178d98bc983d625d00c78d6186d6d5cb46039/examples/animation.py#L6

matthieu-lapeyre:
Should I integrate what I did ?

apirrone:
Closed by #64

PR #64, opened 2025-06-15
Task-space goto

matthieu-lapeyre:
Related to issue #57

Added goto in task space with several method to interpolate time.

It seems to work.

### DIFF
--- a/src/reachy_mini/utils.py
+++ b/src/reachy_mini/utils.py
@@ -108,10 +108,12 @@ def daemon_check(spawn_daemon, use_sim):
         )
 
 
-def linear_pose_interpolation(start_pose:np.ndarray, target_pose:np.ndarray, t:float):
+def linear_pose_interpolation(
+    start_pose: np.ndarray, target_pose: np.ndarray, t: float
+):
     # Extract rotations
     rot_start = R.from_matrix(start_pose[:3, :3])
-    rot_end   = R.from_matrix(target_pose[:3, :3])
+    rot_end = R.from_matrix(target_pose[:3, :3])
 
     # Compute relative rotation q_rel such that rot_start * q_rel = rot_end
     q_rel = rot_start.inv() * rot_end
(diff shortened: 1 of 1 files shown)


src/reachy_mini/utils.py
```diff
@@ -93,3 +94,88 @@ def is_python_script_running(script_name):
...
+    pose = np.eye(4)
+    rot = R.from_euler("xyz", [roll, pitch, yaw], degrees=degrees).as_matrix()
+    pose[:3, :3] = rot
+    pose[:, 3] = [x, y, z, 0]
+    if mm:
+        pose[:3, 3] /= 1000
+
+    pose[2, 3] += 0.177 # :(
+    return pose
+
```
pierre-rouanet: Choose your contender :) And remove one the duplicates (create_pose = create_head_pose).

src/reachy_mini/utils.py
```diff
@@ -93,3 +94,88 @@ def is_python_script_running(script_name):
...
+                return ( (2 * t) ** 2 * ( (c2 + 1) * 2 * t - c2 ) ) / 2
+            else:
+                # phase out
+                return ( ((2 * t - 2) ** 2 * ( (c2 + 1) * (2 * t - 2) + c2 ) ) + 2 ) / 2
+
+        case _:
+            raise ValueError("Unknown interpolation method")
+        
+
+def create_pose(x=0, y=0, z=0, roll=0, pitch=0, yaw=0, mm=False, degrees=True):
```
pierre-rouanet, line 175: @apirrone ask me to only use radians by default 😓 
At least I think we should be consistent (today all joints commands are in radians).

src/reachy_mini/reachy_mini.py
```diff
@@ -60,21 +61,21 @@ def set_position(
             antenna_joint_positions = None
 
         self._send_joint_command(head_joint_positions, antenna_joint_positions)
+        self._last_head_pose = head
```
pierre-rouanet, line 96: I'm not sure to understand why you need this instead of using the current joint pos + fk? Is this to avoid some discontinuities?

src/reachy_mini/utils.py
matthieu-lapeyre: Agree we should keep the same unit everywhere.

But not sure if radians are "user friendly unit"

src/reachy_mini/reachy_mini.py
matthieu-lapeyre: Yes it's quite the same problem we had with pypot goto when you use "present_position" rather than "goal_position", so I store the goal_pose to start from there when starting a goto.

src/reachy_mini/utils.py
matthieu-lapeyre: BTW the create pose func is a quick and dirty func I added because it was annoying to redo the same thing over and over to create a complete 4x4 pose matrix.

src/reachy_mini/utils.py
matthieu-lapeyre: typo

src/reachy_mini/reachy_mini.py
pierre-rouanet: Ok thx, we should probably add a comment in the code not to forgot why is it used in further development.

pierre-rouanet:
Fix #75 before merge as well.

PR #67, opened 2025-06-17
Updating reachy mini description (URDF/MJCF)

apirrone:
Up to date with the Alpha version of the robot. 

- [x] Test in simulation
- [x] Test with the real robot


cdussieux:
tested on real robot, looks good on client_example

cdussieux:
tested again on real robot, still ok

PR #72, opened 2025-06-18
inverting stewart and antennas dofs in onshape, should now match real robot

apirrone:


cdussieux:
tested on robot, sequence and head_tracking ok

PR #78, opened 2025-06-19
77 add ruff linter

FabienDanieau:
using pre-commit package (see README). Ruff should be automatically called before each commit.
What do you think?


pierre-rouanet:
Do we also add the .vscode basic config?

Issue #81, opened 2025-06-19
remove setup.cfg


FabienDanieau:
pyproject.toml should be enough

PR #83, opened 2025-06-19
enhancement #81: remove setup.cfg. use pyproject.toml only

FabienDanieau:
setup.cfg was used on bigger projects because some tools (flake8) were not compatible with pyproject.toml
We can get rid of it and make the configuration simpler (flake8 won't be used see #77 )


pierre-rouanet:
Works fine on a brand new venv 🆗

PR #96, opened 2025-06-23
catch ctrl-c after the first one

cdussieux:
Not quite important for now, but fixing is could be as simple as this
to avoid close error if someone spams ctrl-c


pierre-rouanet:
Pressed Ctrl-C as fast as I can be stupid. Did not manage to break it. Good job 🔥

Issue #90, opened 2025-06-20
URDF is not installed when not using the pip -e mode.


pierre-rouanet:

apirrone:
I don't understand, what is not installed ?

pierre-rouanet:
The file are not copied to the build folder. I guess we should follow https://setuptools.pypa.io/en/latest/userguide/datafiles.html#include-package-data

apirrone:
This needs breaking refactoring (moving stuff around). Let's wait after the shooting :)

PR #99, opened 2025-06-24
Fix `pip install .` (no `-e`)

apirrone:
Moving all assets in the package (`src/reachy_mini/`) and using `importlib` instead of `ROOT_PATH`

Adding a `MANIFEST.in`

Basically following the standard for python packages now

Tested in simulation and on the real robot

This PR should be merged carefully, it is likely to produce conflicts and could be breaking for features that are in branches

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -5,6 +5,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          lfs: 'true'
       - uses: actions/setup-python@v5
         with:
           python-version: '3.10'


cdussieux:
uninstalled the old package
install the #90 version with ` pip install .` (no -e)
No trivial issue

apirrone:
BTW I didn't really know what to do with the "apps" folder, do you think it should be moved too ?

pierre-rouanet:
I've added a unit test to load placo with the urdf.

For the apps folder, I think we should leave it outside so they can be installed independently. Maybe we also think about the possibility to also include reach-mini-motor-controller and have a single repo for everything related to reach mini (probably better in terms of referencement).

PR #112, opened 2025-06-26
Add Documentation

apirrone:
- [x] Docstring for reachy_mini.py. 
  - [x] renaminig goto_position() to goto_target()
  - [x] renaming_send_joint_command() to _set_joint_positions()
  - [x] renaming set_position() to set_target()
- [x] app.py
- [x] abstract.py
- [x] backend.py
- [x] daemon.py
- [x] zenoh_client.py
- [x] zenoh_server.py
- [x] mujoco_backend.py
- [x] mujoco_utils.py
- [x] placo_kinematics.py
- [x] robot_backend.py
- [x] utils.py
- [x] video_udp.py

### DIFF
--- a/src/reachy_mini/reachy_mini.py
+++ b/src/reachy_mini/reachy_mini.py
@@ -7,6 +7,7 @@ os.environ["PYGAME_HIDE_SUPPORT_PROMPT"] = "hide"
 
 from importlib.resources import files
 
+import cv2
 import numpy as np
 import pygame
 from scipy.spatial.transform import Rotation as R
@@ -20,7 +21,6 @@ from reachy_mini.utils import (
     minimum_jerk,
     time_trajectory,
 )
-import cv2
 
 try:
     pygame.mixer.init()
@@ -75,10 +75,11 @@ class ReachyMini:
         localhost_only: bool = True,
         spawn_daemon: bool = False,
         use_sim: bool = True,
+        timeout: float = 5.0,
     ) -> None:
         daemon_check(spawn_daemon, use_sim)
         self.client = Client(localhost_only)
-        self.client.wait_for_connection()
+        self.client.wait_for_connection(timeout=timeout)
         self._last_head_pose = None
 
         self.head_kinematics = PlacoKinematics(self.urdf_root_path)
--- a/examples/hand_track_demo.py
+++ b/examples/hand_track_demo.py
@@ -5,6 +5,7 @@ from reachy_mini import ReachyMini
 import numpy as np
 from scipy.spatial.transform import Rotation as R
 
+from reachy_mini.io.cam_utils import find_camera
 
 def draw_debug(img, palm_center):
     h, w, _ = img.shape
@@ -35,8 +36,7 @@ def draw_debug(img, palm_center):
     )
 
 
-cap = cv2.VideoCapture(4)
-# cap = cv2.VideoCapture(0)
+cap = find_camera()
 
 hand_tracker = HandTracker()
 pose = np.eye(4)
--- a/examples/head_track_demo.py
+++ b/examples/head_track_demo.py
@@ -5,6 +5,7 @@ from reachy_mini import ReachyMini
 import numpy as np
 from scipy.spatial.transform import Rotation as R
 
+from reachy_mini.io.cam_utils import find_camera
 
 def draw_debug(img, eye_center, roll):
     _eye_center = (eye_center.copy() + 1) / 2  # [0, 1]
@@ -45,8 +46,7 @@ def draw_debug(img, eye_center, roll):
     )
 
 
-cap = cv2.VideoCapture(4)
-# cap = cv2.VideoCapture(0)
+cap = find_camera()
 
 head_tracker = HeadTracker()
 pose = np.eye(4)
--- a/examples/lookat_example.py
+++ b/examples/lookat_example.py
@@ -2,8 +2,9 @@ from reachy_mini import ReachyMini
 import cv2
 import numpy as np
 
-cap = cv2.VideoCapture(0)
+from reachy_mini.io.cam_utils import find_camera
 
+cap = find_camera()
 
 def click(event, x, y, flags, param):
     global click_x, click_y, just_clicked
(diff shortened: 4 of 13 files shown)


apirrone:
I'll probably do a re-pass on the docstring in the near future, but I think it's a good idea to merge this PR soon since it renames some api functions too. What do you think @pierre-rouanet @cdussieux @FabienDanieau @matthieu-lapeyre ?

pierre-rouanet:
Yes, agreed. I'll review it now.

README.md
```diff
@@ -66,7 +66,7 @@ from reachy_mini import ReachyMini
 with ReachyMini() as reachy_mini:
     ...
     ...
-    reachy_mini.set_position(head=pose, antennas=antennas)s
+    reachy_mini.set_target(head=pose, antennas=antennas)s
```
pierre-rouanet: There is still a typo with the s outside of ).

examples/goto_interpolation_playground.py
```diff
@@ -10,25 +10,31 @@ def main():
...
 
             for _ in range(3):
                 pose = create_head_pose(
                     x=0.0, y=0.03, z=0, roll=5, yaw=-10, degrees=True
                 )
-                mini.goto_position(
-                    pose, antennas=np.deg2rad([-20, 20]), duration=1.0, method=method
+                mini.goto_target(
+                    pose,
+                    antennas=list(np.deg2rad([-20, 20])),
```
pierre-rouanet: Is it necessary to force a list here? Both should work imo.

examples/goto_interpolation_playground.py
```diff
@@ -10,25 +10,31 @@ def main():
...
                 )
 
                 pose = create_head_pose(
                     x=0.0, y=-0.03, z=0, roll=-5, yaw=10, degrees=True
                 )
-                mini.goto_position(
-                    pose, antennas=np.deg2rad([20, -20]), duration=1.0, method=method
+                mini.goto_target(
+                    pose,
+                    antennas=list(np.deg2rad([20, -20])),
```
pierre-rouanet: same

src/reachy_mini/io/zenoh_server.py
```diff
@@ -7,7 +7,7 @@
 
 
 class ZenohServer(AbstractServer):
-    def __init__(self, backend: Backend, localhost_only: bool = True):
+    def __init__(self, backend: Backend, localhost_only: bool = True):  # type: ignore
```
pierre-rouanet, line 10: Why is this needed?

src/reachy_mini/reachy_mini.py
```diff
@@ -52,8 +52,19 @@
     ]
 )
 
+IMAGE_SIZE = (1280, 720)  # Width, Height in pixels
```
pierre-rouanet, line 55: I would pass this as argument to the look_at_image function as we do not really know at wcich resolution we will open the camera

src/reachy_mini/reachy_mini.py
```diff
@@ -86,62 +98,65 @@ def __init__(
...
-    def __enter__(self):
+    def __enter__(self) -> "ReachyMini":
         return self
 
-    def __exit__(self, exc_type, exc_value, traceback):
+    def __exit__(self, exc_type, exc_value, traceback) -> None:
         self.client.disconnect()
 
-    def set_position(
+    def set_target(
```
pierre-rouanet, line 108: I'm not sure the best way to handle that but I think it would be nice to support both. Like when you use the numpy API, you can usually use both array or list.

src/reachy_mini/reachy_mini.py
```diff
@@ -321,7 +341,20 @@ def _goto_joint_positions(
...
+        """
+        [Internal] Go to a target head joint positions and/or antennas joint positions using joint space interpolation, in "duration" seconds.
+
+        Args:
+            head_joint_positions (Optional[List[float]]): List of head joint positions in radians (length 7).
+            antennas_joint_positions (Optional[List[float]]): List of antennas joint positions in radians (length 2).
+            duration (float): Duration of the movement in seconds. Default is 0.5 seconds.
+        """
+
+        assert duration > 0.0, (
```
pierre-rouanet: I think raising a ValueError is a more pythonic way of dealing with this kind of issue.

src/reachy_mini/reachy_mini.py
apirrone: hmm, the `K` matrix also depends on this resolution, maybe I should just pass image_size, k and d as arguments to the ReachyMini class, with default values ?

examples/goto_interpolation_playground.py
apirrone: It will work, but the type hinting expects a list, so I get an error when passing a np array : 
image

apirrone: Using : `Optional[Union[np.ndarray, List[float]]]` we can use both list and ndarray

src/reachy_mini/reachy_mini.py
apirrone: I set it to `Optional[Union[np.ndarray, List[float]]]`

src/reachy_mini/reachy_mini.py
apirrone: Should I replace all asserts with raising ValueErrors ?

pierre-rouanet: To me the idea is:
- raise a ValueError if this is a user mistake to use this value
- use assert if this should never occurred (like for an internal api)

src/reachy_mini/reachy_mini.py
apirrone: I guess since it's only used in `look_at_image()` for now we can keep it that way ? It will only work with an image of size (1280, 720)  anyways, so these asserts will prevent some weird behaviors

image

apirrone: If it's okay with you @pierre-rouanet, we can merge this PR I think

pierre-rouanet:
Let's resolve the conflict and merge it!

PR #113, opened 2025-06-26
71 find the robot camera opencv id automatically

pierre-rouanet:
I've added support for macos on cv2_enumerate_camers (see https://github.com/chinaheyu/cv2_enumerate_cameras/pull/8). 
Merged in version 1.2.1.

Works on my Mac.
On my raspberry pi it doesn't work with the gestreamer backend so I forced the v4l2 one. Does that seem ok to you @FabienDanieau @apirrone?


apirrone:
It works for me !

PR #118, opened 2025-07-02
Feat added the auto-collision detection + analytical kinematics

askuric:
So this PR does two things:

1. Adds the auto-collision detection to the FK and IK resolution, which will be stopped if it happens. 
    - The default margin of collsion is 5mm, this is currently hardcoded in the code
    - Currently only the head and the torso are checked, not the eyes or the platform bars - this can be added if necessary
    - The collision checks are not active by default, we need to activate them in the function call `set_target(...., collision_check=True)`
         - There is a time penalty of using the collision check so I did not want to set it active by default 
         - If the performace is not an issue we can set this to default later 
2. All the QP solvers constraints have gone to "soft"  to avoid indeasibility violations of the QP
    - This is a patch rather than a fix, and we should probably address it later more in depth
3. Added the analytical IK and jacobian calculation
    - Not used by default, 
    - But can be used as an additional IK. see the example [here](https://github.com/pollen-robotics/reachy_mini/blob/feat_analytical_kinematics/examples/analytical_kinematics/gui_cs_analytical_ik_example.py)
    - We could use this by default later
    - The derivation of the jacobians and the analytical IK can be found in the `examples/analytical_kinematics` forlder - [here](https://github.com/pollen-robotics/reachy_mini/blob/feat_analytical_kinematics/examples/analytical_kinematics/compute_analytical_kinematics.py)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,6 +5,7 @@ repos:
   hooks:
     # Run the linter.
     - id: ruff-check
-      args: ["--select", "I", "--select", "D", "--ignore", "D203", "--ignore", "D213"]
+      args: ["--select", "I"]
+      # args: ["--select", "I", "--select", "D", "--ignore", "D203", "--ignore", "D213"]
     # Run the formatter.
     - id: ruff-format
--- a/examples/mini_head_position_gui.py
+++ b/examples/mini_head_position_gui.py
@@ -1,10 +1,10 @@
 import time
+import tkinter as tk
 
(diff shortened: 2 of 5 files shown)


apirrone:
Fixed a small error that occured when having a  collision set_target()
Otherwise, it all seems to work, all exemples still work the same (except for mirror_xyroll but it's actually broken in develop too) 
I activate collision checking in the hand tracking demo and it works well, we may be able to activate it everywhere. I didn't notice any performance issues.

PR #132, opened 2025-07-08
52 create dance app demo

pierre-rouanet:


RemiFabre:
@pierre-rouanet can you review rhythmic_motion.py please? Most of the changes are there, the examples clean up are coming soon.
- I went for all in the same file again, because now it's part of an utils folder. It differs from interpolation functions because here everything is meant to be rythmic (functions loop, time is in tempo, etc).
- Type hints+docstrings. Iterated a bit on how we explain stuff. It's verbose but should clear confusions for people (e.g. what does t_beats mean)
- No more args are now named, the number of "*" has been reduced :)
- Passes ruff

RemiFabre:
Update:
Improved the demo_dance.py script to read a choreography file. Test it with:
```
cd reachy_mini/examples
python dance_demo.py --bpm 114 --choreography choreographies/another_one_bites_the_dust.json
```
Must listen to this:
https://www.youtube.com/watch?v=rY0WxgSXdEE

src/reachy_mini/utils/rhythmic_motion.py
```diff
@@ -0,0 +1,1385 @@
...
+    form the mathematical foundation.
+2.  **Atomic Moves:** Simple wrappers around primitives that control a single
+    axis or function (e.g., `atomic_pitch`, `atomic_antenna_wiggle`). These
+    are the fundamental building blocks.
+3.  **Choreographed Moves:** The main library of named moves, created by
+    combining multiple atomic moves using the `combine_offsets` utility.
+
+Key Parameters
+--------------
+All move functions are driven by `t_beats` and a set of keyword arguments:
```
pierre-rouanet, line 21: This is really nice, but really written from a "scientific" perspective. I'm wondering if it would make sense to have basically the same thing but written for musician, choreograph, animators, etc.?

src/reachy_mini/utils/rhythmic_motion.py
```diff
@@ -0,0 +1,1385 @@
...
+- `duration_beats` / `repeat_every` (float): Parameters for transient,
+  one-shot moves.
+
+Usage Example
+-------------
+
+```python
+# 1. Get the move function and its default parameters
+move_fn = AVAILABLE_DANCE_MOVES['simple_nod']
+params = MOVE_SPECIFIC_PARAMS['simple_nod']
```
pierre-rouanet, line 40: Maybe it would be easier and less error prone to merge those two dict?

And you could do something like:
```python
(move_fn, params) = AVAILABLE_DANCE_MOVES['simple_nod']
```

src/reachy_mini/utils/rhythmic_motion.py
```diff
@@ -0,0 +1,1385 @@
...
+
+# 2. At each time step, call the function with the current beat time
+t_beats = 2.5
+offsets = move_fn(t_beats, **params)
+
+# 3. Apply the resulting offsets to the robot    "ruff.format": {
```
pierre-rouanet: You see ruff everywhere :)

src/reachy_mini/utils/rhythmic_motion.py
```diff
@@ -0,0 +1,1385 @@
...
+
+
+AVAILABLE_ANTENNA_MOVES: dict[
+    str, Callable[[float, OscillationParams], MoveOffsets]
+] = {"wiggle": atomic_antenna_wiggle, "both": atomic_antenna_both}
+
+# ─────────────────────────── ALL DANCE MOVE FUNCTIONS ─────────────────────────────
+
+
+def move_simple_nod(
```
pierre-rouanet: To make this part more like a library of available moves, I would split them into their own file and maybe add a header with the list of available moves.
Maybe even a markdown with the list and a gif of what each move does.

examples/choreographies/another_one_bites_the_dust.json
```diff
@@ -0,0 +1,37 @@
+[
```
pierre-rouanet: Shouldn't the bpm be defined inside this file as well?

pierre-rouanet: We could add a link to the sound and play it as well. 
There are maybe some licensing issue. I don't really know.

examples/dance_demo.py
```diff
@@ -0,0 +1,438 @@
+#!/usr/bin/env python3
+"""Interactive Dance Move Tester and Choreography Player for Reachy Mini.
+
```
pierre-rouanet, line 3: This made me want to have a GUI with all available moves and their params, and just click directly on a move to play it. Kind like an artist who would explore/try combination.

examples/dance_demo.py
```diff
@@ -0,0 +1,438 @@
...
+
+# --- Constants for UI ---
+INTERACTIVE_HELP_MESSAGE = """
+┌────────────────────────────────────────────────────────────────────────────┐
+│                           INTERACTIVE CONTROLS                             │
+├──────────────────────────────────┬─────────────────────────────────────────┤
+│ Q / Ctrl+C : Quit Application    │ P / Space : Pause / Resume Motion       │
+│ Left/Right : Previous/Next Move  │ Up/Down   : Decrease / Increase BPM     │
+│ W          : Cycle Waveform      │ + / -     : Increase / Decrease Amplitude │
```
pierre-rouanet: my eyes

examples/dance_demo.py
```diff
@@ -0,0 +1,438 @@
...
+        self.prev_move: bool = False
+        self.next_waveform: bool = False
+        self.bpm_change: float = 0.0
+        self.amplitude_change: float = 0.0
+
+    def toggle_pause(self) -> bool:
+        """Toggle the running state and return the new state."""
+        with self.lock:
+            self.running = not self.running
+        return self.running
```
pierre-rouanet: Guess this should be in the lock as well. (Probably has no effect in reality).

examples/dance_demo.py
```diff
@@ -0,0 +1,438 @@
...
+                "amplitude_change": self.amplitude_change,
+            }
+            self.next_move = self.prev_move = self.next_waveform = False
+            self.bpm_change = self.amplitude_change = 0.0
+            return changes
+
+
+# --- Robot Interaction & Utilities ---
+def head_pose(pos: np.ndarray, eul: np.ndarray) -> np.ndarray:
+    """Create a 4x4 homogenous transformation matrix for the head."""
```
pierre-rouanet: Why not use the existing create_head_pose fonction?

examples/dance_demo.py
```diff
@@ -0,0 +1,438 @@
...
+                    print(
+                        f"\n{CHOREO_HELP_MESSAGE if choreography_mode else INTERACTIVE_HELP_MESSAGE}"
+                    )
+                    last_help_print_time = loop_start_time
+
+                time.sleep(max(0, config.control_ts - (time.time() - loop_start_time)))
+
+    except KeyboardInterrupt:
+        print("\nCtrl-C received. Shutting down...")
+    except Exception as e:
```
pierre-rouanet: I would just not catch it. My feeling would be that something unexpected happened, we should not try to hide it.

pierre-rouanet:
Really cool and exciting! I've added some minor comments above.

It definitely makes me wonder how far we can push that by providing a GUI, make the robot dance live using a [pad controller](https://www.amazon.fr/AKAI-Professional-MPD218-Contrôleur-Logiciels/dp/B0116X17JW/ref=asc_df_B0116X17JW?mcid=ab4b956f4a793cc2be213cda22a88745&tag=googshopfr-21&linkCode=df0&hvadid=701605946357&hvpos=&hvnetw=g&hvrand=5399872094153535667&hvpone=&hvptwo=&hvqmt=&hvdev=c&hvdvcmdl=&hvlocint=&hvlocphy=9196235&hvtargid=pla-404286153279&psc=1&hvocijid=5399872094153535667-B0116X17JW-&hvexpln=0), or connect it to LLM and see if we can generate new choreographies, new moves, etc!

src/reachy_mini/utils/rhythmic_motion.py
RemiFabre: Interesting. Not sure how to write it. If I had to do it I'd go with something more visual+sound. But I think this description already works well for LLMs to generate stuff from human description, so maybe non experts can play with this without having to read all of this

examples/dance_demo.py
RemiFabre: lol

src/reachy_mini/utils/rhythmic_motion.py
RemiFabre: Good change, done.

examples/choreographies/another_one_bites_the_dust.json
RemiFabre: Probably licensing issues, although playing the sound ourselves does solve a lot of sound analysis problems. We should get a collab with a service like Spotify :)

(and BPM feature integrated)

src/reachy_mini/utils/rhythmic_motion.py
RemiFabre: Split in 2 files -> done. I hope moves.py will grow with creative PRs from the community!

Added an issue for the GIFs, we did it "by hand" for Reachy2 and it's too time consuming. I'll try to find a way to do it programmatically.

RemiFabre: We also need a text description of each move, this will open up automatic move selections by an LLM for example

RemiFabre:
This version should comply with most of the comments.

Regarding the forbidden_moves file...
image

PR #140, opened 2025-07-11
Feature - gravity compensation + current control

askuric:
Hi guys, 

IMPORTANT:
- This PR is connected directly and requires the PR: https://github.com/pollen-robotics/reachy-mini-motor-controller/pull/5

This PR allows:
- setting the control mode of the motors of the stewart platform using the client-server
- setting the current control target to the stewart platform (not to the antennas and the body rotation as they are feetech at the moment)
-  Calculation of the gravity compensation torques and the jacobian (using placo + pinocchio)
- Making the robot fully compliant (example : https://github.com/pollen-robotics/reachy_mini/blob/torque_control/examples/reachy_complient_demo.py)

This PR encompasses the PR #118 and merging this one will merge that one too.

### DIFF
--- a/src/reachy_mini/daemon/backend/robot/backend.py
+++ b/src/reachy_mini/daemon/backend/robot/backend.py
@@ -220,29 +220,28 @@ class RobotBackend(Backend):
             "Invalid operation mode. Must be one of [0 (torque), 3 (position), 5 (current-limiting position)]."
         )
 
-        if self._head_operation_mode != mode:
-            # if motors are enabled, disable them before changing the mode
-            if self._torque_enabled:
-                self.c.enable_stewart_platform(False)
-            # set the new operation mode
-            self.c.set_stewart_platform_operating_mode(mode)
+        # if motors are enabled, disable them before changing the mode
+        if self._torque_enabled:
+            self.c.enable_stewart_platform(False)
+        # set the new operation mode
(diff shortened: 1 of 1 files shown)


src/reachy_mini/reachy_mini.py
```diff
@@ -490,3 +511,78 @@ def _set_joint_positions(
...
+    def _set_antennas_operation_mode(self, mode: int) -> None:
+        """Set the operation mode for the antennas motors.
+
+        Args:
+            mode (int): The desired operation mode.
+
+        """
+        self.client.send_command(json.dumps({"antennas_operation_mode": mode}))
+
+    def make_compliant(
```
pierre-rouanet: We now have two functions with close behavior.
- set_torque with an ON/OFF kind of mechanism.
- make_compliant with an intermediate behavior (compliant?)

Shall we merge those two methods? At least, we need to better explain this. @askuric  @FabienDanieau any API feelings?

pierre-rouanet:
Really cool stuff!

Before merge, I would like to:
- [ ] decide what to do with set_torque and make_compliant (my personal feeling is that they should be merged into a single method with on/off/compliant argument).
- [ ] Add doc in the readme and api to explain 
   - [ ] check_collision
   - [ ] make_compliant & gravity compensation

PR #147, opened 2025-07-16
Allowing manual body yaw control

RemiFabre:
POC version, don't merge yet.

I tried a version where the yaw control is decided on object creation only. Not sure of allowing it to be dynamically changed is worth it?

After playing a tad with the task weights, I even wonder if we need 2 modes? i.e the body_yaw is kept "automatic" but we change the task's target position dynamically (like I did here). It's less accurate but simpler.

### DIFF
--- a/src/reachy_mini/reachy_mini.py
+++ b/src/reachy_mini/reachy_mini.py
@@ -103,7 +103,6 @@ class ReachyMini:
         self.client.wait_for_connection(timeout=timeout)
         self._last_head_pose = None
 
-        self._automatic_body_yaw = automatic_body_yaw
         self.head_kinematics = PlacoKinematics(
             self.urdf_root_path, automatic_body_yaw=automatic_body_yaw
         )


RemiFabre:
@askuric

askuric:
Looks good to me. 

There are just a couple of questions that came up:
1. body yaw joint seems to have opposite polarity than the head yaw. Its not a big deal but it might be a source of some misunderstandings. So I'd artificially force the `set_target` function to have the same body_yaw and head yaw sign. 
3. ~Does it make sense to provide a mode where the head movement is relative to the body frame, not absolute. Where the body_yaw would be independent of the head yaw. 
Maybe in some applications it makes sense to control the rotation of the robot completely separately from the head motion. 
So maybe to have two different modes that are a bit more easy to interpret.~
<details><summary> <s>proposition discarded</s> </summary>

Mode | body yaw | placo no DOFs| head pose
---- | ---- | ---- | ----
absolute/automatic | automatic | 7 | expressed in the world frame (done now)
relative/separated | user defined | 6 |expressed in the rotated base frame (after the body yaw)
</details>

askuric:
I've added an example `examples/mini_body_yaw_gui.py` to test the different modes (automatic or moanual body yaw). 
Let me know if it makes sense.

RemiFabre:
After discussion with Antun:
- There is no obvious convention choice for orientation frames. Should the head pose be always in world frame or in body frame? In this PR we chose "always world frame" but it's easy to change if needed.
- We converged on 2 modes that are only changeable on item creation. Changing them dynamically is easy to code but there are edge cases that create discontinuities and we think the use case is too rare to justify the work.
- I suspect the automatic_mode will be used in the vast majority of cases
- Main change of this PR is that now the user can request a body_yaw with set_target().

RemiFabre:
@pierre-rouanet to test:
cd reachy_mini/examples/debug
python mini_body_yaw_gui.py
*play with the gui*

then change in the test script with ReachyMini(automatic_body_yaw=True) as mini: --> False

And retry

src/reachy_mini/reachy_mini.py
```diff
@@ -101,7 +103,10 @@ def __init__(
         self.client.wait_for_connection(timeout=timeout)
         self._last_head_pose = None
 
-        self.head_kinematics = PlacoKinematics(self.urdf_root_path)
+        self._automatic_body_yaw = automatic_body_yaw
```
pierre-rouanet: This is not used.

pierre-rouanet:
Tk is really not well supported on OSX, so I could not use your demo. 

Overall, I don't really know what to think about this. The automatic bool feels more like a patch to a non-decision than a proper solution to me. 
My personal feel would probably go towards only giving the "manual" options.

BTW, please also add the doc of those kind of features into the markdown docs.

RemiFabre:
- After discussion with Pierre, we're pushing this with the default value being manual.
- Added the body_yaw interpolation in the goto_target function. Can be tested with `script debug/test_interpol.py`
- Restoring the initial sign convention for the body yaw because making this change in the high level software is dangerous. We either don't do it, or we do it low level.
- Pass on the documentation

PR #149, opened 2025-07-18
144 move placo to daemon

FabienDanieau:
Please check that all the collision detection, and gravity compensation is there.
@RemiFabre could check that the automatic yaw is still working please?

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -4,7 +4,7 @@
 
 ## ReachyMini
 
-Reachy Mini's API is designed to be simple and intuitive. You will mostly interact with the `ReachyMini` class, which provides methods to control the robot's joints such as the head and antennas and interacti with its sensors.
+Reachy Mini's API is designed to be simple and intuitive. You will mostly interact with the `ReachyMini` class, which provides methods to control the robot's joints such as the head and antennas and interact with its sensors.
 
 The first step is to instantiate the `ReachyMini` class. This can be done as follows:
 
--- a/examples/dance_demo.py
+++ b/examples/dance_demo.py
@@ -43,6 +43,7 @@ class Config:
     neutral_pos: np.ndarray = field(default_factory=lambda: np.array([0, 0, 0.0]))
     neutral_eul: np.ndarray = field(default_factory=lambda: np.zeros(3))
     choreography_path: Optional[str] = None
+    disable_keyboard: bool = False
 
 
 # --- Constants for UI ---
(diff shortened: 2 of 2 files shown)


RemiFabre:
> Please check that all the collision detection, and gravity compensation is there. @RemiFabre could check that the automatic yaw is still working please?

I'm having weird behaviors, inconsistent between runs. I've been testing with:
```
cd reachy_mini/examples/debug
python mini_body_yaw_gui.py
```

@askuric an idea? I have a live debug to attend, available to look deeper this afternoon

RemiFabre:
I confirm that the behavior is different in this branch compared to develop. This script often breaks in this branch (as in "the head does not go where it should") but it works in develop:
```
"""Reachy Mini Head Position debug script."""

import time

import numpy as np
from scipy.spatial.transform import Rotation as R

def main():
    """Run a debug script."""
    with ReachyMini() as mini:
        # with ReachyMini(automatic_body_yaw=False) as mini:
        t0 = time.time()

        while True:
            t = time.time() - t0
            target = np.deg2rad(90) * np.sin(2 * np.pi * 0.5 * t)

            head = np.eye(4)
            head[:3, 3] = [0, 0, 0]

            # Read values from the GUI
            roll = np.deg2rad(0.0)
            pitch = np.deg2rad(0.0)
            yaw = np.deg2rad(0.0)
            head[:3, :3] = R.from_euler(
                "xyz", [roll, pitch, yaw], degrees=False
            ).as_matrix()

            mini.set_target(
                head=head,
                antennas=np.array([target, -target]),
                body_yaw=target,
                check_collision=False,
            )
            time.sleep(0.001)


if __name__ == "__main__":
    main()

```

At first I thought it was related to the high freq calls, but it was also possible to make it fail in this branch by re-running the same script at lower freqs (100Hz and even 10Hz).

RemiFabre:
Update: we had some doubts about the state of this PR because spamming commands at high freq (1000Hz) created (occasionally) weird behaviors. These disappear at lower frequencies (100Hz). Not sure if this deserves more investigation, seems OK

RemiFabre:
A request: can someone please list the changes this creates client side so that we can fix the apps/examples?
For example, to get the current pose of the head, it was:
```
            joints, _ = reachy._get_current_joint_positions()
            pose = reachy.head_kinematics.fk(joints)
```
And now:
```
reachy._get_current_head_pose()
```

I can make a pass on the examples afterwards if you want

pierre-rouanet:
If you can list the "private" method called in the examples, that would be great. We should try to only use public API and maintain them while being free to do what we want on the private ones.

RemiFabre:
Ok we finally converged (hihi) on the Placo issues with Antun (he did most of the work). I'll just update the examples and merge this to develop.

PR #153, opened 2025-07-21
updated the gravity compesation in the backend

askuric:
Ok so this PR is just a bit of refactoring. 
1. The gravity compensation is no longer calculated on the call of `reachy.compensate_gravity()` but continously if the gravity compensation mode is enabled. This will produce better and smoother results while removing the need for this `reachy.compensate_gravity()` call in the loop for the user. 
2. I've added the argument to the `reachy.make_motors_compliant()` called `compensate_gravity` which if in True (default False) will enable the gravity compensation mode and every time the motors are in the compliant mode the gravity will be removed. 
3. There are also the separare funcitons `reachy.enable_gravity_compensation()` and `reachy.disable_gravity_compensation()` that enable/disable the behavior.

This way its a bit more clear to the user. 
Let me know what you think.

### DIFF
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -40,7 +40,6 @@ class Backend:
         self.joint_positions_publisher = None  # Placeholder for a publisher object
         self.pose_publisher = None  # Placeholder for a pose publisher object
         self.error = None  # To store any error that occurs during execution
-        
 
     def wrapped_run(self):
         """Run the backend in a try-except block to store errors."""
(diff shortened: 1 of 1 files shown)


src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -202,16 +204,38 @@ def get_status(self):
...
+
+        Args:
+            mode (bool): If True, gravity compensation is enabled.
+
+        """
+        self.gravity_compensation_mode = mode # True (enable) or False (disable)
+
+    def compensate_head_gravity(self) -> None:
+        """Calculate the currents necessary to compensate for gravity."""
+        print("Compensating for gravity...")
```
FabienDanieau: You may want to remove this or use the logger :)

PR #156, opened 2025-07-29
144 move placo to daemon kin problems

askuric:
Hi @RemiFabre and @pierre-rouanet,

So.... 
It was more complicated than I hoped and this PR is not the _**final**_ solution but it is getting us there. 
There are two main issues:
1. Mujoco is pretty bad at the moment, very bouncy and a lot of ocilations, bringing the robot to some strange states
2. The forward and inverse kin solvers were not configured in the same way making their solutions a bit incompatible

Whet I've dione:
Rewritten the forward and inverse kin QP problems with placo 
    - removed some unnecessary constraints and regularisation terms
    -  added some new constratins like: 
         - the yaw angle between the head and the torso is limited to max 75deg (empirical  - can be cahnged )
         - the roll and pitch of the head are limited to 40deg ( empirical too - can be cahnged ) 
    - For this I had to modify the URDF of the robot, added a dummy joint and a link, the other option was to modify placo (maye in the future) 
    -  Tested the outputs with pinocchio and meshcat, it seems reasonable

What I did not do:
-  repair mujoco
   - at the momemtn we use mujoco as a physics simulator which we do not really need I think. It might be better to use it as a visualisator. No simulation (no need to have accurate motor models and the dynamics). 
    - In any case if we'll use mujoco as a simuator or a visualisator we need to repair the dynamcis. Make it a bit less oscilatory.

### DIFF
--- a/examples/debug/measure_drect_kin_error.py
+++ b/examples/debug/measure_drect_kin_error.py
@@ -7,6 +7,7 @@ import time
 import matplotlib.pyplot as plt
 import numpy as np
 from placo_utils.tf import tf
+from placo_utils.visualization import frame_viz, robot_frame_viz, robot_viz
 
 from reachy_mini.analytic_kinematics import ReachyMiniAnalyticKinematics
 from reachy_mini.placo_kinematics import PlacoKinematics
@@ -16,8 +17,12 @@ solver = PlacoKinematics(urdf_path, 0.02)
 robot = solver.robot
 robot.update_kinematics()
 
+viz = robot_viz(robot)
 
-# viz = robot_viz(robot)
+
+robot_limits = []
+for jn in solver.joints_names:
+    robot_limits.append(robot.get_joint_limits(jn))
 
 asolver = ReachyMiniAnalyticKinematics(urdf_path)
 
@@ -35,7 +40,7 @@ while i < 1000:
     # Read GUI values
     px, py, pz = [np.random.uniform(-0.01, 0.01) for _ in range(3)]
     roll, pitch = [np.random.uniform(-np.deg2rad(30), np.deg2rad(30)) for _ in range(2)]
-    yaw = np.random.uniform(-3.1, 3.1)
+    yaw = np.random.uniform(-2.8, 2.8)
     body_yaw = -yaw  # + np.random.uniform(-np.deg2rad(20), np.deg2rad(20))
 
     # Compute the target transformation matrix in the world frame
@@ -62,9 +67,10 @@ while i < 1000:
 
     # Measure time for solver FK
     start_solver_fk = time.time()
-    final_pose = solver.fk(joints)
+    for _ in range(2):
+        final_pose = solver.fk(joints)
     end_solver_fk = time.time()
-    solver_fk_time = end_solver_fk - start_solver_fk
+    solver_fk_time = (end_solver_fk - start_solver_fk)/2
 
     # Save times for later analysis
     solver_ik_times.append(solver_ik_time)
@@ -79,10 +85,26 @@ while i < 1000:
     rot_diff = R_final @ R_target.T
     angle_error = np.arccos(np.clip((np.trace(rot_diff) - 1) / 2, -1.0, 1.0))
 
-    if pos_error > 0.05 or angle_error > np.deg2rad(10):
-        # avoid outliers
+    joints = [robot.get_joint(jn) for jn in solver.joints_names]
+    skip = False
+    for j in range(7):
+        # Check if joint limits are respected
+        if not (robot_limits[j][0] < joints[j] < robot_limits[j][1]):
+            skip = True
+            break
+    if skip:
         continue
-
+        
+    if pos_error > 0.02 or angle_error > np.deg2rad(10):
+        viz.display(robot.state.q)
+        robot_frame_viz(robot, "head")
+        frame_viz("target", 
+                  solver.head_starting_pose @ 
+                    tf.translation_matrix((px, py, pz)) @
+                    tf.euler_matrix(roll, pitch, yaw))
+        print(px, py, pz, roll, pitch, yaw)
+        input("Press Enter to continue...")
+        
     position_errors.append(pos_error)
     angular_errors.append(np.degrees(angle_error))
 


RemiFabre:
Wow OK, this got much deeper than I thought.

I can have a look at this Thursday but if you're confident you can merge this PR (since it's not develop yet and 144 is broken anyway).

Just a question: I don't understand why these problems appeared now and not before? How does moving Placo daemon side change the way all of this behaves?

askuric:
Hi Remi, 

So the main difference is that we almost never called the fk of the robot in the develop. We only called it if the used really needed to know the pose of the head. Which was really not often. So we used the kinematics calculated by the ik (based on the same pinocchio model)
I've checked and if we run the ik and fk in the same loop in develop it brakes.
The reason why is:
- ik and fk use the same pinocchio model and modify its states both which in some cases results in braking the ik or fk resolution
- ik and fk do not have the same qp problems (different constraints and tasks) so when one modifies the model it can brake the other one.  

So in my experiments the FK seems to be less stable (and we did not see it before basically as we never called it). You can see the FK as the "physics" simulator. If everything is ok it will just accept the IK results and give you the pose of the head, but if the solution of the IK is not good (out of constraints) the fk will converge to another solution. So in develop we only ever used this "ideal" ik solution without passing it through the "physics" in the fk simulation. 

In the new branch we need to evaluate the fk all the time and ik only sometimes. So what I've gone here is:
- constrained the IK a bit more to resemble the physics of our robot with the aim that the IK solution stays within what's possible with the FK. 
- Updated the FK tasks that they provide the same weights for regularisation so that the FK and IK converge to the same JS solution
- and in the PR before, I've also used a separate pinocchio model (identical) for IK and FK so that they can have their own states and there is no exchange between them

This is not perfect, but

askuric:
Ok guys, 

I've done another small iteration over the code and I am pretty happy with the robustness. 
Here is what I've done:

1. Made the overshoots in the mujoco a bit more real (reduced) by adding a small kp gain

<details> <summary>
<b>Before video</b></summary>

https://github.com/user-attachments/assets/4a3adc00-57a9-4cdb-b9ac-719a2089ac0c

</details>
<details> <summary>
<b>After video</b></summary>

https://github.com/user-attachments/assets/82fc167f-cd51-4b95-a4a2-9d281b408211

</details>

2. Changed the weights of the head tracking task with respect to the body_yaw task
    - They had an equal weight before, I've set the tracking to be 5 times the weight of the body_yaw
    - This made a nice difference, making the robot a bit better in following a bit larger range of orientations
    - Basically it will leave the body yaw at the target value if possible (until 75 deg is reached), after that it will modify it in order to be able to attain the target value
    - This change makes a big difference in terms of stability, especially if the user sets a very unachievable body_yaw and the head target orientation. In those cases placo can converge to some strange results. This way it prioritizes the traking when this happens
    
<details> <summary>
<b>same weight</b></summary>


https://github.com/user-attachments/assets/fe7005de-5aa3-405e-a62e-9a5994622583



</details>
<details> <summary>
<b>traking weight > body_yaw weight</b></summary>

https://github.com/user-attachments/assets/a2e930a5-52ec-45cc-8815-90cadb0fbaec


</details>

RemiFabre:
Ok this is very interesting work. I have an idea on how to test it, I'll do this soon

RemiFabre:
Ok so I did some light testing and as far as I can tell this works well.

I'd like to test a bit more before merging into develop, probably Saturday morning. And provided a better test script for future regression testing (better than running dance_demo and looking at it).

What do you think needs to be done before merging to develop? What I see:
- General cleanup and CI
- I think we should integrate the limits we discussed now. If you agree can you do it? Like what you did for yaw but for all the task space dimensions (yes, there could be some weird problems for the rotations but I suspect that Euler angles might work because the allowed amplitudes might be low enough to avoid gimbal lock and such no? Or maybe do it in another way)
- Update API.md if needed (like for the allowed limits/new behaviors, but it's not needed to go too deep imo)

I can help with cleanup and such just tell me, I have time Saturday and Tuesday.

Thanks for your work!

askuric:
Hey @RemiFabre, 

Ok so I've done another iteration on the IK+FK and I am pretty happy. 

I did no go through the route of to implement the input filter, at the end I've just constrained the QP enough which sould be enough. 
So I was not able to break this version of the code, I was not able to bring it to some strange state with the sliders neither in `examples/debug/direct_gui_cs_ik.py` (pinocchio) or `examples/dance_demo.py` (mujoco). 
Let me know if you're able to do it :D

**The biggest change**
Now the FK and IK always start from the initial position instead of the current position (so in every iteration it re-optimises to the current pose). The initial zero position is very well placed and makes the optimisation to any other position realtively stable. If we use the current one we sometimes get stuck in positions angle +-2pi or so. 
I have not yet found a problem to this solution, it might come later, but for now it works great.  
This change (to start the FK and IK from the inital postion) does not change much as we do not use the FK or IK as a local optimiasiton (in which we would always unly execute only one iteration of it in each step) but we use it both already as the global ones (both of them call the QP 10 times -  effectively converging to the final pose/joint angles). 
So starting from the inital of the current position does not change much, the intial one is even better in some cases (we could even start form a random one :) ) 

EDIT
There are some small issues with the analytical kinematics, that's why the tests failing. I'm gonna update it soon.

askuric:
Ok @RemiFabre, 
- the analytical kinematics is updated
- the ruff and pytest passes
- the API.md is updated

Let me know if you want me to do some other stuff before the merge.

RemiFabre:
So I was plotting stuff as discussed this morning, but I kept having weird low frequencies for the current head pose. 
image
The numbers on top are the period of the main loop daemon side (Mujoco), the numbers are quite variable and the period is occasionally above the goal of 5ms (on my laptop that has a very decent CPU, I'll ask Alina to test this on the rasp).

But the thing that I don't understand is the numbers on the bottom. It's the period between receptions of the present head pose client side, it's around 22Hz. This matches my plots.

So the daemon publishes at 200Hz and the client is updated at 22hz, any idea what is going on?

RemiFabre:
It looks like this :'(

groovy_sway_and_roll_xyzrpy_vs_time

RemiFabre:
So the previous charts were done in simulation. This one is on the physical robot:
side_to_side_sway_xyzrpy_vs_time
So the problem seems Mujoco specific, which is weird because the control loop seems fine on the Mujoco implementation

RemiFabre:
Ok so I though that this would immediately publish:
image
But apparently it's more complicated than that. Seems like a mission for @pierre-rouanet :)

Ok so:
- The current state works well and it's much more robust than develop
- We have 4 branches that start with 144
- There is a known problem in simulation, but it's not critical (the simulation is perfectly usable as is, as a matter of fact we didn't see the problem before plotting this)

My call would be: we merge this and open an issue.

Let's raise hands.

:hand: 

Excellent. Unanimous vote.

docs/API.md
```diff
@@ -168,6 +168,58 @@ with ReachyMini() as reachy:
         time.sleep(0.01)
 ```
 
+⚠️ <b>BEWARE: Mini's head range of motion is limited</b> ⚠️
```
RemiFabre, line 171: Very good!

examples/dance_demo.py
```diff
@@ -40,7 +40,7 @@ class Config:
     beats_per_sequence: int = 8  # Switch move every 8 beats
     start_move: str = "simple_nod"
     amplitude_scale: float = 1.0
-    neutral_pos: np.ndarray = field(default_factory=lambda: np.array([0, 0, 0.02]))
```
RemiFabre, line 43: Tested all the moves with this setup, in simu and real. This was causing issues, now it works fine.

src/reachy_mini/placo_kinematics.py
```diff
@@ -42,7 +43,7 @@ def __init__(
 
         # they should be hard but we use soft to avoir singularities and
         # tradeoff the precision for the robustness
-        constrant_type = "soft"  # "hard" or "soft"
+        constrant_type = "hard"  # "hard" or "soft"
```
RemiFabre, line 47: Can you double check this is what you want? It's inconsistent with the above comment

src/reachy_mini/placo_kinematics.py
```diff
@@ -70,40 +97,36 @@ def __init__(
         self.head_starting_pose = np.eye(4)
         self.head_starting_pose[:3, 3][2] = self.head_z_offset
         self.head_frame = self.ik_solver.add_frame_task("head", self.head_starting_pose)
-        self.head_frame.configure("head", "soft", 1.0, 1.0)
+        # equivalance to ~1cm = 1deg weights
```
RemiFabre, line 101: Straight to jail, everyone knows that it should be 1mm = 1deg  (I'm joking, we'll tune this later if needed)

askuric:
OK so I've spent some time analysing the execution time of the proposed IK/FK approach and the error due to always starting from the initial position. 

Here is what I've done:
- Took 1000 random (I'd say almost all reachable) poses of the head + body yaw 
- Checked the execution time of the FK and IK
- Checked the error of the IK+FK solution with respect to the distance (postion, angle) between the target and the initial postion
    - comparing different numbers of iterations (default 10)
    - comparing different max velocities of the robot (default 10rad/s) 

Run the code with 
```
python examples/debug/measure_drect_kin_error.py
```    


### iterations: `10`  + max velocity: `10` rad/s
10_it_time_exec
10_it_error_from_initial

### iterations: `15`  + max velocity: `10` rad/s
15_it_error_dist
15_it_error_from_initial



### iterations: `20`  + max velocity: `10` rad/s
20_it_time_exec
20_it_error_from_initial


### iterations: `10`  + max velocity: `20` rad/s
10_it_time_exec_mv_20
10_it_error_from_initial_mv_20


### Other things I've tested

I've setting the max number of iterations of the ik/fk. Setting the max  number of iterations to for example 20 and then stopping when either the ik/fk have converged or if the max number of iteration is reached. 
The execution time is a bit more variable, and the results are pretty good.
max20_it_time_exec
max20_it_error_from_initial




### Result discussion

- The 10 iteration solution has a relatively considerable error of about 2 degree in worst case, for the furthest distances from the initial position
- The 15 is a lot better and 20 is almost perfect. But they come with a linear computational cost increase (15 is 1.5x longer and 20 is 2x longer)
- Increasing the max velocity of the robot twice (from 10rad/s to 20 rad/s in the urdf),  improves the FK/IK precision significantly. Maing the 10 iteration IK/FK be as good as 20 iteration ones.  The donwside is that the velcoity does no longer represent the physical velocity limits of the robot.
- The adaptive solution (until converges or until max iterations reached) is great in terms of precision and does not require the change in the URDF. But it makes the IK/FK execution time less deterministic. But on average the execution time corresponds to `15` iteration version

So for the possible steps forward are:
- Either we are fine with this 2deg of error and we work on it later, maybe increase to `15` or `20` iterations 
- Either we dont care about the urdf and we update the velocity to 20rad/s and we leave the code as is
- Or we go full on with the convergence based adaptive method and accept the small variability in the execution time

RemiFabre:
These measures are on point :+1: 

Question: why do we care that the inner velocity limits used during the steps are close to the physical limits? if I remember correctly, the main limit with high speeds is not getting into collision with the limits too soon no?

askuric:
Yep, thats one reason and the other is the local linearisation. If the velocity is too high the local linarisation in the QP could potentially lead to oscilations. 
But in our case I think that we are not reaching this point with the speeds 10-20rad/s

RemiFabre:
I think the current need is to reduce computational needs rather than the other way around, I think it's worth it to try to stay at 10 iterations and try rhe other stuff.
- I think we can increase the max speed as long as no practical problems appear, just make it clear through comments that these values are numerical optimization variables and not a physical representation
- in your graphs the angle errors are quite low but the position errors sometimes are quite big. Isn't this indicative that the weights on tasks should be retuned? (The 1cm == 1 deg vs 1 mm == 1 deg discussion). Or are these unreachable poses that should be ignored?
- the adaptative method seems good too, maybe just add limits to avoid edge cases?

These plots are really good, the script should definitely be pushed and maybe we could do a small blog post on this stuff later down the road?

RemiFabre:
This is what I got on my laptop (almost 4year old hardware but powerful CPU)

Figure_1
Figure_1_2
Figure_1_3

RemiFabre:
On my machine I get results almost 5 times faster than you, and with much less variance. Are you sure you ran these graphs with the same softer I did (no local changes?).

We should run this on a rasp to have better guidelines. Maybe Matthieu will confirm the final hardware version this afternoon?

I still believe we should merge this soon and create issues to optimize this stuff after the merge because other team members are working on develop on things that might be a tad annoying to merge.

askuric:
Ok so the velocities in the URDF are artificially inflated to 20rad/s instead of 10rad/s that is conisdent with the datasheet (maybe still a bit too high). 

These are the final plots (10 iterations) 
final
final1
final2

>On my machine I get results almost 5 times faster than you, and with much less variance. Are you sure you ran these graphs with the same softer I did (no local changes?).

I think it's just my PC that is bad, I've got the CPU: Intel(R) Core(TM) i5-3230M CPU @ 2.60GHz (from 2013)
https://cpu.userbenchmark.com/Intel-Core-i5-3230M/Rating/2772

RemiFabre:
groovy_sway_and_roll_xyzrpy_vs_time
I ran the measure_tracking.py script again on all dances. Results seem very similar to before the changes.
For example the above graphs are to be compared to these ones:
https://github.com/pollen-robotics/reachy_mini/pull/156#issuecomment-3155691325

askuric:
Ok I've done one more pass. I think that the 20rad/s is a bit too much at the end. My final proposal is this:
- Inrease the max  speed  to `13rad/s` 
    - in the software not in the URDF (so we can easily update it later) 
- Increase the number or iterations to `12`

This  seems to be a sweetspot. 
- Its a lot more stable, especially around 180 degree rotations in yaw. 
- Even though this is not really the best use-case.  

final
final1
final2

RemiFabre:
Deal!

PR #164, opened 2025-08-07
Adding a mechanism to record set_target commands

RemiFabre:


src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -194,6 +196,37 @@ def set_target_head_joint_current(self, current: List[int]) -> None:
...
+            record (dict): A dictionary containing the record data to be appended.
+
+        """
+        if self.is_recording:
+            self.recorded_data.append(record)
+
+    def start_recording(self) -> None:
+        """Start recording data."""
+        self.is_recording = True
+        self.recorded_data = []
```
FabienDanieau: Would it be better to init the array before self.is_recording = True?

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -194,6 +196,37 @@ def set_target_head_joint_current(self, current: List[int]) -> None:
...
+
+    def stop_recording(self) -> None:
+        """Stop recording data and publish the recorded data."""
+        self.is_recording = False
+        recorded_data = self.recorded_data.copy()
```
FabienDanieau: is the copy necessary here?

src/reachy_mini/reachy_mini.py
```diff
@@ -533,6 +544,23 @@ def _set_head_pose(self, pose: np.ndarray, body_yaw: float = 0.0) -> None:
...
+        """Start recording data."""
+        self.client.send_command(json.dumps({"start_recording": True}))
+        self.is_recording = True
+        print("Recording started. Data will be logged until stop_recording is called.")
+
+    def stop_recording(self) -> List[Dict]:
+        """Stop recording data and return the recorded data."""
+        self.client.send_command(json.dumps({"stop_recording": True}))
+        self.is_recording = False
+        print("Recording stopped. Retrieving recorded data...")
```
FabienDanieau: use a logger instead print

RemiFabre: Removed the prints, they are not very useful

src/reachy_mini/daemon/backend/abstract.py
RemiFabre: Good catch, it's a bad way to handle the potential race condition. I pushed a version with a lock. Also there was some typos in the previous code I fixed with this commit

PR #172, opened 2025-08-13
Feat calculate ik at control frequency

askuric:
OK this PR is still a draft. 

The IK is not calculated on each call of the server's setting the head pose target but in the control loop only if required. 
This can significantly reduce the computational cost of the IK.

Especially in situations where the user spams the client with new poses all the time.

### DIFF
--- a/src/reachy_mini/placo_kinematics.py
+++ b/src/reachy_mini/placo_kinematics.py
@@ -22,7 +22,7 @@ class PlacoKinematics:
         urdf_path: str,
         dt: float = 0.02,
         automatic_body_yaw: bool = False,
-        log_level: str = "INFO"
+        log_level: str = "INFO",
     ) -> None:
         """Initialize the PlacoKinematics class.
 
@@ -33,8 +33,10 @@ class PlacoKinematics:
             log_level (str): Logging level for the kinematics computations.
 
         """
-        self.fk_reached_tol = np.deg2rad(0.1)  # 0.1 degrees tolerance for the FK reached condition
-        
+        self.fk_reached_tol = np.deg2rad(
+            0.1
+        )  # 0.1 degrees tolerance for the FK reached condition
+
         self.robot = placo.RobotWrapper(urdf_path, placo.Flags.ignore_collisions)
         self.robot_ik = placo.RobotWrapper(urdf_path, placo.Flags.ignore_collisions)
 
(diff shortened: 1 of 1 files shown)


askuric:
New iteration for performance improvement:
Stuff changed:

1. The mujoco does not need us to calculate the FK, it already does it internally so I just used mujoco and removef the FK completely. So now we only run the IK in the loop if the new command is sent. 
2. Improved the perfomance of the FK. The FK runs a 2 iterations by default and starts from the current position. 
    - If it ends up in some strange state I allow it to retry from the inital state with 5 iterations (so max is 7)
    - Typically when the FK starts from the inital state it converges in 99% (did not really quantify it but its close to 100%) of cases so after the retry we are more or less certain that the FK is going in converging
    - Finally, as the real robot's states are continuous (measured in real time) the FK is never going to have to converge to some very far value from the inital state (this is not the case for IK)
  
IMPORTANT: 
IK has stayed the same for the moment.

Voila les pers pour moi:
image

askuric:
New iteration

- This time both IK and FK are modified a lot
- Improved the precision and the execution time

## IK
- does the local optimisaiton
   - starting from its current configuration
   - taking 2 steps per call (can maybe be lowered to 1, did not try)
- If the current position and the taget position are within 0.1mm and 0.1deg the IK is no longer calculated
- If the new target pose it too far from the current one (more than 180 degrees rpy compound) the optim will start from the initial position rather than from the current position
- If the QP fails for some reason the IK will restart form the initial position with a bit more iterations in this call  (added 6 - completely empirically) 
- number of iterations
  - normal operation: 2
  - new target pose to fat: 8 (2+ 6)
  - worst case (failed QP): 10 (2 + 8) 
  - now a variable (`no_iterations`) so it can be modified by the user 

## FK
- does the local optimisation 
   - starting from its current configuration
   - taking 2 steps per call (can maybe be lowered to 1, did not try)
- If the QP fails for some reason the FK will restart form the initial position with a bit more iterations in this call  (added 6 - completely empirically) 
- number of iterations
  - normal operation: 2
  - worst case (failed QP): 10 (2 + 8) 
  - now a variable (`no_iterations`) so it can be modified by the user 
   

error_distance
time_ik_fk

askuric:
The latest version of the code has even more restricted number of iterations:

### IK:
- normal operation: 2
- worst case (failed QP): 6 (normal operation + 4)
- if the IK converged: 0 

### FK
- normal operation: 2
- worst case (failed QP): 6 (normal operation + 4)
- if the FK converged: 0 


### Execution times
image

Total iterations: 2000
IK time: 0.55 ms (max 4.24 ms)
FK time: 0.33 ms (max 5.21 ms)

PR #182, opened 2025-08-26
Integrate NN kinematics

apirrone:
Working NN kinematics (still needs to be thoroughly tested)

We can also now choose the kinematics engine (Placo, NN, Analytical) from the daemon cli. (Analytical is not supported yet)

### DIFF
--- a/examples/debug/compare_recordings.py
+++ b/examples/debug/compare_recordings.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3
 # compare_recordings.py
-"""
-Compare two ReachyMini dance-run recordings by regenerating overlay plots.
+"""Compare two ReachyMini dance-run recordings by regenerating overlay plots.
 
 Usage
 -----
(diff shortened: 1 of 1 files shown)


RemiFabre:
Added a test script that compares errors between runs, the results between Placo and NN are very similar. Based on just watching the dance moves, every thing seemed identical. But looking at the charts there are small differences like:
image
image

Be careful when looking at these charts, if you see "big changes" it's probably that the Y scale is zoomed in for moves where that coordinate almost doesn't move


Generated with:
reachy_mini/examples/debug$ python compare_recordings.py measures/2025-08-27_16-04-16 measures/2025-08-27_15-30-30

RemiFabre:
The only "big" problem we see is the yaw+body_yaw management that is done "manually" in this version and that sometimes explodes or has a different behavior than the Placo implementation. Antoine will probably retrain with an extra dimension. But other than that, this is very promising work. And it's super fast. On Antoine's computer:

image

apirrone:
Thanks for your review @RemiFabre. 

I'm keeping this PR open until we have a satisfactory solution for handling the body yaw. I'm working on it

PR #185, opened 2025-08-28
Recorded Moves API

apirrone:
# TODO:

- [ ] Make moves interruptible
- [x] Migrate record script
- [x] Make library dataset script (aggregate recorded moves and upload to HF)

### DIFF
--- a/examples/debug/is_realtive_tests/test_combined_absolute_relative.py
+++ b/examples/debug/is_realtive_tests/test_combined_absolute_relative.py
@@ -15,7 +15,7 @@ import threading
 import time
 
 from reachy_mini import ReachyMini, utils
-from reachy_mini.motion.dance_move import DanceMove
+from reachy_mini.motion.dance.dance_move import DanceMove
 
 
 def absolute_rectangle_thread(
--- a/examples/debug/is_realtive_tests/test_play_on_relative.py
+++ b/examples/debug/is_realtive_tests/test_play_on_relative.py
@@ -14,7 +14,7 @@ import threading
 import time
 
 from reachy_mini import ReachyMini, utils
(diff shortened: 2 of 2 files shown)


src/reachy_mini/motion/dance/dance_move.py
```diff
@@ -60,7 +61,7 @@ def evaluate(self, t: float) -> tuple[np.ndarray, np.ndarray]:
         head_pose = create_head_pose(
             x=x, y=y, z=z, roll=roll, pitch=pitch, yaw=yaw, degrees=False, mm=False
         )
-        return head_pose, offsets.antennas_offset
+        return head_pose, offsets.antennas_offset, 0, None
```
RemiFabre, line 64: Please update the docstrings too, this new "0, None" output is a bit obscure to me

src/reachy_mini/motion/recorded/recorded_move.py
```diff
@@ -0,0 +1,146 @@
...
+        self.hf_dataset_name = hf_dataset_name
+        self.local_path = snapshot_download(self.hf_dataset_name, repo_type="dataset")
+        self.moves = {}
+        self.sounds = {}
+
+        self.process()
+
+    def process(self):
+        """Populate recorded moves and sounds."""
+        move_paths = glob(f"{self.local_path}/*.json")
```
RemiFabre, line 34: Where does this end up being saved?

src/reachy_mini/motion/dance/dance_move.py
RemiFabre: I assume it's "body_yaw, sound", correct?

PR #193, opened 2025-09-01
Making this repo "core"

apirrone:
Removing stuff that we don't want as core features, reducing dependencies etc


apirrone:
@pierre-rouanet if that's ok with you we can merge this pretty quickly

PR #196, opened 2025-09-02
192 make move interruptible

pierre-rouanet:
Make goto and moves play_on async/sync. This open the possibility to spawn the async version as tasks and cancel them.

A cleaner/simpler API which hide the async could be defined later.
Later cleaning the ReachyMini API would also be good.


apirrone:
~~LGTM~~
Actually LGTM

PR #200, opened 2025-09-03
199 cleanup motor mode

pierre-rouanet:
I propose we simplify the way we control motors and only exposes 3 basic mode:
- motor disabled (torque off)
- motor enabled (torque on and controlled in position)
- gravity compensation

More advanced mode can be added later if we think they will be useful.

It requires to update reachy-mini-motor-controller to 0.6 (now the starting control mode of the robot is checked, before it was assumed it was disabled).


apirrone:
LGTM

PR #206, opened 2025-09-03
Rework goto_sleep() (daemon side)

apirrone:
- If we are already very close to the sleep position, we do nothing.
- If we are far from the sleep position:
    - If we are far from the initial position, we move there first.
    - Elsewe move directly to the sleep position.

### DIFF
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -714,6 +714,11 @@ class Backend:
             self.get_current_head_pose(), self.INIT_HEAD_POSE
         )
 
+        print("dist to sleep pose:", dist_to_sleep_pose)
+        print("dist to init pose:", dist_to_init_pose)
+
+        sleep_time = 2.0
+
         # Thresholds found empirically.
         if dist_to_sleep_pose > 10:
             if dist_to_init_pose > 30:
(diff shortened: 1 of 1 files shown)


src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -698,36 +699,38 @@ async def wake_up(self) -> None:
...
+            self.get_current_head_pose(), self.INIT_HEAD_POSE
         )
-        if dist > 0.2:
-            await self.async_goto_target(
-                self.INIT_HEAD_POSE, antennas=[0.0, 0.0], duration=1
-            )
-            await asyncio.sleep(0.2)
 
-        # Pfiou
-        self.play_sound("go_sleep.wav")
```
RemiFabre, line 723: imo we should play the sound no matter what, it becomes a clear cue that we closed the daemon. On Reachy2 we got used to sound cues and it's a decent state change indicator

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -698,36 +699,38 @@ async def wake_up(self) -> None:
         await self.async_goto_target(self.INIT_HEAD_POSE, duration=0.2)
 
     async def goto_sleep(self) -> None:
```
RemiFabre, line 701: Also I wonder if we should disable torque on sleep and enable it on wake up?  No strong opinions here

src/reachy_mini/daemon/backend/abstract.py
apirrone: done

src/reachy_mini/daemon/backend/abstract.py
apirrone: No strong opinion here neither, what do you think @pierre-rouanet ?

pierre-rouanet: It's done on the daemon side. I kind a think it makes more sense this way but no strong opinion.

apirrone: Ok let's keep it that way for now, we'll see if we need it in the future

PR #214, opened 2025-09-05
Add daemon initialization in pytest

haixuanTao:
Adding daemon initialization in pytest

### DIFF
--- a/examples/debug/bench_cpp_kin.py
+++ b/examples/debug/bench_cpp_kin.py
@@ -1,0 +1,32 @@
+import time  # noqa: D100
+
+import numpy as np
+
+from reachy_mini.kinematics import CPPAnalyticKinematics
+
+cpp_kin = CPPAnalyticKinematics(
+    urdf_path="../../src/reachy_mini/descriptions/reachy_mini/urdf/robot.urdf"
+)
+
+ik_times = []
+fk_times = []
+
+pose = np.eye(4)
+t0 = time.time()
+for i in range(1000):
+    t = time.time() - t0
+    pose[:3, 3][2] += 0.1 * np.sin(2 * np.pi * 0.5 * t)
+
+    ik_t0 = time.time()
+    joints = cpp_kin.ik(pose)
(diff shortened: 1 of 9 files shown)


haixuanTao:
It seems that we can't run the daemon within the CI/CD. I thought it was linked to mujoco but couldn't fix it. I don't know if there's a known quick fix for this?

pierre-rouanet:
Have you played with headless Mujoco already @apirrone?

apirrone:
> Have you played with headless Mujoco already @apirrone?

~~Never~~ Only with MJX in Mujoco playground, I can take a look if you want

apirrone:
In theory it should be as simple as not launching the viewer and not syncing it. For example, this : 

```python
import mujoco
import time


model = mujoco.MjModel.from_xml_path(
    "/home/antoine/Pollen/reachy_mini/src/reachy_mini/descriptions/reachy_mini/mjcf/scenes/empty.xml"
)
model.opt.timestep = 0.002
decimation = 10
data = mujoco.MjData(model)

step = 0
while True:
    s = time.time()
    if step % decimation == 0:
        # control
        pass

    mujoco.mj_step(model, data)
    step += 1

    took = time.time() - s
    time.sleep(max(0, model.opt.timestep * decimation - took))

```
Runs in a headless ssh session on a remote computer.

We could add an option to run the daemon in headless mode for simulation ?

apirrone:
@haixuanTao  I added a `--headless` argument when launching the daemon. This will run mujoco in headless mode. Let me know if it works for you !

haixuanTao:
Yeah there's a weird error in the CI that just crash and I don't know why. I have tried to reach out to lerobot folks as they have a mujoco pipeline running but it does not seems to be this. Maybe something else like memory or access to the camera

apirrone:
I forgot to remove the offscreen renderer when in headless mode, should work now @haixuanTao  !

haixuanTao:
I'll let you merge if that's ok with your workflow!

PR #215, opened 2025-09-05
211 remove dances from core

RemiFabre:


RemiFabre:
I removed the symbolic dances, removed the symbolic dance example, updated the minimal_dance_player with Antoine’s dataset, and updated API.md

PR #217, opened 2025-09-08
207 move rework and simplification

pierre-rouanet:
I think this is ready for merge. Do you see something missing @apirrone?


apirrone:
LGTM

PR #221, opened 2025-09-15
Serve web api with daemon

pierre-rouanet:
Modify reachy-mini-daemon entry point so on top of previous feature it now also serves a web API.

* List of routes can be found at: [http://0.0.0.0:8000/docs/](http://0.0.0.0:8000/docs/)
* wip dashboard at [http://0.0.0.0:8000/](http://0.0.0.0:8000/)

It should not break anything on the python client side. @apirrone Can you check your dev app is still working?

I did not test much performance on real conditions, yet. Doc is still missing as well.
But if it's ok, I would prefer to merge it as soon as possible to see the problem sooner than later.

### DIFF
--- a/src/reachy_mini/kinematics/analytical_kinematics.py
+++ b/src/reachy_mini/kinematics/analytical_kinematics.py
@@ -46,7 +46,7 @@ class AnalyticalKinematics:
                 1 if motor["solution"] else -1,
             )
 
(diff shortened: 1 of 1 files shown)


apirrone:
I launched our conversation demo with this branch, seems to work exactly the same

Issue #242, opened 2025-09-30
Invalid number of channels with sound_play.py


FabienDanieau:
sounddevice.PortAudioError: Error opening OutputStream: Invalid number of channels [PaErrorCode -9998]

Issue #237, opened 2025-09-23
librosa.load failed when using mjpython


pierre-rouanet:
When running via mjpython (for mujoco rendering on macOS), librosa fails to load. I'm assuming it's due to the fact that my vent python and the system one are not the same version.

Full track:
```python
In [6]: await d.backend.wake_up()
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File ~/.venvs/reachy-mini/lib/python3.13/site-packages/llvmlite/binding/ffi.py:141, in _lib_wrapper.__getattr__(self, name)
    140 try:
--> 141     return self._fntab[name]
    142 except KeyError:
    143     # Lazily wraps new functions as they are requested

KeyError: 'LLVMPY_AddSymbol'

During handling of the above exception, another exception occurred:

OSError                                   Traceback (most recent call last)
File ~/.venvs/reachy-mini/lib/python3.13/site-packages/llvmlite/binding/ffi.py:122, in _lib_wrapper._load_lib(self)
    119 with _suppress_cleanup_errors(_importlib_resources_path(
    120         __name__.rpartition(".")[0],
    121         get_library_name())) as lib_path:
--> 122     self._lib_handle = ctypes.CDLL(str(lib_path))
    123     # Check that we can look up expected symbols.

File /opt/homebrew/Cellar/python@3.13/3.13.7/Frameworks/Python.framework/Versions/3.13/lib/python3.13/ctypes/__init__.py:390, in CDLL.__init__(self, name, mode, handle, use_errno, use_last_error, winmode)
    389 if handle is None:
--> 390     self._handle = _dlopen(self._name, mode)
    391 else:

OSError: dlopen(/Users/pierrerouanet/.venvs/reachy-mini/lib/python3.13/site-packages/llvmlite/binding/libllvmlite.dylib, 0x0006): Library not loaded: @rpath/libz.1.dylib
  Referenced from: <C7E9C606-AA76-3C22-9E01-9EBD7DCC0C28> /Users/pierrerouanet/.venvs/reachy-mini/lib/python3.13/site-packages/llvmlite/binding/libllvmlite.dylib
  Reason: tried: '/opt/homebrew/lib/libz.1.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/lib/libz.1.dylib' (no such file), '/libz.1.dylib' (no such file)

The above exception was the direct cause of the following exception:

[...]

PR #243, opened 2025-09-30
bug #242: auto clean up audio output device

FabienDanieau:
The daemon plays a sound during the wake up and goto sleep procedure. But it's actually the client (ReachyMini) that will directly use the mic. 
The issue was that the audio interface was considered already used by sounddevice. The daemon is now properly releasing the audio device when done playing.


FabienDanieau:
Antoine's issue is related to librosa. Let's discuss it in this issue #237

Issue #133, opened 2025-07-08
Camera open by default on 1080p while image_size is defined as 720p in reachy_mini.py


pierre-rouanet:

FabienDanieau:
MacOs opens the camera with a higher resolution than 1280 which is hardcoded [here](https://github.com/pollen-robotics/reachy_mini/blob/5ef1e84c13fcb4f4e42308de6b72667ff00e61bd/src/reachy_mini/reachy_mini.py#L311)
cc @CarolinePascal and @alozowski

CarolinePascal:
Using OpenCV as media backend, you can [specify the resolution and fps on startup](https://github.com/huggingface/lerobot/blob/c378a325f05a652d0d3808f76d18e9e889de93c2/src/lerobot/cameras/opencv/camera_opencv.py#L228). 
I genuinely think it works on Linux out of luck, as we never provide the desired resolution to the camera in the codebase 😅 Same goes for the fps value.

Using GStreamer as media backend, the resolution and fps are already specified on startup 👍 

I can create a PR porting LeRobot `validate` functions in Reachy Mini codebase - Lmk.

CarolinePascal:
Up @FabienDanieau if you have time 👀

PR #245, opened 2025-10-01
enhancement #133: setting the camera resolution. adding unit tests.

FabienDanieau:
To test with a plugged robot:
`pytest -m "video"`

robot + daemon
`python examples/look_at_image.py`


alozowski:
@FabienDanieau it works 👍🏻

PR #246, opened 2025-10-02
244 expose direction of arrival

FabienDanieau:
I've integrated Jerry's code to get the DoA and the speech detection.

unit test
`pytest -m audio`

The example script is:
`python examples/debug/sound_doa.py`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ dependencies = [
     "asgiref",
     "aiohttp",
     "log-throttling==0.0.3",
+    "pyusb>=1.2.1",
 ]
 
 


examples/debug/sound_doa.py
```diff
@@ -0,0 +1,56 @@
...
+        level=logging.DEBUG, format="%(asctime)s [%(levelname)s] %(message)s"
+    )
+
+    with ReachyMini(log_level="DEBUG", automatic_body_yaw=True) as mini:
+        last_doa = -1
+        THRESHOLD = 5.0  # degrees
+        while True:
+            doa = mini.media.audio.get_DoA()
+            print(f"DOA: {doa}")
+            if doa[1] and np.abs(doa[0] - last_doa) > THRESHOLD:
```
CarolinePascal, line 28: Not sure the DoA difference threshold is relevant : there might be a mobile source that rotates around Reachy Mini such that you always have the same DoA even though Reachy Mini turns to catch it up !

examples/debug/sound_doa.py
```diff
@@ -0,0 +1,56 @@
...
+                print(f"  Speech detected at {doa[0]:.1f}°")
+                p_head = [1.0, 1 - doa[0] / 90.0, 0.0]
+                print(
+                    f"  Pointing to x={p_head[0]:.2f}, y={p_head[1]:.2f}, z={p_head[2]:.2f}"
+                )
+                T_world_head = mini.get_current_head_pose()
+                R_wc = T_world_head[:3, :3]
+                p_world = R_wc @ p_head
```
CarolinePascal: Nit picking : R_wc -> R_world_head

examples/debug/sound_doa.py
```diff
@@ -0,0 +1,56 @@
...
+                print(
+                    f"  In world coordinates: x={p_world[0]:.2f}, y={p_world[1]:.2f}, z={p_world[2]:.2f}"
+                )
+                mini.look_at_world(*p_world, duration=0.5)
+                last_doa = doa[0]
+            else:
+                if not doa[1]:
+                    print("  No speech detected")
```
CarolinePascal, line 44: Should we slowly come back to the default pose if no speech is detected ?

src/reachy_mini/media/audio_base.py
```diff
@@ -59,3 +77,64 @@ def push_audio_sample(self, data):
...
+            result = response.tolist()
+        elif data[4] == "radians":
+            byte_data = response.tobytes()
+            float1, float2, float3, float4 = struct.unpack("<ffff", byte_data[1:17])
+            result = [
+                np.rad2deg(float1),
+                np.rad2deg(float2),
+                np.rad2deg(float3),
+                np.rad2deg(float4),
+            ]
```
CarolinePascal, line 136: Should we rather keep radians ?

examples/debug/sound_doa.py
FabienDanieau: this is in the case that when the sound source is stationary, the DoA is not precise so the head is always adjusting its position

FabienDanieau:
stand by. waiting for Jerry's review.

src/reachy_mini/media/audio_base.py
FabienDanieau: done

examples/debug/sound_doa.py
FabienDanieau: this is just a debug script, it's a good idea for an actual example

PR #260, opened 2025-10-08
Make move cancellable

haixuanTao:
So, it might be useful to cancel move on interruption which is made possible by this PR


KevinMacDermid:
I'd like this functionality as well, is there a way for me to help this get merged?

PR #264, opened 2025-10-09
bug #237: remove librosa, use scipy

FabienDanieau:
scipy can resample. It may be more naive than a dedicated audio lib though. 

@apirrone, does it fix your issue? launching 
`reachy-mini-daemon --sim`


apirrone:
Yes it works !

Issue #248, opened 2025-10-03
Handle onboard Pi case for the serial port


pierre-rouanet:
The serial port used for the onboard CM4 version will be:
- /dev/ttyAMA3

But no information is set in the ListPortInfo instance.

PR #274, opened 2025-10-13
180 2 stream audiovideo from rpi

FabienDanieau:
Stream audiovideo from reachy mini RPI. start deamon with:
`GST_DEBUG=3 reachy-mini-daemon --wireless`

test with:
`pytest -m wireless_gstreamer`

or 
`python examples/look_at_image.py --backend gstreamer`

Note that gstreamer is mandatory on the client side for now. The aiortc client will be provided don't worry.
Audio client to daemon will be addressed in another issue as well.

Edit:
The basic examples are now functional
```
python examples/debug/sound_play.py --backend gstreamer
python examples/debug/sound_record.py --backend gstreamer
```


FabienDanieau:
wait for #248 and use wireless arg instead of webrtc
Edit: ok

PR #293, opened 2025-10-15
94 rerun support or similar 2

FabienDanieau:
There is a new class in utils to launch a rerun instance and start logging.
See example here, it's based on the compliant one. Manually the real robot will move it in rerun.
`python examples/rerun_viewer.py`

The placo engine is required:
`reachy-mini-daemon  --kinematics-engine Placo [--sim]`

For now I'm using the local URDF file. I assume that rerun is mostly used by dev who have the local repo.
Maybe we'll need to ship the URDF in the future pip package. Another option is to get URDF (already there) and STL from the REST API.
What do you think?


pierre-rouanet:
Really cool! Works like a charm on my Mac.

- The URDF is already in the pip package, no?
- For the STL, I'll definitely add an API route! #318

PR #305, opened 2025-10-16
Fixed the gravity compensation on the beta version

askuric:
The gravity compensation did not work as well on betas as for alphas. 

I am not 100% sure what the real reason is but the fix is the increase in the torque constant correction factor from 3.0 to 4.0. 

Tested on the reachy mini at Hugging Face's Paris office.


apirrone:
Works great !

PR #307, opened 2025-10-16
bug #306: soundevice fails if respeaker is selected in system default

FabienDanieau:
Now, if the respeaker is available it will be used. If not, because it selected by the system, we fallback to default that uses the respeaker in the end.

test:
```
python examples/debug/sound_record.py 
```
then play back recorded_audio.wav

and
```
python examples/debug/sound_play.py 
```


apirrone:
It works !

PR #314, opened 2025-10-18
fix(NoBackendError): adding clean handling of NoBackendError when looking for the ReSpeaker

CarolinePascal:
Making the error message clearer to help the user identify the cause of the error.

### DIFF
--- a/src/reachy_mini/media/audio_base.py
+++ b/src/reachy_mini/media/audio_base.py
@@ -90,10 +90,15 @@ class AudioBase(ABC):
 
     def _init_respeaker_usb(self) -> Optional[usb.core.Device]:
         try:
-            dev = usb.core.find(idVendor=0x2886, idProduct=0x001A, backend=get_libusb1_backend())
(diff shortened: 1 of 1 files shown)


src/reachy_mini/media/audio_base.py
```diff
@@ -88,12 +88,17 @@ def play_sound(self, sound_file: str) -> None:
...
     def _init_respeaker_usb(self) -> Optional[usb.core.Device]:
-        dev = usb.core.find(idVendor=0x2886, idProduct=0x001A)
-        if not dev:
+        try:
+            dev = usb.core.find(
+                idVendor=0x2886, idProduct=0x001A, backend=get_libusb1_backend()
+            )
+            return dev
+        except usb.core.NoBackendError:
+            self.logger.error(
```
FabienDanieau: Just changed this raise to a log.error. We don't have to stop everything if it fails here.
Also I return None, so the self._repseaker attribute exists and the __del__ doesn't fail

PR #320, opened 2025-10-20
Add route for stl files.

pierre-rouanet:
Add a route to get stl files directly from the API.


8bitkick:
It works for me - see https://huggingface.co/spaces/8bitkick/reachy_mini_3d_web_viz

PR #322, opened 2025-10-20
Cleanup and organize .gitignore

alozowski:
A small PR to add venv to gitignore and organise it a little bit better

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,30 +1,255 @@
-# Python
+# Byte-compiled / optimized / DLL files
 __pycache__/
-*.egg-info/
-src/*.egg-info/
-build/*
-dist/*
+*.py[codz]
+*$py.class
 
-# Virtual environments
-.venv/
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,7 +1,13 @@
+# Python
 __pycache__/
-reachy_mini.egg-info/
 *.egg-info/
+src/*.egg-info/
 build/*
 dist/*
+
+# Virtual environments
+/.venv/
```
FabienDanieau: is the first / on purpose?
We have this in [conversation app](https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/4268afb490f8889c7c52900dce8cb592d5e5703a/.gitignore#L7)

```
.venv/
venv/
ENV/
env/
```

alozowski: I think you're right, let's use a more defensive I approach from the conversation app:
```
.venv/
venv/
ENV/
env/
```

alozowski:
Took [this official .gitignore](https://github.com/github/gitignore/blob/main/Python.gitignore), and extended it with project-specific ignores at the end. It should cover different platforms and IDEs.

PR #333, opened 2025-10-23
Fix for audio default device

andimarafioti:
These functions' docs claim to return the default audio device if 'respeaker' isn't found, but they actually return the second audio device in the list from sounddevices. My fix returns the currently selected audio device for both input and output. These may differ from one another. In my current system, I have a device for macbook microphone and another one for macbook speakers. Same for the airpods, two different sound devices for input and output :)

### DIFF
--- a/src/reachy_mini/media/audio_sounddevice.py
+++ b/src/reachy_mini/media/audio_sounddevice.py
@@ -219,7 +219,7 @@ class SoundDeviceAudio(AudioBase):
         )
         return self._safe_query_device('input')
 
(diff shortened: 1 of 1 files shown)


andimarafioti:
First commit was breaking for the linux test, since linux doesn't require a default audio device. So I now added a little safeguard around that. If there aren't default audio devices, it returns the same second device in the list as before.

FabienDanieau:
Works for me (linux). I've asked our beta testers to also try

PR #334, opened 2025-10-24
Feat collisions mujoco

askuric:
This PR adds the collisions to the mujoco simulation. 

<p>
imageimage
image
</p>

imageimageimage

The collisions are hand made and are composed of multiple convex shapes, capsule for the head and couple of boxes for the body. It helps a lot with making the simulation more realistic, avoiding the unfortunate head inversions. 

This PR has two effects:
1. Makes the simulaiton more realistic, the head and he body can still intersect a bit but its minimal
2. Makes the simulation more stable, constraining the unfeasible configurations where the body and the head collide which correspond relatively well with the singularities that make the head go to its inverse position

The PR makes both placo and symbolic IK more stable and makes it much harder (even though still possible in some very strange situations) to break the simulation. 

When in simulation, you can see the collision objects by pressing on number `3` and hide the visual object by pressing on `2`

**IMPORTANT:** 
- The PR includes the PR #286 
- I've also updated the damping, friction and other motor parameters to make the simulation a bit more realistic (more stiff and damped)
- I've constrained a bit more the solvers in placo to make the simulation more robust.


askuric:
Ok so I was not really happy with the earlier vesion. 
Too many heuristics. 

I've spent some time on this one and remade the collider objects to correspond much more to the real objects and reduced their size as much as reasonably possible. 

Here are the new colliders:
imageimage


imageimage

A short video of it in in action:

https://github.com/user-attachments/assets/6ff4e0c3-9feb-4312-bbb1-b98031517cf7


The execution time of mujoco is a bit longer with this update. I've spent some time on it reducing the size of meshes (fixed htem to 50 vertices) and minimal reasonable number of convex shapes which approximates this highly non-convex body torso (settled for 7). I've used CoACD library for all this (in combination with trimesh to handle stl files). 
So the execution time of mojoco simulation went from 150-200us to 180-250us. 

If the performance is not a bottleneck I've also included a bit more detailed colliders in the `assets/collision/fine` folder. We can go to even more fine ones if necessary. 

If the performance is an issue we can set the this enable mujoco collisions with the `--check-collisions` flag in daemon. But I prefer leaving it on by default for now, because it makes the IK and FK resolution in simulation basically unbreakable.

apirrone:
It seems to work very well ! I couldn't break the robot in simulation

askuric:
Before we merge I'd like to have @RemiFabre's famous unbreakability certificate. :D

RemiFabre:
"makes the IK and FK resolution in simulation basically unbreakable."
image
image

(sorry I had to)

RemiFabre:
So the procedure to break it:
Start from clean state.
X +MAX
Y+MAX
Z-MAX

What seems to happen is that huge forces appear when the head goes down and hits of the side of the body:
image

RemiFabre:
Tested the same procedure on:
feat_placo_collisions
With:
reachy-mini-daemon --sim  --kinematics-engine Placo --check-collision

And it does not break. The forces in play do seem big based on the size of the arrows (they tend to be small when I make small collisions with the body) although I'm not sure as I'm unable to read a numeric value
image
image

RemiFabre:
This version is much harder to "break" than develop, imo it's a good improvement and should be merged

PR #343, opened 2025-10-27
Add pip as a dependency to allow install from uv.

pierre-rouanet:


FabienDanieau:
Works for me! (Linux + uv env)

PR #352, opened 2025-10-30
new behavior for placo and collisions

askuric:
Repaired collision check with placo. 

- Added simpified collisions in URDF 
- Using placo's collision avoidance rather than the collision check from before 

imageimage

imageimage

**TAKEAWAY**
Placo can now plan for the collisions and adapt the motor control to still be able to execute the task, rather than just collide with the robot body and hope the target is reached approximately.

WITHOUT collision avoidance | **WITH** collision avoidance 
--- | ---
Robot `body_yaw` does not move - the head collides<video src="https://github.com/user-attachments/assets/269dc8ff-9311-4996-a12d-12632b61390c.mp4" controls="controls" muted="muted" class="d-block rounded-bottom-2 border-top width-fit" style="max-height:640px; min-height: 200px">|Robot `body_yaw` adapts to allow for low `z` position<video src="https://github.com/user-attachments/assets/9a883303-59b0-4d48-9167-f466d2f0d7f7.mp4" controls="controls" muted="muted" class="d-block rounded-bottom-2 border-top width-fit" style="max-height:640px; min-height: 200px">
The robot's `body_yaw` movement planned without collisions so the head cannot look back when low<video src="https://github.com/user-attachments/assets/02ef25d1-449d-4be6-8951-541c22963254.mp4" controls="controls" muted="muted" class="d-block rounded-bottom-2 border-top width-fit" style="max-height:640px; min-height: 200px">|The robot automatically set the `body_yaw` to be able to look back even if very low<video src="https://github.com/user-attachments/assets/6098f3b0-bbdb-40e1-8690-079112a901e4.mp4" controls="controls" muted="muted" class="d-block rounded-bottom-2 border-top width-fit" style="max-height:640px; min-height: 200px">

### DIFF
--- a/src/reachy_mini/kinematics/placo_kinematics.py
+++ b/src/reachy_mini/kinematics/placo_kinematics.py
@@ -451,9 +451,6 @@ class PlacoKinematics:
             }
         )
 
-        # save the initial configuration
-        q = self.robot.state.q.copy()
-
         done = True
         # do the inital ik with 2 iterations
         for i in range(no_iterations):


RemiFabre:
I was able to reproduce the avoidance mechanism on an IRL Beta Reachy mini, the same type of movements shown in the simulation videos above. This is very cool!

However, even when multiplying x10 the values below, some collisions persisted:

image

This seems to be related to the cases where the collision model doesn't quite fit the physical model. Antun tried more complex shapes but it created instabilities.

This version is a clear improvement still, as in most cases the collision is greatly reduced. It should merged imo, good work!

For future readers that want to activate this, run the daemon as follows:
reachy-mini-daemon --kinematics-engine Placo --check-collision

PR #356, opened 2025-10-30
Allowed for over 7V input voltage.

pierre-rouanet:
This should allow for voltage up to 7.3V. 

I've made the conversion without any real hardware so you should check if it's working @apirrone!


apirrone (approved):
It works !

apirrone:
I merged too fast, I tested and got no voltage errors, but after merging to develop they are back ... Investigating

PR #362, opened 2025-11-04
Fix Windows audio streaming

brainwavecoder9:
### Problem
`stream.write()` blocks indefinitely on Windows with USB audio devices (ReSpeaker), causing silent TTS playback.

### Solution
Replace blocking `.write()` with callback-based streaming using a FIFO queue.

### Details
- `push_audio_sample()` queue chunks instead of blocking write
- `_streaming_callback()` drains FIFO to audio output
- Tunable prebuffer and queue overflow protection 
- Thread lock prevents race on _queued_samples

### DIFF
--- a/src/reachy_mini/media/audio_sounddevice.py
+++ b/src/reachy_mini/media/audio_sounddevice.py
@@ -37,14 +37,19 @@ class SoundDeviceAudio(AudioBase):
         )
 
         # Streaming output state
-        self._streaming_active = False
-        self._chunk_fifo: deque[npt.NDArray[np.float32]] = deque()
-        self._queued_samples: int = 0
-        self._tail: Optional[npt.NDArray[np.float32]] = None
-        self._target_buffer_ms: int = 80
-        self._max_queue_seconds: float = 5.0
-        self._underflows: int = 0
-        self._lock = threading.Lock()
+        self._output_chunk_fifo: deque[npt.NDArray[np.float32]] = deque()
+        self._output_queued_samples: int = 0
+        self._output_tail: Optional[npt.NDArray[np.float32]] = None
+        self._output_max_queue_seconds: float = 5.0
+        self._output_underflows: int = 0
+        self._output_overflows: int = 0
+        self._output_lock = threading.Lock()
+        self._output_max_queue_samples = int(self.get_output_audio_samplerate() * self._output_max_queue_seconds)
+
+    @property
+    def _output_streaming_active(self) -> bool:
+        """Check if output stream is active."""
+        return self._output_stream is not None and self._output_stream.active
 
     def start_recording(self) -> None:
         """Open the audio input stream, using ReSpeaker card if available."""
@@ -119,7 +124,7 @@ class SoundDeviceAudio(AudioBase):
         Note: Channel conversion is handled by MediaManager before this is called.
         Data should already be in the correct format for the output device.
         """
-        if not self._streaming_active or self._output_stream is None:
+        if not self._output_streaming_active:
             self.logger.warning("Output stream not active. Call start_playing() first.")
             return
 
@@ -127,22 +132,19 @@ class SoundDeviceAudio(AudioBase):
         data = np.ascontiguousarray(data, dtype=np.float32)
 
         # Prevent unbounded queue growth
-        with self._lock:
-            max_samples = int(self.SAMPLE_RATE * self._max_queue_seconds)
-            if self._queued_samples + data.shape[0] > max_samples:
-                while self._queued_samples + data.shape[0] > max_samples and self._chunk_fifo:
-                    dropped = self._chunk_fifo.popleft()
-                    self._queued_samples -= dropped.shape[0]
+        with self._output_lock:
+            if self._output_queued_samples + data.shape[0] > self._output_max_queue_samples:
+                while self._output_queued_samples + data.shape[0] > self._output_max_queue_samples and len(self._output_chunk_fifo) > 0:
+                    dropped = self._output_chunk_fifo.popleft()
+                    self._output_queued_samples -= dropped.shape[0]
+                self._output_overflows += 1
                 self.logger.warning(
-                    f"Audio queue overflow ({self._queued_samples} samples), dropped old chunks"
+                    f"Audio queue overflow ({self._output_queued_samples} samples), dropped old chunks"
                 )
 
-            self._chunk_fifo.append(data)
-            self._queued_samples += data.shape[0]
+            self._output_chunk_fifo.append(data)
+            self._output_queued_samples += data.shape[0]
 
-    def _target_buffer_samples(self) -> int:
-        """Prebuffer watermark for smooth playback."""
-        return int(self.SAMPLE_RATE * (self._target_buffer_ms / 1000.0))
 
     def _streaming_callback(
         self,
@@ -156,33 +158,29 @@ class SoundDeviceAudio(AudioBase):
         Note: Data from MediaManager is already formatted to match output channels.
         """
         if status and status.output_underflow:
-            self._underflows += 1
-            if self._underflows % 10 == 1:
-                self.logger.debug(f"Audio underflow count: {self._underflows}")
+            self._output_underflows += 1
+            if self._output_underflows % 10 == 1:
+                self.logger.debug(f"Audio underflow count: {self._output_underflows}")
 
         # Zero the output buffer (silence by default)
         outdata[:] = 0.0
-
-        target = self._target_buffer_samples()
         written = 0
 
         # Drain carryover tail first
-        if self._tail is not None and self._tail.size:
-            take = min(frames, self._tail.shape[0])
-            outdata[:take, :] = self._tail[:take]
+        if self._output_tail is not None and self._output_tail.size:
+            take = min(frames, self._output_tail.shape[0])
+            outdata[:take, :] = self._output_tail[:take]
             written += take
-            self._tail = self._tail[take:] if take < self._tail.shape[0] else None
+            self._output_tail = self._output_tail[take:] if take < self._output_tail.shape[0] else None
 
         # Drain FIFO chunks
         while written < frames:
-            with self._lock:
-                if self._queued_samples < target:
-                    break
-                if not self._chunk_fifo:
+            with self._output_lock:
+                if len(self._output_chunk_fifo) == 0:
                     break
 
-                chunk = self._chunk_fifo.popleft()
-                self._queued_samples -= chunk.shape[0]
+                chunk = self._output_chunk_fifo.popleft()
+                self._output_queued_samples -= chunk.shape[0]
 
             need = frames - written
             if chunk.shape[0] <= need:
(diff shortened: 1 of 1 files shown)


FabienDanieau:
Thank you for the PR! I’ll test it as soon as possible.

Just a quick note: The files in the *tests/* directory are designed to be run using pytest. You can see how this is implemented in our CI [here](https://github.com/pollen-robotics/reachy_mini/blob/c7c2695638de75da8fbda63510c9d8ac393617bb/.github/workflows/pytest.yml#L26). Locally, we adjust the `-m not audio|video|...` flag based on what we want to test.
If you’d like to keep your file in its current form, consider moving it to *examples/debug/*. Eventually, we’ll relocate these files to a more appropriate directory.

brainwavecoder9:
Understood, and thank you for the details. I moved the file as that does sound more appropriate.

This would still need revisited before a pull, but leaving open because of #452.

Happy to cleanup/review any of the changes that you might be interested in, but fyi the other updates are: 
- Fix play_sound() callback closure issue using mutable container
- Add FIFO-based streaming with watermark buffering (to prevent underruns)
- Implement underflow/overflow tracking (diag)
- Use explicit sample rate constant 
- Separate input/output device selection methods

brainwavecoder9:
Have revisited, now ready for pull

brainwavecoder9:
FYI Input normalization (required in all cases) has been separated for ease of review

brainwavecoder9:
Rebased for ease of consideration.

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -92,24 +103,117 @@ def stop_recording(self) -> None:
...
-            )
+        """Push audio samples to the output stream for playback."""
+        if not self._streaming_active or self._output_stream is None:
+            self.logger.warning("Output stream not active. Call start_playing() first.")
+            return
+
+        # Convert to mono if needed (match original behavior)
+        if data.ndim > 1:
+            data = np.mean(data, axis=1)
+        data = np.asarray(data, dtype=np.float32, order="C")
```
CarolinePascal: We slightly changed this part to make it more robust and to remove the `mean` that can be risky with microphones arrays. PR should be merged soon, a small rebase will be required !

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -36,6 +37,16 @@ def __init__(
...
 
+        # Streaming output state
+        self._streaming_active = False
+        self._chunk_fifo: deque[npt.NDArray[np.float32]] = deque()
+        self._queued_samples: int = 0
+        self._tail: Optional[npt.NDArray[np.float32]] = None
+        self._target_buffer_ms: int = 80
+        self._max_queue_seconds: float = 3.0  # Reduced from 5.0 for better responsiveness
+        self._underflows: int = 0
+        self._lock = threading.Lock()
```
CarolinePascal: Could we add an `output` prefix to the variables (e.g. `_output_tail`, `_output_lock`) ? So that we can make a difference between input and output variables.

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -92,24 +103,117 @@ def stop_recording(self) -> None:
...
+        status: sd.CallbackFlags,
+    ) -> None:
+        """Fill audio output buffer from queued chunks."""
+        if status and status.output_underflow:
+            self._underflows += 1
+            if self._underflows % 10 == 1:
+                self.logger.debug(f"Audio underflow count: {self._underflows}")
+
+        out = outdata[:, 0]
+        out[:] = 0.0
```
CarolinePascal: Why don't we work on `outdata` directly here ?

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -92,24 +103,117 @@ def stop_recording(self) -> None:
...
+                        f"Audio queue overflow ({self._queued_samples + dropped_count} samples), "
+                        f"dropped {dropped_count} samples to {self._queued_samples} samples"
+                    )
+
+            self._chunk_fifo.append(data)
+            self._queued_samples += data.shape[0]
+
+    def _target_buffer_samples(self) -> int:
+        """Prebuffer watermark for smooth playback."""
+        return int(self.SAMPLE_RATE * (self._target_buffer_ms / 1000.0))
```
CarolinePascal: `self.SAMPLE_RATE` is now `self.output_sample_rate`, we can also make `_target_buffer_samples` a class attribute/property as the value won't change over time.

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -92,24 +103,117 @@ def stop_recording(self) -> None:
...
         """Open the audio output stream."""
         if self._output_stream is not None:
             self.stop_playing()
+
+        with self._lock:
+            self._streaming_active = True
+            self._chunk_fifo.clear()
+            self._queued_samples = 0
+            self._tail = None
+            self._underflows = 0
```
CarolinePascal: This will be mostly done in `stop_playing` we don't need to do it again here.

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -36,6 +37,16 @@ def __init__(
             ["Reachy Mini Audio", "respeaker"], device_io_type="input"
         )
 
+        # Streaming output state
+        self._streaming_active = False
```
CarolinePascal: This could be defined as a property returning : `self.output_stream is not None and self.output_stream.active`

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -36,6 +37,16 @@ def __init__(
...
+        self._chunk_fifo: deque[npt.NDArray[np.float32]] = deque()
+        self._queued_samples: int = 0
+        self._tail: Optional[npt.NDArray[np.float32]] = None
+        self._target_buffer_ms: int = 80
+        self._max_queue_seconds: float = 3.0  # Reduced from 5.0 for better responsiveness
+        self._underflows: int = 0
```
CarolinePascal: What about a dictionary to keep track of all the input and output underflow/overflow ? That's how we keep track of logs in LeRobot

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -92,24 +103,117 @@ def stop_recording(self) -> None:
...
-        if self._output_stream is not None:
-            if data.ndim > 1:  # convert to mono
-                data = np.mean(data, axis=1)
-            self._output_stream.write(data)
-        else:
-            self.logger.warning(
-                "Output stream is not open. Call start_playing() first."
-            )
+        """Push audio samples to the output stream for playback."""
+        if not self._streaming_active or self._output_stream is None:
```
CarolinePascal: If we define `_streaming_active` as suggested above, we do not need to check for ` self._output_stream is None`!

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -92,24 +103,117 @@ def stop_recording(self) -> None:
...
+            return
+
+        # Convert to mono if needed (match original behavior)
+        if data.ndim > 1:
+            data = np.mean(data, axis=1)
+        data = np.asarray(data, dtype=np.float32, order="C")
+
+        # Prevent unbounded queue growth - only drop if truly over capacity
+        with self._lock:
+            max_samples = int(self.SAMPLE_RATE * self._max_queue_seconds)
```
CarolinePascal: This could be defined as a class attribute/property as well.

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -92,24 +103,117 @@ def stop_recording(self) -> None:
...
+            
+            # Only drop chunks if we're significantly over capacity
+            # This prevents premature dropping that causes the overflow warnings
+            if self._queued_samples > max_samples:
+                dropped_count = 0
+                target_after_drop = int(max_samples * 0.5)  # Drop to 50% capacity
+                
+                while self._queued_samples > target_after_drop and self._chunk_fifo:
```
CarolinePascal: I'm not sure `self._chunk_fifo` will ever be `None`

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -92,24 +103,117 @@ def stop_recording(self) -> None:
...
+
+        # Drain FIFO chunks - minimal lock time
+        while written < frames:
+            chunk = None
+            
+            with self._lock:
+                # Quick check: do we have enough buffered data?
+                if self._queued_samples < target:
+                    break
+                if not self._chunk_fifo:
```
CarolinePascal: I'm not sure `self._chunk_fifo` will ever be None

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -92,24 +103,117 @@ def stop_recording(self) -> None:
...
+            else:
+                self._tail = None
+
+        # Drain FIFO chunks - minimal lock time
+        while written < frames:
+            chunk = None
+            
+            with self._lock:
+                # Quick check: do we have enough buffered data?
+                if self._queued_samples < target:
```
CarolinePascal: I think I would still go on and empty the queue. We can consider adding a fade-out if there are not enough samples to fill the requested `frames` samples.

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -92,24 +103,117 @@ def stop_recording(self) -> None:
...
+                    break
+                if not self._chunk_fifo:
+                    break
+
+                # Pop the chunk while holding the lock
+                chunk = self._chunk_fifo.popleft()
+                self._queued_samples -= chunk.shape[0]
+            
+            # Process chunk outside the lock
+            if chunk is not None:
```
CarolinePascal: Following up on the "we should go on and empty the queue" : if `chunk` is None here, we can fill the remaining `frames - written` samples with `np.linspace(out[written],np.zeros(out.shape[1]),frames - written)`

brainwavecoder9:
Caroline thank you for your comments! It is super helpful and I will work through them.

That said, I have some thoughts about switching up the approach. I will respond privately.

src/reachy_mini/media/audio_sounddevice.py
brainwavecoder9: Saw 474 came through, brought it in

src/reachy_mini/media/audio_sounddevice.py
brainwavecoder9: added clarity for truthiness

src/reachy_mini/media/audio_sounddevice.py
brainwavecoder9: Ah, good catch. I was noticing this but hadn't taken the time to figure out what was causing it.

brainwavecoder9: I removed the stop and I like the fade out. I'll pick up things like abruptness and it kills the magic for me. Added something quick and definitely needs reviewed, but I won't be able to test before you see it.

src/reachy_mini/media/audio_sounddevice.py
brainwavecoder9: no reason

brainwavecoder9:
Got through what I could. Won't have a chance to test before you get at it, but here are some quick notes...

- Looks like we're moving to MediaManager, so I'm trusting its assumptions.
- The fine-grained locking in the callback is probably something you'll want to take a look at. At a minimum there's room for optimization.
- A decision needs to be made for play_sound() and streaming mutual exclusivity. I can imagine various use cases for blocking/non-blocking, and ideally I would like the ability to choose... not sure what you're thinking around this.
- A dictionary does seem like a better approach but I didn't have a chance to get to it. 
- stop_playing() discards any remaining queued audio. Might want to consider a drain before stopping.
- Not sure if we wanted cumulative or per-session details for _output_underflows and _output_overflows, but could see value in having both 

Hopefully this helps takes the edge off. Will circle back when I can.

brainwavecoder9:
FYI not ready for merge but moving out of draft on off chance you don't have perms

brainwavecoder9:
After more than a month of maintaining this, it seems that the team has decided to go in a different direction 3d851da.

Given the overhead of keeping the app dependency in sync, I've decided to switch to a non-Windows platform until #452 is resolved.

@andimarafioti feel free to let me know if you have any related questions.

PR #365, opened 2025-11-04
363 support of uv

FabienDanieau:
Mostly adding the uv.lock so we can use the --frozen in the iso when we set up the env
Could you check that it's working fine with mac?


pyproject.toml
```diff
@@ -51,7 +51,7 @@ examples = ["pynput"]
 mujoco = ["mujoco==3.3.0"]
 nn_kinematics = ["onnxruntime==1.22.1"]
 placo_kinematics = ["placo==0.9.14"]
-gstreamer = ["PyGObject>=3.42.2,<=3.46.0", "gst-signalling==1.1.1"]
+gstreamer = ["PyGObject>=3.42.2,<=3.46.0", "gst-signalling>=1.1.1"]
```
alozowski, line 54: nice! no more pinned `gst-signalling`

FabienDanieau: Yes I planned to release the constraint of websocket in gst-signalling so I'll make a new release soon

alozowski (approved):

PR #371, opened 2025-11-05
bug #370: set gstreamer plugin path in the launcher

FabienDanieau:
To be tested with https://github.com/pollen-robotics/pi-gen/pull/12


FabienDanieau:
tested by @alozowski

PR #390, opened 2025-11-11
Remote control of reachy -> Publish controls for robot, audio and video.

andimarafioti:
Hi 👋 !
This PR adds everything needed to remotely control a Reachy Mini robot from a Hugging Face Space. The goal was simple: open a Space, press a button, and drive a real robot (or its simulation) from the browser, with live video and audio. To make that possible, the daemon needed a full WebSocket-based control and media pipeline, optional audio, and a more modular streaming backend. This PR delivers that.

The daemon now supports an optional WebSocket connection that exposes a JSON command API and streams media back to the Space. When --websocket-uri and --stream-media are enabled, the daemon automatically starts a WebSocket controller for commands, a video streamer, and an audio streamer. Both streaming paths use new async utilities that handle batching, deduplication, reconnect logic, and safe shutdown.

Audio across the stack is now fully optional and redesigned around non-blocking input/output buffers. This keeps real-time audio flowing smoothly whether you run with physical hardware, without audio, or in a mixed browser-to-robot setup.

On the simulation side, new utilities for WebSocket, TCP, and UDP frame transport allow frames from the MuJoCo renderer to be streamed in several ways. A new secondary “studio_close” camera was added for nicer remote-control views, and the offscreen rendering loop was refactored so it can drive multiple cameras and transport layers. The simulation can now publish its video and audio to the same WebSocket pipeline used by the physical robot, which helped prototype the remote-control workflow.

The daemon’s now has more CLI options. There's an audio flag to disable audio, another one to forward websocket URIs, and another one to log everything to a file.

Taken together, this PR transforms the daemon into something that can be controlled entirely from a Space. You can connect from the browser, send commands, and receive realtime video and audio whether working with a real robot or simulation. The system is modular, resilient, and ready for remote demos.

### DIFF
--- a/src/reachy_mini/media/audio_sounddevice.py
+++ b/src/reachy_mini/media/audio_sounddevice.py
@@ -110,8 +110,6 @@ class SoundDeviceAudio(AudioBase):
     def push_audio_sample(self, data: npt.NDArray[np.float32]) -> None:
         """Push audio data to the output device."""
         if self._output_stream is not None:
-            if data.ndim > 1:  # convert to mono
-                data = np.mean(data, axis=1)
             with self._output_lock:
                 self._output_buffer.append(data.copy())
         else:
@@ -127,7 +125,6 @@ class SoundDeviceAudio(AudioBase):
         self._output_stream = sd.OutputStream(
             samplerate=self.get_output_audio_samplerate(),
             device=self._output_device_id,
-            channels=1,
             callback=self._output_callback,
             blocksize=self.frames_per_buffer,
         )
@@ -151,10 +148,11 @@ class SoundDeviceAudio(AudioBase):
                 # Get the first chunk from the buffer
                 chunk = self._output_buffer[0]
                 available = len(chunk)
-                
+                chunk = self.ensure_chunk_shape(chunk, outdata.shape)
+
                 if available >= frames:
                     # We have enough data for this callback
-                    outdata[:, 0] = chunk[:frames]
+                    outdata[:] = chunk[:frames]
                     # Remove the used portion
                     if available > frames:
                         self._output_buffer[0] = chunk[frames:]
@@ -162,12 +160,27 @@ class SoundDeviceAudio(AudioBase):
                         self._output_buffer.pop(0)
                 else:
                     # Not enough data, fill what we can and pad with zeros
-                    outdata[:available, 0] = chunk
-                    outdata[available:, 0] = 0
+                    outdata[:available] = chunk
+                    outdata[available:] = 0
                     self._output_buffer.pop(0)
             else:
                 # No data available, output silence
                 outdata.fill(0)
+
+    def ensure_chunk_shape(self, chunk: npt.NDArray[np.float32], target_shape: tuple[int, ...]) -> npt.NDArray[np.float32]:
+        """Ensure chunk has the shape (frames, num_channels) as required by outdata.
+        
+        - If chunk is 1D, tile to required num_channels.
+        - If chunk is 2D with mismatched channels, use column 0.
+        - If chunk is already correct, return as-is.
+        """
+        num_channels = target_shape[1] if len(target_shape) > 1 else 1
+        if chunk.ndim == 1:
+            return np.tile(chunk[:, None], (1, num_channels))
+        elif chunk.shape[1] != num_channels:
+            # Broadcast first channel only
+            return np.tile(chunk[:, [0]], (1, num_channels))
+        return chunk
 
     def stop_playing(self) -> None:
         """Close the audio output stream."""


RemiFabre:
Pinging @cdussieux as you worked on something similar

src/reachy_mini/daemon/backend/mujoco/video_tcp.py
```diff
@@ -0,0 +1,172 @@
+"""TCP JPEG Frame Streaming.
+
+Simple length-prefixed JPEG over TCP for one client.
+"""
+
+import io
+import socket
+import struct
+from typing import Any, Optional, Tuple
```
andimarafioti: I ended up not using this, we can remove it

src/reachy_mini/daemon/backend/mujoco/video_udp.py
```diff
@@ -51,3 +54,57 @@ def send_frame(self, frame: npt.NDArray[np.uint8]) -> None:
...
+class UDPJPEGFrameReceiver:
+    """A class to receive JPEG frames over UDP."""
+
+    def __init__(self, listen_ip: str = "127.0.0.1", listen_port: int = 5005, timeout: float = 1.0) -> None:
+        """Initialize the UDPJPEGFrameReceiver.
+
+        Args:
+            listen_ip (str): The IP address to listen on.
+            listen_port (int): The port to listen on.
+            timeout (float): The timeout in seconds.
```
andimarafioti: I'm also not using this, we can also remove it

pierre-rouanet:
Hi again @andimarafioti,

I think it would be really great to indeed integrate audio/video streaming using ws or webrtc directly within the daemon (cc @FabienDanieau to sync effort)
And for the control part, I've pointed the existing ws route. Let me know if they do the job for you or if there is something missing/not working!

Also, can you provide a space example to better understand how they interacting?

andimarafioti:
Where did you pointed to the existing ws route? I didn't see that

pierre-rouanet:
Just a few line above, copy pasting here:

> Hi @andimarafioti,
> What I was mentioning during our meeting before. You can use directly this route to send command to the daemon via websocket: https://github.com/pollen-robotics/reachy_mini/blob/develop/src/reachy_mini/daemon/app/routers/move.py#L243
You also have the other way (getting state from the robot) using https://github.com/pollen-robotics/reachy_mini/blob/develop/src/reachy_mini/daemon/app/routers/state.py#L108

andimarafioti:
Ah great, I checked those and we can't really use them. There, you are connecting to the daemon from the outside. If you're on the same machine, this is easy, but if not, you need to expose the daemon to the network. In the same wifi this might still be doable, but as soon as you're outside the wifi you need to go into the router and open ports and make your home connection less safe.
My approach is the opposite, I have a server that the robot connects to. The server can run on a hugging face space (ports already exposed), or locally (no need to expose ports if you're in the same computer/network)

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -95,7 +98,7 @@ def __init__(
         elif self.kinematics_engine == "AnalyticalKinematics":
             from reachy_mini.kinematics import AnalyticalKinematics
 
-            self.head_kinematics = AnalyticalKinematics()
+            self.head_kinematics = AnalyticalKinematics(automatic_body_yaw=False)
```
andimarafioti: I'm getting issues moving the robot without this. I'm not sure if it's my code or an error introduced in develop

src/reachy_mini/daemon/backend/mujoco/video_udp.py
andimarafioti: But at the same time, since we have a video_udp file, it feels weird to have a sender and no receiver

andimarafioti:
You can try out the connection with this space:
https://huggingface.co/spaces/HuggingFaceM4/reachy_mini_remote_control
I'm running locally:
`mjpython -m reachy_mini.daemon.app.main --log-file daemon.log --websocket-uri wss://HuggingFaceM4-reachy-mini-remote-control.hf.space --stream-media`

src/reachy_mini/media/media_manager.py
```diff
@@ -34,6 +34,7 @@ def __init__(
         backend: MediaBackend = MediaBackend.DEFAULT,
         log_level: str = "INFO",
         use_sim: bool = False,
+        use_audio: bool = True,
```
FabienDanieau: Forwarding the use_audio here may not be necessary. The point of the match was to avoid sub cases like DEFAULT with no audio which is NO_MEDIA actually. 
I'd add the backends DEFAULT_AUDIO_ONLY (rename of DEFAULT_NO_VIDEO), DEFAULT_VIDEO_ONLY, and remove. You can do the same with gstreamer if you want to.
Then you would select the backend you want when MediaManager is created

src/reachy_mini/daemon/backend/abstract.py
FabienDanieau: Let's revert this and isolate this issue in another PR

src/reachy_mini/daemon/backend/mujoco/video_udp.py
FabienDanieau: The receiver is actually opencv directly is you look at the camera_opencv.py.
We can remove this code if not explicitly used

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -96,25 +99,63 @@ def push_audio_sample(self, data: npt.NDArray[np.float32]) -> None:
...
+                    else:
+                        self._output_buffer.pop(0)
+                else:
+                    # Not enough data, fill what we can and pad with zeros
+                    outdata[:available, 0] = chunk
+                    outdata[available:, 0] = 0
+                    self._output_buffer.pop(0)
+            else:
+                # No data available, output silence
+                outdata.fill(0)
```
FabienDanieau, line 168: The autoclean feature has been removed. Can you still open the deamon (with audio) and then use the client to play sound? I had a problem that the "wake up" sound was not releasing the device.

FabienDanieau (changes requested):
Very useful use case :)
Just few minor comments to address!

src/reachy_mini/media/media_manager.py
andimarafioti: Got it, missed that, thank you!

src/reachy_mini/media/audio_sounddevice.py
andimarafioti: I tested this PR with the conversation app several times (last one just now), and the audio worked perfectly every time. The way the logic changed here, the "wake up" sound and the following sounds being played are using the same output stream (unless the user explicitly closes and opens a new one), so the device never needs to be released. I think the issue you had is that you were trying to open multiple streams on one device.

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -39,32 +43,33 @@ def start_recording(self) -> None:
...
         frames: int,
         time: int,
         status: sd.CallbackFlags,
     ) -> None:
+        # TODO: Handle OOM for never cleaning the input buffer
         if status:
             self.logger.warning(f"SoundDevice status: {status}")
 
-        self._buffer.append(indata[:, :MAX_INPUT_CHANNELS]) # Sounddevice callbacks always use 2D arrays. The slicing handles the reshaping and copying of the data.
+        self._input_buffer.append(indata[:, :MAX_INPUT_CHANNELS].copy()) # Sounddevice callbacks always use 2D arrays. The slicing handles the reshaping and copying of the data.
```
CarolinePascal, line 66: Good catch with the `copy()` ! I got confused between numpy indexing (that creates a view and not a copy) and advanced indexing (that creates a copy).

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -105,24 +110,65 @@ def stop_recording(self) -> None:
     def push_audio_sample(self, data: npt.NDArray[np.float32]) -> None:
         """Push audio data to the output device."""
         if self._output_stream is not None:
-            self._output_stream.write(np.ascontiguousarray(data))
+            if data.ndim > 1:  # convert to mono
+                data = np.mean(data, axis=1)
```
CarolinePascal: This part is now handled at a higher level in the `MediaManager`. If used correctly, `push_audio_sample` will always get an array with the correct shape as input !

src/reachy_mini/media/audio_sounddevice.py
andimarafioti: :D

src/reachy_mini/media/audio_sounddevice.py
andimarafioti: I tried removing this but it leads to a bug: 
```bash
Exception ignored from cffi callback <function _StreamBase.__init__.<locals>.callback_ptr at 0x113fa42c0>:
Traceback (most recent call last):
  File "/Users/andresmarafioti/Documents/reachy_mini/.venv/lib/python3.12/site-packages/sounddevice.py", line 873, in callback_ptr
    return _wrap_callback(callback, data, frames, time, status)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/andresmarafioti/Documents/reachy_mini/.venv/lib/python3.12/site-packages/sounddevice.py", line 2735, in _wrap_callback
    callback(*args)
  File "/Users/andresmarafioti/Documents/reachy_mini/src/reachy_mini/media/audio_sounddevice.py", line 155, in _output_callback
    outdata[:, 0] = chunk[:frames]
    ~~~~~~~^^^^^^
ValueError: could not broadcast input array from shape (256,2) into shape (256,)
```
You mentioned 'if used correctly', am I not using it correctly ?

nathansegers:
Can we use the daemon without camera's and audio and still connect to the websocket?
I'm trying to get a Dockerfile for the daemon and the Simulator.
I found this one, and running that in a Docker Container, that's almost ready ...
https://huggingface.co/spaces/HuggingFaceM4/reachy_mini_remote_control

andimarafioti:
Hi @nathansegers, yes you can use the daemon with the simulator and connect to the websocket. If you look at that space, you just have to not setup the websockets on the other side and they won't connect/send anything. If you choose to connect the video_feed on simulation mode, it will be a view of the simulation in the daemon

nathansegers:
I might need more documentation to understand exactly what to set up.
I'm still getting issues to get the whole Reachy Mini Daemon into a Container, as it complains about camera's (which I'm don't need right now)

That makes it difficult to know what websocket uri's to connect to, or what the "HuggingFace Space" should contain.
I think it would be best to have the HF Space only contain the code needed for controlling and visualising, but does it need the whole Reachy-Mini SDK codebase, like in the linked Space?

PR #393, opened 2025-11-12
Add volume output on dashboard

alozowski:
This PR addresses issue [#364](https://github.com/pollen-robotics/reachy_mini/issues/364)

### DIFF
--- a/src/reachy_mini/daemon/app/routers/volume.py
+++ b/src/reachy_mini/daemon/app/routers/volume.py
@@ -23,7 +23,6 @@ logger = logging.getLogger(__name__)
 
 # Constants
 AUDIO_COMMAND_TIMEOUT = 2  # Timeout in seconds for audio commands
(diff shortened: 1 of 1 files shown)


src/reachy_mini/daemon/app/routers/volume.py
```diff
@@ -0,0 +1,523 @@
...
+    if system == "Linux":
+        # Try to detect if Respeaker is available
+        try:
+            result = subprocess.run(
+                ["aplay", "-l"],
+                capture_output=True,
+                text=True,
+                timeout=AUDIO_COMMAND_TIMEOUT,
+            )
+            if "respeaker" in result.stdout.lower():
```
FabienDanieau: Could you check for the name "Reachy Mini Audio" ? this is the final name

alozowski:
Can't test on Windows atm, I'm temporarily removing Windows support and will restore it later. Works on macOS though

PR #400, opened 2025-11-14
Add a route to retrieve a curated list of apps for the dashboard.

pierre-rouanet:
This PR provides a way to define a list of apps that will be directly shown in the dashboard/desktop app. This way we can:
- make sure the first app that users will try will work, have been tested, etc.
- change this list whenever we want without having to update the daemon/dashboard/desktop app.

## Technicals details 

- [x] Create a list of "selected apps" for the dashboard here: https://huggingface.co/spaces/pollen-robotics/Reachy-Mini_Best_Spaces/blob/main/generate-dashboard-list.py
- [x] Retrieve and parse this json from the AppManager of the daemon
- [x] The list can be retrieved via /api/apps/list-available/dashboard_selection

### DIFF
--- a/src/reachy_mini/apps/__init__.py
+++ b/src/reachy_mini/apps/__init__.py
@@ -1,20 +1,22 @@
 """Metadata about apps."""
 
-from dataclasses import dataclass, field
+from dataclasses import field
 from enum import Enum
 from typing import Any, Dict
+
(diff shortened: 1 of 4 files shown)


pierre-rouanet:
Close in favor of #401

PR #405, opened 2025-11-14
Add /api/camera/stream endpoint for MJPEG video streaming

tfrere:
- Add camera router with /stream endpoint
- Support both simulation (UDP) and real robot (hardware camera)
- Share single camera instance between all clients
- Optimize with 640x480 resolution and 80% JPEG quality
- Ensure Mujoco rendering loop starts in headless mode for UDP stream

### DIFF
--- a/src/reachy_mini/daemon/app/routers/camera.py
+++ b/src/reachy_mini/daemon/app/routers/camera.py
@@ -1,23 +1,28 @@
 """Camera streaming API routes."""
 
 import asyncio
+from typing import AsyncGenerator
 
 import cv2
 from fastapi import APIRouter, Depends
 from fastapi.responses import StreamingResponse
 
-from ...backend.mujoco.backend import MujocoBackend
-from ...daemon import Daemon
 from reachy_mini.media.camera_constants import CameraResolution
 from reachy_mini.media.camera_opencv import OpenCVCamera
+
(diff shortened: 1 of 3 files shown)


tfrere:
So ? 

I  would also like to expose microphone and speaker datas via web handlers like this one, maybe we should think this as a more global feature ?

@FabienDanieau @pierre-rouanet

src/reachy_mini/daemon/app/routers/camera.py
```diff
@@ -0,0 +1,70 @@
...
+        cam = await _get_shared_camera(is_sim)
+        try:
+            while True:
+                f = cam.read()
+                if f is not None:
+                    resized = np.asarray(cv2.resize(f, (640, 480)), dtype=np.uint8)
+                    _, j = cv2.imencode(".jpg", resized, [cv2.IMWRITE_JPEG_QUALITY, 80])
+                    if _:
+                        yield b"--frame\r\nContent-Type: image/jpeg\r\n\r\n" + j.tobytes() + b"\r\n"
+                await asyncio.sleep(0.04)
```
andimarafioti, line 65: The sleep here make FPS float depending on how much the processing above takes + the sleep. If you want a more fixed FPS you can do a sleep 0.04-processing time.

PR #406, opened 2025-11-14
Use Rerun 0.27 + native URDF loader

oxkitsune:
### What

This PR updates the Rerun example and bumps it to Rerun `0.27.1`. The example now uses Rerun’s native URDF logger, and it fixes a bug where the kinematics model wouldn’t update when running with the MuJoCo backend. 

I also switched `PlacoKinematics::_pose_distance` over to using SciPy to compute the Euler angles to avoid the deprecated `pinocchio` method.

### DIFF
--- a/examples/debug/sound_doa.py
+++ b/examples/debug/sound_doa.py
@@ -1,8 +1,8 @@
 """Reachy Mini sound playback example.
 
-This script demonstrates how to use the microphone array to detect the 
-Direction of Arrival (DoA) of speech. It calculates the position of the 
-sound source relative to the head, transforms it into world coordinates, 
+This script demonstrates how to use the microphone array to detect the
+Direction of Arrival (DoA) of speech. It calculates the position of the
+sound source relative to the head, transforms it into world coordinates,
 and commands the robot to look towards the speaker.
 """
 
--- a/examples/debug/sound_play.py
+++ b/examples/debug/sound_play.py
@@ -32,7 +32,10 @@ def main(backend: str) -> None:
         if samplerate_in != mini.media.get_output_audio_samplerate():
             data = scipy.signal.resample(
                 data,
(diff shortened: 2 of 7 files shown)


pyproject.toml
```diff
@@ -23,22 +23,23 @@ dependencies = [
...
-    "soundfile==0.13.1",    
+    "soundfile==0.13.1",
     "reachy-mini-rust-kinematics>=1.0.1",
     "asgiref",
     "aiohttp",
     "log-throttling==0.0.3",
     "pyusb>=1.2.1",
     "libusb_package>=1.0.26.3",
     "pip>=25",
+    "rerun-sdk==0.27.1",
```
FabienDanieau: Could you leave rerun dependency in the optional dev section? We'd like to keep the core python package as small as possible.

pyproject.toml
```diff
@@ -57,11 +58,7 @@ rerun = [
     # See https://github.com/rerun-io/rerun-loader-python-example-urdf/issues/12
```
FabienDanieau: You may also want to add the dependency `urdf-parser-py`.
This dependency was previously included with rerun-loader-python-example-urdf, which is no longer needed. We commented out rerun-loader-python-example-urdf because it was preventing the PyPI builder from running. This comment should be removed since rerun has a native loader now.

FabienDanieau (changes requested):
Awesome, thank you! This is much simpler. 😊
Could you make these minor adjustments? Once they’re done, I’ll merge the PR.

pyproject.toml
oxkitsune: Yea good shout, I didn't notice the optional section 😄

FabienDanieau:
Just a small ruff formatting and we're good to go!

FabienDanieau:
Sorry :) `src/reachy_mini/kinematics/placo_kinematics.py:6:1: I001 Import block is un-sorted or un-formatted`

Issue #417, opened 2025-11-18
bump gst-signalling version


FabienDanieau:
Fix compatibility issues in sub dependencies (websocket)

PR #418, opened 2025-11-18
enhancement #417: bump gst-signalling version

FabienDanieau:


pierre-rouanet:
Do we need to update uv.lock as well?

PR #432, opened 2025-11-19
fix camera not released when exiting reachy mini context

### LINKED ISSUE #431: Bug: Camera is not released properly when exiting reachy_mini's context

apirrone:


pierre-rouanet (approved):
lgtm

PR #438, opened 2025-11-20
437 audio record from webrtc is to fast

FabienDanieau:
The respeaker outputs a stereo channel even if they are the same. I'll made it more explicit on the gstreamer side. It seems that it was producing a mono output, badly resampled by the sound_record example.

to be tested with `python examples/debug/sound_record.py --backend gstreamer` and the wireless. On the develop branch the play back of the wav is too fast.

We could output a mono but it has to be changed in the default backend as well, and break the API (i.e. conversation app).

### DIFF
--- a/tests/test_audio.py
+++ b/tests/test_audio.py
@@ -23,11 +23,11 @@ def test_play_sound_default_backend() -> None:
 def test_record_audio_and_file_exists() -> None:
     """Test recording audio and check that the file exists and is not empty."""
     media = MediaManager(backend=MediaBackend.DEFAULT_NO_VIDEO)
-    duration = 2  # seconds
+    DURATION = 2  # seconds
     tmpfile = tempfile.NamedTemporaryFile(suffix='.wav', delete=False)
     tmpfile.close()
     media.start_recording()
-    time.sleep(duration)
+    time.sleep(DURATION)
     media.stop_recording()
     audio = media.get_audio_sample()
     samplerate = media.get_input_audio_samplerate()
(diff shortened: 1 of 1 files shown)


tests/test_audio.py
```diff
@@ -70,14 +71,30 @@ def test_play_sound_gstreamer_backend() -> None:
...
-    assert audio is not None    
+
+    while time.time() - t0 < DURATION:
+        sample = media.get_audio_sample()
+
+        if sample is not None:
+            audio_samples.append(sample)
+
     media.stop_recording()
+    samplerate = media.get_input_audio_samplerate()
```
CarolinePascal:
```suggestion
```

Duplicate line

tests/test_audio.py
CarolinePascal:
```suggestion
    DURATION = 2  # seconds
```

CarolinePascal:
```suggestion
    time.sleep(DURATION)
```

tests/test_audio.py
```diff
@@ -70,14 +71,30 @@ def test_play_sound_gstreamer_backend() -> None:
...
+
+    assert len(audio_samples) > 0
+    audio_data = np.concatenate(audio_samples, axis=0)
```
CarolinePascal, line 90: Should we also check the number of channels ?

```
assert audio_data.ndim == 2 and audio_data.shape[1] == 2
```

CarolinePascal:
Tested on the wireless version by @alozowski !

PR #440, opened 2025-11-24
fix(respeaker usb device) : fix case where no ReSpeaker USB device is actually plugged

CarolinePascal:
This PR fixes the `init_respeaker_usb` method to return `None` when no ReSpeaker USB device is actually plugged.


FabienDanieau (approved):
Good catch, thank you!

PR #441, opened 2025-11-24
fixed automatics body yaw

askuric:
This PR fixes the behavior for automatic body yaw flag for placo and analytic kinematics

Behaviors 

`automatic_body_yaw` | `placo` | `AnalyticKinematics`
--- | -- | ---
`True` |  Body yaw automatically adapted if the head yaw is high. Head yaw is almost always equal to the target head yaw. <br> If target head yaw - target body yaw > 65 deg: <br> target body yaw ≠ body yaw <br> target head yaw = head yaw <video src="https://github.com/user-attachments/assets/8226571e-c059-486c-b462-f96f716e00b3"> |  body yaw adapted automatically in order to maintain the relative distance between the body yaw and the head yaw under 65deg. <br> If target head yaw - target body yaw > 65 deg: <br> target body yaw ≠ body yaw <br> target head yaw = head yaw <video src="https://github.com/user-attachments/assets/e398948b-7b04-483f-a7c9-72351c495002">  
`False` |  Body yaw trying to stay at the target body yaw. Head yaw will not always be able to reach the target head yaw.<br> If target head yaw - target body yaw > 65 deg: <br> target body yaw ≈ body yaw <br> target head yaw ≠ head yaw  <video src="https://github.com/user-attachments/assets/f842dcc5-f762-46ca-bd0b-3632f985490c">| body yaw fixed <br> If target head yaw - target body yaw > 65 deg: <br>target body yaw = body yaw <br> target head yaw ≠ head yaw <video src="https://github.com/user-attachments/assets/4f96cd16-39c3-4658-b2c9-929179c87c22">

From this point on the `automatic_body_yaw` will be `True` by default. 
- It was `False` by default before, but the flag was not used
- The default behaviour does not change with this PR (only the default value of the flag)

### DIFF
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -593,7 +593,7 @@ class Backend:
             body_yaw (float): The yaw angle of the body.
 
         """
-        self.head_kinematics.start_body_yaw = body_yaw
+        self.head_kinematics.set_automatic_body_yaw(automatic_body_yaw=body_yaw)
 
     def get_urdf(self) -> str:
         """Get the URDF representation of the robot."""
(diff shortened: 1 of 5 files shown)


apirrone (approved):
Works nicely !

PR #449, opened 2025-11-25
Fix Windows XMOS handling in audio control utils.

### LINKED ISSUE #442: firmware 2.1.3 breaks xvf_host
I was attempting to run `xvf_host REBOOT 1` and got:
`Device (USB)::device_init() -- No device found`

Firmware 2.1.3 sets USB\VID_38FB (Pollen Robotics), but xvf_host expects USB\VID_20B1. Rolling back didn't work, but that's likely because 2.1.2 does not bother to set the default ID.

brainwavecoder9:
Added Windows compatibility for XMOS XVF3800 USB control by handling platform-specific kernel driver operations.

Closes #442

### DIFF
--- a/src/reachy_mini/media/audio_control_utils.py
+++ b/src/reachy_mini/media/audio_control_utils.py
@@ -321,7 +321,7 @@ class ReSpeaker:
 
 def find(vid: int = 0x2886, pid: int = 0x001A) -> ReSpeaker | None:
     """Find and return the ReSpeaker USB device with the given Vendor ID and Product ID."""
-    dev = usb.core.find(idVendor=vid, idProduct=pid)
+    dev = usb.core.find(idVendor=vid, idProduct=pid, backend=get_libusb1_backend())
     if not dev:
         return None
 
@@ -331,22 +331,29 @@ def find(vid: int = 0x2886, pid: int = 0x001A) -> ReSpeaker | None:
 def init_respeaker_usb() -> Optional[ReSpeaker]:
     """Initialize the ReSpeaker USB device. Looks for both new and beta device IDs."""
     try:
+        # Try new firmware first
         dev = usb.core.find(
             idVendor=0x38FB, idProduct=0x1001, backend=get_libusb1_backend()
         )
+        
+        # If not found, try old firmware
         if dev is None:
             dev = usb.core.find(
                 idVendor=0x2886, idProduct=0x001A, backend=get_libusb1_backend()
             )
-            if dev is None:
-                logging.error("No ReSpeaker USB device found !")
-                return None 
-            else:
-                logging.warning("Old firmware detected on ReSpeaker USB device. Please update the firmware!")
+            if dev is not None:
+                logging.warning("Old firmware detected. Please update the firmware!")
+
+        # If still not found, raise error
+        if dev is None:
+            logging.error("No Reachy Mini Audio USB device found!")
+            return None
+        
         return ReSpeaker(dev)
+
     except usb.core.NoBackendError:
         logging.error(
-            "No USB backend was found ! Make sure libusb_package is correctly installed with `pip install libusb_package`."
+            "No USB backend was found! Make sure libusb_package is correctly installed with `pip install libusb_package`."
         )
         return None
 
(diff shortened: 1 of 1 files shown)


brainwavecoder9:
FYI this fixes Windows, Linux TBD.

I previously thought udev rules are still necessary for Linux (like the ones [suggested here](https://discord.com/channels/519098054377340948/1428484261802934312/1442818761668427856) but I'm no longer confident about that. I realized today that the Linux permission issue may have come from having the daemon running while I tried to run the script: `Error executing command REBOOT: [Errno 13] Access denied (insufficient permissions)`

I'm not 100% sure though, since I'm no longer at my Linux system to confirm.

CarolinePascal:
Hi @brainwavecoder9 !

First, I'd like to thank you (again) for hunting down these nasty bugs on Windows 🫶 

It's a very good catch for the missing `backend = ...`, but I don't see why the configuration change is needed. Could you describe the bug you had before adding the fix ? I'll try to find a Linux machine to replicate it !

Best,

Caroline.

brainwavecoder9:
Neither of the other changes are essential, but they add a little robustness. The configuration call ensures the device is ready, and the interface claim attempts exclusive access. 

The original error was `Error executing command VERSION: Operation not supported or unimplemented on this platform` which was caused by the backend issue. Fabien suggested the udev rules not knowing I was on Windows at the time.

The Linux error that I received was `Error executing command REBOOT: [Errno 13] Access denied (insufficient permissions)` **BUT** after getting the same error on Windows (after a successful run) I realized that on the Windows machine that the issue was not permissions but because I had the daemon running... but I no longer had access to the Linux machine to confirm it wasn't something else.

In any case I've added an error message for clarity.

CarolinePascal:
Ok, makes sense ! I could replicate the `Permission denied` on a Linux machine. The provided udev rule didn't work, but this slightly modified one did :
`SUBSYSTEM=="tty", ATTRS{idVendor}=="1a86", ATTRS{idProduct}=="55d3", MODE:="0666"`

Anyway, I'm not confortable with the user having to sudo-set udev rules before using the robot, I prefer your solution. 

What do you think @pierre-rouanet @apirrone @FabienDanieau ?

CarolinePascal:
Side remark : I did not get any permission issue on MacOS or Windows !

brainwavecoder9:
I think this is still a necessary improvement for Windows, but yes removing any need for users to interact with this is actually the ideal solution.

I don't believe I have access to the firmware source, but I think the reason we need to REBOOT is because these modules are failing to initialize on cold boot. I see errors and timeouts related to module not ready or module not responding properly for:

- AEC (Acoustic Echo Cancellation) - Resource ID 33
- AUDIO_MGR (Audio Manager) - Resource ID 35
- PP (Post-Processing) - Resource ID 17
- I2S (Audio Interface) - Resource ID 35

brainwavecoder9:
Caroline's idea of removing the need for users to do this entirely is for sure the right way to go, but FYI that I did confirm on another Linux system that the udev rules are necessary. 

FYI what worked for me: 
```
	sudo nano /etc/udev/rules.d/99-reachy.rules
		# For 2.1.2 & 2.1.3 firmware
		SUBSYSTEM=="usb", ATTR{idVendor}=="38fb", ATTR{idProduct}=="1001", MODE="0666", GROUP="plugdev"
		SUBSYSTEM=="usb", ATTR{idVendor}=="2886", ATTR{idProduct}=="001a", MODE="0666", GROUP="plugdev"

	# refresh
	sudo udevadm control --reload-rules
	sudo udevadm trigger
	
	# add current user to have perms 
	sudo usermod -aG plugdev $USER
```

brainwavecoder9:
Ready for pull:
- Confirmed success on Windows, and that util doesn't mind if daemon is running
- Enabled original intent of firmware check, clarified logic 
- Removed kernel ops in draft
- Cleaned up messaging

FabienDanieau:
Rules are necessary on Linux indeed. We can add them in the firmware folder and a install script. This could be useful for some beta testers. The REBOOT is still necessary although the bug should be very less frequent.
Since this is a Windows oriented issue here we can open another one dedicated to the linux udev rules. As you prefer.

CarolinePascal:
I have a PR ready here for the udev rules : #460 @FabienDanieau ;) Let's move the discussion there !

PR #460, opened 2025-11-28
Adding missing documentation for udev rules in README

CarolinePascal:
This PR adds missing documentation in the README regarding necessary udev rules on Linux systems. Related to #449.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,29 @@ uv run reachy-mini-daemon
 
 The same package provides both the daemon and the Python SDK.
 
+### Linux udev rules setup
+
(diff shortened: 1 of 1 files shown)


CarolinePascal:
We can put the udev rules in a dedicated `99-reachy-mini.rules` file to be copy-pasted if you prefer @FabienDanieau and even put everything into a single bash file to be sudo-executed (though I'm not a huge fan, anything can be hidden in such scripts).

FabienDanieau (approved):
Thanks, that should be enough for Linux users indeed. We'll make a script later if needed

PR #462, opened 2025-11-30
Add raw write method

haixuanTao:
This new WS API routes make it possible to swipe between websocket <=> webserial in conjunction with https://github.com/pollen-robotics/reachy-mini-motor-controller/pull/33 as well as providing lower level control to user who might want to have low level control remotely.


haixuanTao:
Could use a review on this to support #534 . The linter error seems to be unrelated to this PR

haixuanTao:
Gentle ping @apirrone @tfrere , so that we could have some low level control on the api, like torque load, temperature, and so on...

haixuanTao:
Gentle ping @apirrone

Issue #463, opened 2025-11-30
Add audio playback support for emotions in recorded_moves_example


alozowski:
[pollen-robotics/reachy-mini-emotions-library](https://huggingface.co/datasets/pollen-robotics/reachy-mini-emotions-library) dataset includes `.wav` sound files for each emotion, but the `recorded_moves_example.py` script currently only plays the movements without their corresponding audio. This makes the emotions less expressive and engaging.

alozowski:
Closing, I think we already have scattered solutions of this problem (see PR https://github.com/pollen-robotics/reachy_mini/pull/425)

PR #464, opened 2025-11-30
Add sound to emotions example

alozowski:
This PR adds support for playing audio files alongside recorded moves when starting the example script.
Try run `python examples/recorded_moves_example.py -l emotions` to play all emotions with synchronised sound effects.


alozowski:
Fixing https://github.com/pollen-robotics/reachy_mini/issues/463

PR #466, opened 2025-12-01
426 wireless make streaming optional

FabienDanieau:
The deamon has a `--stream `options that enables the video streaming. It is activated by default in the launcher to mimic the current default behavior.

Removing options allows to use ReachyMini client as you would on the Lite version with a direct access to the camera. This doesn't change the way you get frames and audio but there is no unnecessary video encoding.

@alozowski can you confirm that the conversation app is working on device.
@apirrone  can you get the demo you wanted working locally?

We would need a system to remove the option dynamically from the dashboard, or depending on the local app you start. Maybe this could be done in a different issue what do you think?


alozowski:
I confirm that the conversation app is working on device

PR #472, opened 2025-12-01
Normalize input audio in callback to prevent multichannel dilution

### LINKED ISSUE #465: Downstream averaging inadvertently mutes input
Some sound backends (e.g. modern Linux systems running PipeWire) occasionally return shape (N,) or (N, C).

When a backend exposes virtual channels (e.g., 8, 32, 64) but only the first one or two contain real audio, collapsing with mean(axis=1) leads to a mathematical average of near-nothing.

The result is the most silent of silent failures.

brainwavecoder9:
NOTE:
	This method currently fixes input-hardware quirks where requesting 1 channel we may still get more
	Extra channels are virtual, empty, or noise. Averaging all channels causes silence
	Hardware provides stereo (CH0/CH1), but extra junk channels are possible
	Expecting that downstream code should always receive a clean (N,) mono array

Handles mono, stero, and multi-channel input

Fixes #465

### DIFF
--- a/src/reachy_mini/media/audio_sounddevice.py
+++ b/src/reachy_mini/media/audio_sounddevice.py
@@ -61,9 +61,16 @@ class SoundDeviceAudio(AudioBase):
             self.logger.warning(f"SoundDevice status: {status}")
 
         data = indata.copy()
-        
-        if data.ndim > 1:
-            data = data[:, :2].mean(axis=1)
+
+        if data.ndim == 2:
+            channels = data.shape[1]
+            if channels > 4:
+                data = data[:, :2]
+        elif data.ndim == 1:
+            pass 
+        else:
+            self.logger.error(f"Unexpected audio data shape: {data.shape}")
+            return 
 
         self._buffer.append(data)
 


CarolinePascal:
Hi @brainwavecoder9,

There is actually a way to constrain the callback to directly only return the two first channels only, by simply adding `channels=2` in the `InputStream` definition.
I removed this constraint, because I thought it would be nice to allow the users to use a microphone with 1-2-4 channels depending on their needs.

However, if this causes an issue, I'm willing to put it back. Could you quickly explain how to reproduce the bug ? 

Best,

Caroline.

brainwavecoder9:
Hi Caroline,

Thanks for the context on the original design decision, that makes sense! My thinking here is to solve known problems in a way that does not have to be overly concerned with hypothetical scenarios.

The problem with the current code is an _incorrect assumption_ that every channel will always have audio. However, on some modern Linux configurations (particularly Ubuntu 22.10+, Fedora 34+ etc. where Pipewire is the default) virtual channels are presented which do not contain actual audio. 

#### Reproducing the issue:
The issue arises when we ask sounddevice to handle this, because we're stuck with whatever decisions it makes. In some configurations even explicitly setting to channels=1 does not work. This is because in my particular setup (PipeWire on ARM-based Linux) PortAudio does not have a native PipeWire backend, so instead, ALSA exposes a high number of virtual channels regardless of how many are requested. It is not uncommon to have 8-64 channels on modern Linux distros.

Only the first two of those virtual channels contain audio, but the code dutifully provides the "average across all channels" which in my case happens to be 64. As a result, we end up with 2/64 of actual signal... so even a 100% max input is only seen as ~3%, which isn't enough to recognize voice activity, let alone do anything useful with the audio. 

#### Re: Support for alternative configurations
It would be easy to add `if num_channels == 4: data = data.mean(axis=1)` which I could maybe see in cases of a creative re-utilization of the reSpeaker. Adding that condition would solve for 1-2-4, but I'd be hesitant to get too fancy past that since we're getting into the territory of supporting obscure alternative hardware.

Playing devil's advocate for the proposed change: there is a small risk for someone with a very unusual setup where this breaks them, or cases where Reachy is recognized with mics not on channels 0-1. I think these scenarios are highly uncommon, and unlikely enough that they are best addressed when/if they arise.

#### Re: Callback vs. constraint
In addition to fixing the bug, I think defining the policy in callback is the preferred approach. It's less compute overhead, gives us explicit control over channel handling (rather than relying on potentially problematic black-box behavior down the line), and simplifies expectations. I also don't think we lose anything since everything already expects mono once you get past the boundary. 

#### TL;DR:
**My expectation is that the priority is the fastest path to Reachy working as shipped.** Although additional enhancements for alternative configurations probably aren't necessary at this time, this approach makes any future updates easier to manage. 

Let me know if you'd like me to add the 4-channel handling, and I am also happy to continue discussing anything else related to this approach.

brainwavecoder9:
I removed the assumption that _buffer should contain an expected array.

This should reduce the scope of consideration, and limits scope to directly targeting the Linux specific concern. It will only limit to first 2 channels if more than 4 channels are presented.

brainwavecoder9:
#474 looks great

PR #474, opened 2025-12-03
fix(audio channel): making the handling of multiple audio channels more robust with the sounddevice backend.

CarolinePascal:
This PR makes the handling of multiple channels audio inputs/outputs more robust with sounddevice backend : 

- If the input device (microphone) has more than `MAX_INPUT_CHANNELS = 2` channels, all the channels above `MAX_INPUT_CHANNELS` will be cropped out. This fix is inspired by #472 and solves #465. 
That way, `get_audio_sample` will **always** return a 2D numpy array with shape `(samples, channels)` with `channels <= MAX_INPUT_CHANNELS`.

- If the output device (speaker) has more than one output channel, and has been fed with mono-channel data, the data will be replicated over all the channels below `MAX_OUTPUT_CHANNELS = 2` (again because of #465). The same rationale will be applied if the provided data does not have enough channels to fill all the output channels of the device. Inversely, if stereo-channel data is fed to a speaker with only one channel, the second channel will be cropped out to fit the output.
That way `push_audio_sample` will accept **any** 1D/2D numpy array with either one dimension larger than the other (assumption : `channels < samples`).

### DIFF
--- a/src/reachy_mini/media/audio_sounddevice.py
+++ b/src/reachy_mini/media/audio_sounddevice.py
@@ -81,6 +81,19 @@ class SoundDeviceAudio(AudioBase):
             sd.query_devices(self._output_device_id, "output")["default_samplerate"]
         )
 
+    def get_input_channels(self) -> int:
+        """Get the number of input channels of the audio device."""
+        return min(
+            int(sd.query_devices(self._input_device_id, "input")["max_input_channels"]),
+            MAX_INPUT_CHANNELS
+        )
+
+    def get_output_channels(self) -> int:
+        """Get the number of output channels of the audio device."""
+        return int(
+            sd.query_devices(self._output_device_id, "output")["max_output_channels"]
+        )
+
     def stop_recording(self) -> None:
         """Close the audio stream and release resources."""
         if self.stream is not None:
@@ -92,26 +105,6 @@ class SoundDeviceAudio(AudioBase):
     def push_audio_sample(self, data: npt.NDArray[np.float32]) -> None:
         """Push audio data to the output device."""
         if self._output_stream is not None:
-            if data.ndim > 2 or data.ndim == 0:
-                self.logger.warning(f"Audio samples arrays must have at most 2 dimensions and at least 1 dimension, got {data.ndim}")
-                return
-            
-            # Transpose data to match sounddevice channels last convention
-            if data.ndim == 2 and data.shape[1] > data.shape[0]:
-                data = data.T
-
-            # Fit data to match output stream channels
-
-            # Mono input to multiple channels output : duplicate to fit
-            if data.ndim == 1 and self._output_stream.channels > 1:
-                data = np.column_stack((data,) * self._output_stream.channels)
-            # Lower channels input to higher channels output : reduce to mono and duplicate to fit
-            elif data.ndim == 2 and data.shape[1] < self._output_stream.channels:
-                data = np.column_stack((data[:,0],) * self._output_stream.channels)
-            # Higher channels input to lower channels output : crop to fit
-            elif data.ndim == 2 and data.shape[1] > self._output_stream.channels:
-                data = data[:, :self._output_stream.channels]
-
             self._output_stream.write(np.ascontiguousarray(data))
         else:
             self.logger.warning(
--- a/src/reachy_mini/media/audio_base.py
+++ b/src/reachy_mini/media/audio_base.py
@@ -18,6 +18,7 @@ class AudioBase(ABC):
     """Abstract class for opening and managing audio devices."""
 
     SAMPLE_RATE = 16000  # respeaker samplerate
+    CHANNELS = 2  # respeaker channels
 
     def __init__(self, log_level: str = "INFO") -> None:
         """Initialize the audio device."""
@@ -47,6 +48,14 @@ class AudioBase(ABC):
     def get_output_audio_samplerate(self) -> int:
         """Get the outputsamplerate of the audio device."""
         return self.SAMPLE_RATE
+
+    def get_input_channels(self) -> int:
+        """Get the number of input channels of the audio device."""
+        return self.CHANNELS   
+
+    def get_output_channels(self) -> int:
+        """Get the number of output channels of the audio device."""
+        return self.CHANNELS
 
     @abstractmethod
     def stop_recording(self) -> None:
--- a/src/reachy_mini/media/audio_gstreamer.py
+++ b/src/reachy_mini/media/audio_gstreamer.py
@@ -57,7 +57,7 @@ class GStreamerAudio(AudioBase):
     def _init_pipeline_record(self, pipeline: Gst.Pipeline) -> None:
         self._appsink_audio = Gst.ElementFactory.make("appsink")
         caps = Gst.Caps.from_string(
-            f"audio/x-raw,rate={self.SAMPLE_RATE},channels=2,format=F32LE,layout=interleaved"
+            f"audio/x-raw,rate={self.SAMPLE_RATE},channels={self.CHANNELS},format=F32LE,layout=interleaved"
         )
         self._appsink_audio.set_property("caps", caps)
         self._appsink_audio.set_property("drop", True)  # avoid overflow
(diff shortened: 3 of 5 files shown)


src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -109,7 +128,7 @@ def start_playing(self) -> None:
         self._output_stream = sd.OutputStream(
             samplerate=self.get_output_audio_samplerate(),
             device=self._output_device_id,
-            channels=1,
+            dtype="float32",
```
andimarafioti: Why are you changing the dtype here? Seems unrelated to the PR

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -39,10 +39,10 @@ def __init__(
     def start_recording(self) -> None:
         """Open the audio input stream, using ReSpeaker card if available."""
         self.stream = sd.InputStream(
-            blocksize=self.frames_per_buffer,
             device=self._input_device_id,
-            callback=self._callback,
             samplerate=self.get_input_audio_samplerate(),
+            dtype="float32",
```
andimarafioti: Why change the dtype?

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -14,18 +14,18 @@
 
 from .audio_base import AudioBase
 
+MAX_INPUT_CHANNELS = 2
+MAX_OUTPUT_CHANNELS = 2
```
andimarafioti: We do have 4 mics in the robot, right? Eventually we might want to access all of them, will this damage that?
And also, we only have one speaker, right? So why having 2 output channels?

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -59,8 +59,8 @@ def _callback(
     ) -> None:
         if status:
             self.logger.warning(f"SoundDevice status: {status}")
-
-        self._buffer.append(indata.copy())
+            
+        self._buffer.append(indata[:, :MAX_INPUT_CHANNELS]) # Sounddevice callbacks always use 2D arrays. The slicing handles the reshaping and copying of the data.
```
andimarafioti, line 61: Given that we created the streams in this file, this check on every callback that we aren't using more channels than necessary seems excessive. It's better to set the channels when we create the stream.

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -94,8 +94,27 @@ def stop_recording(self) -> None:
...
+
+            # Fit data to match output stream channels
+            output_channels = min(MAX_OUTPUT_CHANNELS, self._output_stream.channels)
+
+            # Mono input to multiple channels output : duplicate to fit
+            if data.ndim == 1 and output_channels > 1:
+                data = np.column_stack((data,) * output_channels)
+            # Lower channels input to higher channels output : reduce to mono and duplicate to fit
+            elif data.ndim == 2 and data.shape[1] < output_channels:
+                data = np.column_stack((data[:,0],) * output_channels)
```
andimarafioti: You're assuming that this is transforming to stereo, it would break if you have a stereo signal and output_channels was 3. I know you set max to 2 output channels, but if that changes, this would break. IMO, it would be better to make it extra explicit and put here `data.ndim == 2 and data.shape[1] == 1 and output_channels == 2`

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -94,8 +94,27 @@ def stop_recording(self) -> None:
...
+            # Higher channels input to lower channels output : crop to fit
+            elif data.ndim == 2 and data.shape[1] > output_channels:
+                data = data[:, :output_channels]
+
```
andimarafioti: if it's stereo to mono, maybe we could take the mean instead of just one channel? There are risks with destructive interferences, but here you have risks with one channel being silent.

src/reachy_mini/media/audio_sounddevice.py
CarolinePascal: True, it slipped out of the next PR I'm cooking !

src/reachy_mini/media/audio_sounddevice.py
CarolinePascal: Same as before.

src/reachy_mini/media/audio_sounddevice.py
CarolinePascal: It's unfortunately not as easy : we may have 4 microphones and one speaker, the sound card actually exposes 2 inputs and 2 outputs.
It means we can only read 2 microphones simultaneously, but also that things can get messy if you only provide mono-output (nasty bug on Windows). 

These values are set based on the sound card we are using, but they may be increased if we want to open more doors to the users (e.g. 4).

src/reachy_mini/media/audio_sounddevice.py
CarolinePascal: This is actually a necessary fix for #465 raised by #472 - In some cases Linux drivers do not go well with PortAudio and come up with 32/64 virtual channels that are not correctly detected by PortAudio. As a consequence, the callback methods will receive all the channels, despite the stream being initialized with only a few channels.

src/reachy_mini/media/audio_sounddevice.py
CarolinePascal: This is actually what motivated the PR initially : if the two channels are actually out of phase, you may end up canceling the signal by averaging them. 
This is a bit of an edge case, where the user removes the on chip processing, but in the nominal case both channels output the same signal, so picking the first one or the average should lead to the same result !

src/reachy_mini/media/audio_sounddevice.py
CarolinePascal: Removed in c80cf6b6bb60a3269c8cb2f11b4003955645dcba

src/reachy_mini/media/audio_sounddevice.py
CarolinePascal: Removed in c80cf6b6bb60a3269c8cb2f11b4003955645dcba

src/reachy_mini/media/audio_sounddevice.py
andimarafioti: Mmm, okay, I see how actually because these mics are so close from each other, the intereference could be destructive. Good call!

src/reachy_mini/media/audio_sounddevice.py
CarolinePascal: I've been thinking about this situation (stereo input and 3 channels output), and thought the most sensible thing to do would be to just duplicate the first channel over the 3 output channels. It does not make sense to send stereo on the two first channels and then zeros (?) on the last one. But I'd be happy to discuss it !

src/reachy_mini/media/audio_sounddevice.py
andimarafioti: Oh wow, that's so weird. But can't you still start the stream with the max_input_channels instead of checking it on every callback?

src/reachy_mini/media/audio_sounddevice.py
andimarafioti: Oki, good call then!

src/reachy_mini/media/audio_sounddevice.py
andimarafioti: Cool, thanks!

src/reachy_mini/media/audio_sounddevice.py
CarolinePascal: You know what, I'll do it just in case.

FabienDanieau (approved):
Works for me thanks!

src/reachy_mini/media/audio_sounddevice.py
brainwavecoder9: Yes you unfortunately can't slice in start_recording() because you don't have access to the actual audio; indata only exists inside the callback. The MAX_INPUT_CHANNELS approach (good call) means the check only has to happen once.

PR #476, opened 2025-12-03
Update read function for numpy compatibility

alozowski:
Removed the deprecated `np.asarray(frame, dtype=np.uint8, copy=False)` call in `camera_opencv.py`'s `read()` function, which is incompatible with numpy 2.x. The function now validates the frame type at runtime and only converts to `uint8` if needed using `astype(np.uint8, copy=False)`, making the code compatible with both numpy 1.x and 2.x.

### DIFF
--- a/src/reachy_mini/media/camera_opencv.py
+++ b/src/reachy_mini/media/camera_opencv.py
@@ -69,7 +69,7 @@ class OpenCVCamera(CameraBase):
             The frame as a uint8 numpy array, or None if no frame could be read.
             
         Raises:
-            RuntimeError: If the camera is not opened or if the frame is not a valid numpy array.
+            RuntimeError: If the camera is not opened.
 
         """
         if self.cap is None:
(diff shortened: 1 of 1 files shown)


alozowski:
I also decided to add explicit Returns and Raises for clarify

src/reachy_mini/media/camera_opencv.py
```diff
@@ -63,14 +63,27 @@ def open(self, udp_camera: Optional[str] = None) -> None:
...
         if self.cap is None:
             raise RuntimeError("Camera is not opened.")
         ret, frame = self.cap.read()
         if not ret:
             return None
-
-        return np.asarray(frame, dtype=np.uint8, copy=False)
+        # Validate frame type
+        if not isinstance(frame, np.ndarray):
+            raise RuntimeError("Expected numpy.ndarray from cv2.VideoCapture.read()")
```
CarolinePascal: `self.cap.read()` will always return a numpy array if `ret` is True !

CarolinePascal:
Only a smol comment and we are good to merge !

PR #479, opened 2025-12-04
Fix body yaw jumping during wake up when starting away from zero position

### LINKED ISSUE #478: Fix body yaw jumping during wake up when starting away from zero position

apirrone:
Disabling automatic body yaw just before wake up, re enabling it afterwards


andimarafioti:
I think the issue might be larger than this, I experience it in other movements and also solved it by deactivating the automatic_body_yaw. I created [an issue](https://github.com/pollen-robotics/reachy_mini/issues/480).

andimarafioti (approved):

Issue #436, opened 2025-11-20
Possible OOM in audio_sounddevice for long sessions


andimarafioti:
When the user passes `start_recording` to an audio device in reachy mini, it will start recording. That is expected, but what happens then is a bit strange:
1- When the user asks for the audio with `get_audio_sample`, reachy_mini returns everything it's recorded so far, and erases it. IMO, there is no reason for the robot to delete the recording at this point
2- If the user doesn't call `get_audio_sample`, reachy will just keep on recording forever. For short sessions on laptops, this isn't an issue, but once we start shipping reachy's with raspberry pi's, they would run out of memory in 1-2 hours.

Proposed solution:
1- We set a maximum size for the buffer. How it is used right now, it feels like more than a couple of minutes of buffer serves no purpose.
2- We write the buffer to disk after x minutes. 

I prefer 1 !

FabienDanieau:
Hey good catch! 
Yes option 1 works for me.

In the [webrtc branch](https://github.com/pollen-robotics/reachy_mini/blob/4a85dc2687e5d33de8c5a4f638692697b6099c71/src/reachy_mini/media/webrtc_client_default.py#L32) I'm just dropping the new samples if the queue is full. This is also what happens in the gstreamer backend.

PR #485, opened 2025-12-04
fix(input buffer): adding sliding input buffer to avoid samples overflow.

CarolinePascal:
This PR solves #436 with a sliding input buffer mechanism. The buffer size has been set to 1 minute, but may be increased if needed.

The PR may be tested with the `test_record_audio_above_max_queue_seconds()` test in `test_audio.py`.


FabienDanieau (approved):
Works well thank you!

PR #486, opened 2025-12-04
Fix gstreamer support (Media + Audio)

OriNachum:
I tested it on Jetson Thor, but I tried to keep it generic.
I had an issue with camera device detection - I added support for original, new and other relevant one.
Also, audio came up 'mousy' - I fixed sample rate in gstreamer.
I also changed from hw: to plughw: for better sample rate support

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "reachy_mini"
-version = "1.2.5rc4"
+version = "1.2.5"
 authors = [{ name = "Pollen Robotics", email = "contact@pollen-robotics.com" }]
 description = ""
 readme = "README.md"


src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -60,37 +60,59 @@ def __init__(self, log_level: str = "INFO") -> None:
 
     def _init_pipeline_record(self, pipeline: Gst.Pipeline) -> None:
         self._appsink_audio = Gst.ElementFactory.make("appsink")
-        caps = Gst.Caps.from_string(
-            f"audio/x-raw,rate={self.SAMPLE_RATE},channels={self.CHANNELS},format=F32LE,layout=interleaved"
+        
+        # 1. Define the target caps (What we WANT: 16k, F32LE)
```
FabienDanieau: Add comments when there is something that we can't understand from the code. No need to specify each step

src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -60,37 +60,59 @@ def __init__(self, log_level: str = "INFO") -> None:
...
+        # 2. Select Source (Hardware vs Auto)
         if self._id_audio_card == -1:
-            audiosrc = Gst.ElementFactory.make("autoaudiosrc")  # use default mic
+            self.logger.info("Using autoaudiosrc (Default System Device)")
+            audiosrc = Gst.ElementFactory.make("autoaudiosrc")
         else:
+            self.logger.info(f"Using alsasrc device plughw:{self._id_audio_card},0")
             audiosrc = Gst.ElementFactory.make("alsasrc")
-            audiosrc.set_property("device", f"hw:{self._id_audio_card},0")
+            audiosrc.set_property("device", f"plughw:{self._id_audio_card},0")
```
FabienDanieau: is the plughw necessary? hw should work on Linux system.

src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -60,37 +60,59 @@ def __init__(self, log_level: str = "INFO") -> None:
...
-        pipeline.add(self._appsink_audio)
-
-        audiosrc.link(queue)
-        queue.link(audioconvert)
-        audioconvert.link(audioresample)
-        audioresample.link(self._appsink_audio)
+        
+        # 4. CRITICAL FIX: Explicit CapsFilter
+        # This forces the audioresample to actually work before data hits the sink
+        capsfilter = Gst.ElementFactory.make("capsfilter")
```
FabienDanieau: There is no need for the capsfilter here since they are already set to the appsink

src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -60,37 +60,59 @@ def __init__(self, log_level: str = "INFO") -> None:
...
+
+        # Add all to pipeline
+        for elem in elements:
+            pipeline.add(elem)
+
+        # 5. Link Them: Source -> Queue -> Convert -> Resample -> CapsFilter -> AppSink
+        if not audiosrc.link(queue):
+            raise RuntimeError("Failed to link audiosrc -> queue")
+        if not queue.link(audioconvert):
+             raise RuntimeError("Failed to link queue -> audioconvert")
```
FabienDanieau: watch out for the indentation!

src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -161,16 +183,18 @@ def _get_sample(self, appsink: GstApp.AppSink) -> Optional[bytes]:
...
+        """Read a sample from the audio card."""
         sample = self._get_sample(self._appsink_audio)
         if sample is None:
             return None
-        return np.frombuffer(sample, dtype=np.float32).reshape(-1, 2)
+            
+        # DYNAMIC FIX: Use self.CHANNELS instead of hardcoded '2'
+        # If channels=1 and you reshape to 2, you get chipmunk audio.
+        try:
+            return np.frombuffer(sample, dtype=np.float32).reshape(-1, self.CHANNELS)
```
FabienDanieau: the audio sample cannot be mono since we specify the appsink to output stereo in the caps

FabienDanieau:
Thank for the PR! I just reviewed the audio part for now. some questions:
> Also, audio came up 'mousy' - I fixed sample rate in gstreamer.

What do you mean? Do you need another sample rate? Since there is a audioresample component this can be done automatically just by setting the caps.
How to you use the output sound?

> I also changed from hw: to plughw: for better sample rate support

Can you explain me the difference between the two? Thanks!

src/reachy_mini/media/audio_gstreamer.py
OriNachum: According to AI Overview (I think it's safe here)

> `hw` directly accesses the audio hardware without changes, requiring apps to match the device's exact sample rate/format, while `plughw` (plug hardware) acts as a smart layer, automatically resampling and reformatting audio on-the-fly to bridge gaps, making it more flexible and user-friendly for apps with different needs. Use `hw` for bit-perfect, lowest-latency output when you control the app's settings, and `plughw` for general use when you need compatibility with varied audio streams. 

It makes me think we may want to support both options - `plughw` as the easy default, and `hw` for those who want tight control.

OriNachum:
> Thank for the PR! I just reviewed the audio part for now. some questions:
> 
> > Also, audio came up 'mousy' - I fixed sample rate in gstreamer.
> 
> What do you mean? Do you need another sample rate? Since there is a audioresample component this can be done automatically just by setting the caps. How to you use the output sound?
> 
> > I also changed from hw: to plughw: for better sample rate support
> 
> Can you explain me the difference between the two? Thanks!

I used media.get_audio_sample here (get_audio_sample method)
https://github.com/OriNachum/autonomous-intelligence/blob/main/reachy-mini/conversation_app/reachy_controller.py

Then process it on listen here:
https://github.com/OriNachum/autonomous-intelligence/blob/main/reachy-mini/conversation_app/gateway_audio.py

I'm relying on Reachy's sample rate, and it worked great on webrtc, but failed on move to gstreamer.
(After fixing device capture, it created the 'chipmunk' voice)

About the gw vs. plughw, I put here as well.
From Google - 
> `hw` directly accesses the audio hardware without changes, requiring apps to match the device's exact sample rate/format, while `plughw` (plug hardware) acts as a smart layer, automatically resampling and reformatting audio on-the-fly to bridge gaps, making it more flexible and user-friendly for apps with different needs. Use `hw` for bit-perfect, lowest-latency output when you control the app's settings, and `plughw` for general use when you need compatibility with varied audio streams. 

I suggest plughw as default for simplicity, and hw as configuration for tighter control.

src/reachy_mini/media/audio_gstreamer.py
OriNachum: Done

src/reachy_mini/media/audio_gstreamer.py
OriNachum: I'll remove and test without it

src/reachy_mini/media/audio_gstreamer.py
OriNachum: Reverted as requested - tested successfully!

src/reachy_mini/media/audio_gstreamer.py
OriNachum: Tested successfully!

src/reachy_mini/media/audio_gstreamer.py
OriNachum: It is overridable, but plughw is default

Generally, how are we with environment variables for defining these?

OriNachum:
@FabienDanieau I did cleanup and followed the feedback/requests - I tested again and it's successful.
Note I added audio stream stability, and took Pi4 into consideration.
Also added windows & mac support (Alsa won't work there - but we may still need testing and adjusting. I can do on another PR both Windows and Mac)

OriNachum:
@FabienDanieau the linter looks at the wrong file (old history)
I changed in ci to `pip install -e .[env]` but it needs to be in `develop` to take effect

FabienDanieau:
Hey can you give the output of these two commands? I'm not quite sure why the camera is not found on the jetson
`gst-device-monitor-1.0 Video/Source`
`v4l2-ctl --list-devices`

FabienDanieau:
> I suggest plughw as default for simplicity, and hw as configuration for tighter control.
In our case, the audiosrc is connected to audioresample and audioconvert so gstreamer will take care of the format conversion. 
Now maybe this solves your issue
```
            audiosrc.set_property("latency-time", self.DEFAULT_LATENCY_US)
            audiosrc.set_property("buffer-time", self.DEFAULT_BUFFER_US)
```
Would you mind trying just this on our develop branch?
You can use `python examples/debug/sound_record.py --backend gstreamer` to check the quality of the recorded audio.

FabienDanieau:
One thing I've noted [here](https://github.com/OriNachum/autonomous-intelligence/blob/c13efe0c6bd9ffbefffd6b9263c2235c14340edc/reachy-mini/conversation_app/gateway_audio.py#L197), is that you need a 16bit audio. I think we could add a function to change the caps of self._appsink_audio, so the `get_audio_sample` will return exactly what you need for your app

OriNachum:
As agreed with Fabio on Discord - I will create 3 new different PRs for this.
I'm keeping it open for reference and handling tonight/tomorrow.

Issue #487, opened 2025-12-04
Streaming options doesn't work


alozowski:
I found out that `reachy-mini-daemon --wireless-version --stream` doesn't work, so I can't connect to the robot from my machine.

Issue #382, opened 2025-11-07
Wireless: change hostname in dashboard


FabienDanieau:
Make it easy to manage multiple reachy mini on the same network.

CarolinePascal:
Cc @cdussieux

Issue #30, opened 2025-06-05
Multiple robots support


pierre-rouanet:
How to find/connect a specific robot if multiple are connected on the same network?

We could have the zenoh daemon storing all available robots (identified by their name) and use this name in the pub/sub routes?

PR #488, opened 2025-12-04
Fix zenoh config

alozowski:
This PR fixes issue https://github.com/pollen-robotics/reachy_mini/issues/487. 
The problem was caused by zenoh using multicast for autodiscovery, which doesn't always work well on Wi-Fi. To fix this, I suggest turning off autoconfiguration and using a direct connection instead.

Additionally, I updated and pinned the zenoh version and regenerated the uv.lock file accordingly.

### DIFF
--- a/src/reachy_mini/io/zenoh_client.py
+++ b/src/reachy_mini/io/zenoh_client.py
@@ -37,6 +37,7 @@ class ZenohClient(AbstractClient):
                 )
             )
         else:
+            # TODO use the name of the robot for multiple robots support 
             c = zenoh.Config.from_json5(
                 json.dumps(
                     {


FabienDanieau:
I think @pierre-rouanet would be a better reviewer than me here

src/reachy_mini/io/zenoh_client.py
```diff
@@ -29,17 +29,24 @@ def __init__(self, localhost_only: bool = True):
...
                 )
             )
         else:
-            c = zenoh.Config()
+            c = zenoh.Config.from_json5(
+                json.dumps(
+                    {
+                        "mode": "client",
+                        "connect": {
+                            "endpoints": ["tcp/reachy-mini.local:7447"]
```
pierre-rouanet, line 46: Regarding #30  and #382, maybe we will also use the "name" of the robot here?

I guess it does not work for the case where you run a daemon with the lite on one computer and tries to run the client on another? But I'm not sure this is a case we want to specifically maintain.

pierre-rouanet:
Hi @alozowski! 

I was interested in zenoh at first for its auto-discovery feature, but to be honest I do not really understand how it works and sometimes I've seen some weird behavior. So I'm fine getting rid of it!

I quickly tested it on my side, but I currently do not have access to a wireless to test it. But if works for you, let's go!

src/reachy_mini/io/zenoh_client.py
alozowski: I was thinking about the same thing, can we extract the robot's name anyhow? If yes, sure, we can try to add it here
UPD: let's leave it as TODO and merge later, the approach to extract a robot's name should be tested

pierre-rouanet: Linking https://github.com/pollen-robotics/reachy_mini/compare/develop...30-multiple-robots-support for reference. 
I haven't really tested it yet so I'm fine with adding a todo for later!

PR #499, opened 2025-12-09
fix: use snapshot_download for HuggingFace Spaces instead of git clone

tfrere:
## 🐛 Problem

When installing apps from HuggingFace Spaces via the Tauri desktop application, users get:

```
git-lfs filter-process: git-lfs: command not found
```

This happens because:
- HuggingFace repos use Git LFS for large files
- `pip install git+https://...` uses `git clone` which requires `git-lfs`
- The Tauri app bundles Python but not `git-lfs`

## ✅ Solution

Use `huggingface_hub.snapshot_download()` instead of `git+<url>` to download HuggingFace Spaces.

### Benefits:
- ✅ **No git-lfs required** - `snapshot_download()` handles LFS files natively
- ✅ **Uses HF cache** - Downloads go to `~/.cache/huggingface/hub/` (shared with other HF tools)
- ✅ **Non-blocking** - Uses `asyncio.to_thread()` for async operation
- ✅ **huggingface-hub already a dependency** - No new dependencies needed

## 📝 Changes

- Modified `src/reachy_mini/apps/sources/local_common_venv.py`
- Extract repo_id from URL (e.g., `https://huggingface.co/spaces/owner/repo` → `owner/repo`)
- Download with `snapshot_download(repo_id, repo_type="space")`
- Then `pip install` from the local downloaded path

## 🧪 Testing

Tested with the Tauri desktop app on macOS - apps now install successfully without git-lfs.


pierre-rouanet (approved):
Nice!

cdussieux:
works fine for me here, updated with the venv system

PR #508, opened 2025-12-10
Objects on the floor in mujoco minimal scene

### LINKED ISSUE #504: Objects on the floor in mujoco minimal scene
Image

apirrone:
collisions are disabled at startup for smoother wake up animations. But they were disabled for everything, so the objects fell to the ground. Found a way to not disable collisions for the objects in the minimal scene


ClaireHzl (approved):
:100:

PR #511, opened 2025-12-10
wireless: allow local use and webrtc

FabienDanieau:
With the --stream option (default behavior), the daemon starts a werbrtc server that takes control of the camera and audio.
To enable the local use of the devices by ReachyMini.media, the camera is redirected to a unix socket, and the alsa is configured to allow multiple client.

To test the PR, you need to manually launch
`src/reachy_mini/daemon/app/services/wireless/generate_asoundrc.sh`
@pierre-rouanet I have added this script to the launcher. It should be automatically called with the daemin once the update is done.

Then you can start the daemon as usual
`GST_DEBUG=3 reachy-mini-daemon --wireless-version --stream`

And try different local examples:
```
GST_DEBUG=3 python examples/debug/sound_record.py --backend gstreamer
GST_DEBUG=3 python examples/debug/sound_play.py --backend gstreamer
GST_DEBUG=3 python examples/debug/take_picture.py --backend gstreamer
```

### DIFF
--- a/src/reachy_mini/daemon/app/main.py
+++ b/src/reachy_mini/daemon/app/main.py
@@ -127,6 +127,7 @@ def create_app(args: Args, health_check_event: asyncio.Event | None = None) -> F
     app.state.args = args
     app.state.daemon = Daemon(
         robot_name=args.robot_name,
+        stream=args.stream,
         wireless_version=args.wireless_version,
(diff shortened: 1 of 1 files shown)


apirrone (approved):
Tested with the conversation demo, @RemiFabre tested too from this branch, all seems to work !! 🚢

PR #522, opened 2025-12-10
Playback sounds associated to moves

FabienDanieau:
Triggering a move should play back the associated sound again.

To test:
- play an emotion from the dashboard (moves are played on the deamon side)
- use the example `python examples/recorded_moves_example.py -l emotions` (moves are played on the client side)

Please check on the lite and the wireless since the media backend are still different.
@RemiFabre please check that the conversation app is not affected.

Expected issue: starting an emotion from the dashboard and playing a sound from the client at the same time. The current default media will fail.

### DIFF
--- a/src/reachy_mini/apps/assistant.py
+++ b/src/reachy_mini/apps/assistant.py
@@ -238,7 +238,7 @@ def check(console: Console, app_path: str) -> None:
         sys.exit(1)
 
     # Extract app name
-    with open(pyproject_file, "rb") as f:
+    with open(pyproject_file, "r", encoding="utf-8") as f:
         pyproject_content = toml.load(f)
         project = pyproject_content.get("project", {})
         app_name = project.get("name", None)
@@ -324,7 +324,7 @@ def check(console: Console, app_path: str) -> None:
     console.print(f"✅ {app_name}/main.py exists.")
 
     # - <app_name>/main.py contains a class named <AppName> that inherits from ReachyMiniApp
-    with open(main_file, "r") as f:  # type: ignore
+    with open(main_file, "r") as f:
         main_content = f.read()
     class_name = "".join(
         word.capitalize() for word in app_name.replace("-", "_").split("_")


RemiFabre:
Nice! I'll test this tomorrow!

RemiFabre:
Thanks for the clear test instructions.

Tested on the Lite version, no sounds played on the first run of the daemon (even though I did pip install after pulling this branch), but it did on the second onward (it has always worked s

https://github.com/user-attachments/assets/5fe2739d-08de-4b99-86cc-1b6de73966c0

ince).

Tested both dashboard and SDK client approaches and both work but there is a significant delay in the sound play when done from the client. I use the emotion "success1" to best test this.

RemiFabre:
The problem seems very simple but with the current implementation there is no elegant fix I can think of. cf comments inline

examples/recorded_moves_example.py
```diff
@@ -16,18 +16,22 @@ def main(dataset_path: str) -> None:
...
-    with ReachyMini(use_sim=False, media_backend="no_media") as reachy:
+    with ReachyMini() as reachy:
         print("Connection successful! Starting dance sequence...\n")
         try:
+            reachy.media.start_playing()
             while True:
                 for move_name in recorded_moves.list_moves():
                     move: RecordedMove = recorded_moves.get(move_name)
                     print(f"Playing move: {move_name}: {move.description}\n")
+                    reachy.media.play_sound(move.sound_path)
```
RemiFabre: The problem is that the emotion doesn't actually start for the duration of the initial interpolation (initial_goto_duration).
setting initial_goto_duration to 0 before playing the sound fixes the problem but is a bad solution imo

RemiFabre:
Ideally the sound play should start after this function has finished:
https://github.com/pollen-robotics/reachy_mini/blob/dfa013ae4ff0ef553166eebecd0591805c3ac8d4/src/reachy_mini/reachy_mini.py#L711


I tried using nitial_goto_duration=0 but it's way too brutal

FabienDanieau:
I see. Is it ok from when you launch the emotions from the dashboard? The sound is played [after the init](https://github.com/pollen-robotics/reachy_mini/blob/c3b670275bcadd6984bd3dc3947f39913aa741d4/src/reachy_mini/daemon/backend/abstract.py#L376). I should do the same in the local play_move actually

FabienDanieau:
The sync should be good for the moves played from the client. If the backend is set to "no_media" they won't be played back. 
Is that still fine with the conversation app? There might be a problem a move is triggered while audio is already playing.

RemiFabre:
> The sync should be good for the moves played from the client. If the backend is set to "no_media" they won't be played back. Is that still fine with the conversation app? There might be a problem a move is triggered while audio is already playing.

Maybe the conversation app should handle 100% of is sound on its side? Feels weird to have a special case just for emotions because we get into the problems you're describing no?

RemiFabre (approved):
Everything works now! In both client and server, with no delays. Also fixed the CI crash, wasn't the fault of this branch (came from develop).

Great improvement, thanks! Merging

Issue #526, opened 2025-12-11
Expression 'ret' failed in 'src/hostapi/alsa/pa_linux_alsa.c', line: 1736


FabienDanieau:
Reported by @RemiFabre 

```
Expression 'ret' failed in 'src/hostapi/alsa/pa_linux_alsa.c', line: 1736
Expression 'AlsaOpen( &alsaApi->baseHostApiRep, params, streamDir, &self->pcm )' failed in 'src/hostapi/alsa/pa_linux_alsa.c', line: 1904
Expression 'PaAlsaStreamComponent_Initialize( &self->playback, alsaApi, outParams, StreamDirection_Out, NULL != callback )' failed in 'src/hostapi/alsa/pa_linux_alsa.c', line: 2175
Expression 'PaAlsaStream_Initialize( stream, alsaHostApi, inputParameters, outputParameters, sampleRate, framesPerBuffer, callback, streamFlags, userData )' failed in 'src/hostapi/alsa/pa_linux_alsa.c', line: 2839
ERROR:reachy_mini.daemon.daemon:Error while waking up Reachy Mini: Error opening OutputStream: Device unavailable [PaErrorCode -9985]
```

PR #527, opened 2025-12-11
bug #526: cleaning up sounddevice after goto sleep

FabienDanieau:
The device was not closed on the daemon side so the audio card was not released properly.


RemiFabre (approved):
Tested on the wireless, works! Thanks

PR #535, opened 2025-12-12
484 install dashboard with uv

alozowski:
This PR adds `uv pip` installation as a default way to install apps from the dashboard to increase installation speed and for better dependency management, but fallback to a standard `pip` installation

### DIFF
--- a/src/reachy_mini/apps/sources/local_common_venv.py
+++ b/src/reachy_mini/apps/sources/local_common_venv.py
@@ -536,7 +536,10 @@ async def uninstall_package(
                     app_name,
                 ]
             
-            return await running_command(uninstall_cmd, logger=logger)
+            ret = await running_command(uninstall_cmd, logger=logger)
+            if ret == 0:
+                logger.info(f"Successfully uninstalled '{app_name}'")
+            return ret
         else:
             # Desktop: remove the entire per-app venv directory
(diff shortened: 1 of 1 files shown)


src/reachy_mini/apps/sources/local_common_venv.py
```diff
@@ -15,6 +15,20 @@
...
+    try:
+        proc = await asyncio.create_subprocess_exec(
+            "uv", "--version",
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+        )
+        await proc.wait()
+        return proc.returncode == 0
+    except FileNotFoundError:
+        return False
```
CarolinePascal:
```
import shutil
if shutil.which("uv"):
    print("uv is installed")
else:
    print("uv is NOT installed")
```

README.md
```diff
@@ -63,6 +63,27 @@ We support and test on Linux and macOS. It's also working on Windows, but it is
...
+> 
+> **Windows:**
+> ```powershell
+> powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"
+> ```
+> 
+> **Or with pipx (alternative method):**
+> ```bash
+> pipx install uv
+> ```
```
CarolinePascal: Not necessary !

PR #542, opened 2025-12-14
fix(recorded-moves): pre-download datasets at startup to prevent timeout

tfrere:
## Problem

When playing emotions/dances for the first time, the robot would freeze or timeout because `snapshot_download()` from HuggingFace Hub was making synchronous network calls. This caused:
- Frontend timeout (10s) in the desktop app
- Robot appearing unresponsive during first emotion playback

### 1. Use local cache during playback
- Added `local_files_only=True` to `snapshot_download()` in `RecordedMoves`
- Ensures instant loading from cache without network latency

### 2. Pre-download datasets at daemon startup (opt-in)
- Added `preload_default_datasets()` function that downloads emotions and dances libraries
- Called during daemon lifespan startup in a background thread
- Non-blocking: daemon starts while datasets download in parallel

### 3. CLI flag for control
- `--preload-datasets`: enable pre-download at startup (default: **disabled**)
- `--no-preload-datasets`: explicitly disable (default behavior)

### 4. Graceful fallback
- If local cache doesn't exist, falls back to network download with a warning
- Ensures the feature always works, even if pre-download was not enabled

## Changes

- `src/reachy_mini/motion/recorded_move.py`:
  - Added `preload_dataset()` and `preload_default_datasets()` functions
  - Added fallback logic in `RecordedMoves.__init__()`
  
- `src/reachy_mini/daemon/app/main.py`:
  - Added `preload_datasets` arg to `Args` dataclass (default: False)
  - Added pre-download call in lifespan context manager
  - Added `--preload-datasets` / `--no-preload-datasets` CLI flags

## Usage

```bash
# Default: no pre-download (existing behavior preserved)
reachy-mini-daemon --sim

# Enable pre-download for faster first emotion playback
reachy-mini-daemon --sim --preload-datasets
```

## Testing

- Tested with emotions and dances - no more freezing or timeout when using --preload-datasets
- Tested fallback when cache is empty - downloads correctly with warning
- Default behavior unchanged (no network access at startup)

### DIFF
--- a/src/reachy_mini/daemon/app/main.py
+++ b/src/reachy_mini/daemon/app/main.py
@@ -487,7 +487,7 @@ def main() -> None:
         "--no-preload-datasets",
         action="store_false",
         dest="preload_datasets",
-        help="Do not pre-download datasets at startup (default: True).",
+        help="Do not pre-download datasets at startup (default: False).",
     )
     # Zenoh server options
     parser.add_argument(
--- a/src/reachy_mini/daemon/app/main.py
+++ b/src/reachy_mini/daemon/app/main.py
@@ -79,6 +79,7 @@ class Args:
     wake_up_on_start: bool = True
     goto_sleep_on_stop: bool = True
     preload_datasets: bool = False
+    dataset_update_interval_hours: float = 24.0  # 0 to disable periodic updates
 
     robot_name: str = "reachy_mini"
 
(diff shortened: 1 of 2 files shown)


src/reachy_mini/daemon/app/main.py
```diff
@@ -99,6 +101,20 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
         """Lifespan context manager for the FastAPI application."""
         args = app.state.args  # type: Args
 
+        # Pre-download recorded move datasets in background to avoid delays on first play
+        # This runs in a thread pool and completes independently (fire and forget)
```
pierre-rouanet: Am I missing the thread pool?

src/reachy_mini/motion/recorded_move.py
```diff
@@ -105,12 +154,33 @@ def evaluate(
...
+    If not cached, falls back to network download (which may cause delays).
+    """
 
     def __init__(self, hf_dataset_name: str):
         """Initialize RecordedMoves."""
         self.hf_dataset_name = hf_dataset_name
-        self.local_path = snapshot_download(self.hf_dataset_name, repo_type="dataset")
+        # Try local cache first (instant, no network)
+        try:
+            self.local_path = snapshot_download(
```
pierre-rouanet, line 169: If we update the dataset. They will not be download again right?
If this is a case, we will maybe need some way to make sure they are still updated at some point I guess.

tfrere:
**Responses to review comments:**

**Thread pool question:** `run_in_executor(None, ...)` uses asyncio's default ThreadPoolExecutor. Clarified the comment in 16c175d.

**Dataset updates:** Added a background task that checks for updates every 24h (configurable via `--dataset-update-interval`). This ensures datasets stay fresh even when the daemon runs 24/7 in wireless mode. See commit aaebeb3.

pierre-rouanet (approved):

PR #544, opened 2025-12-15
fix output callback

andimarafioti:
There was a bug with the output buffer.
If the size of chunks added to the buffer wasn't an integer multiple of the size of the blocksize in the output stream, it would lead to every X sample getting filled with zeros, even when the buffer had extra samples.

### DIFF
--- a/src/reachy_mini/media/audio_sounddevice.py
+++ b/src/reachy_mini/media/audio_sounddevice.py
@@ -195,7 +195,6 @@ class SoundDeviceAudio(AudioBase):
             filled = 0
             while filled < frames and self._output_buffer:
                 chunk = self._output_buffer[0]
-                chunk = self.ensure_chunk_shape(chunk, outdata.shape)
                 
                 needed = frames - filled
                 available = len(chunk)
@@ -263,8 +262,7 @@ class SoundDeviceAudio(AudioBase):
             data = scipy.signal.resample(
                 data, int(len(data) * (samplerate_out / samplerate_in))
             )
-        if data.ndim > 1:  # convert to mono
-            data = np.mean(data, axis=1)
+        data = self.ensure_chunk_shape(data, (-1, self.get_output_channels()))
 
         self.logger.debug(f"Playing sound '{file_path}' at {samplerate_in} Hz")
 


src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -190,30 +190,30 @@ def _output_callback(
...
-
+        
         with self._output_lock:
-            if self._output_buffer:
-                # Get the first chunk from the buffer
+            filled = 0
+            while filled < frames and self._output_buffer:
                 chunk = self._output_buffer[0]
-                available = len(chunk)
                 chunk = self.ensure_chunk_shape(chunk, outdata.shape)
```
CarolinePascal: Slightly off topic : as we discussed last week `ensure_chunk_shape` is actually not needed here, and should be moved to `play_sound` instead !

src/reachy_mini/media/audio_sounddevice.py
```diff
@@ -190,30 +190,30 @@ def _output_callback(
...
-                    outdata[available:] = 0
+                    # Fully consumed this chunk
                     self._output_buffer.pop(0)
-            else:
-                # No data available, output silence
-                outdata.fill(0)
+            
+            # Only pad with zeros if buffer is truly empty
+            if filled < frames:
+                outdata[filled:] = 0
```
CarolinePascal, line 215: It would be better to add a fade-out here to avoid sharp transitions ! Linear interpolation should do : 
```
remaining = frames - filled
fade_len = min(64, remaining)     #64 is arbitrary here
fade_start_val = outdata[filled - 1, :]
fade_curve = np.linspace(1.0, 0.0, fade_len, dtype=np.float32)
outdata[filled:filled + fade_len, :] = fade_start_val * fade_curve[:, np.newaxis] 
```

CarolinePascal: Also this part could be outside the ` with self._output_lock:` to release the lock asap ! `filled=0` will have to be pushed outside the `with` context in that case.

src/reachy_mini/media/audio_sounddevice.py
CarolinePascal: (Feel free to ignore this comment and mark the discussion as resolved)

andimarafioti: It actually breaks without this check :(

src/reachy_mini/media/audio_sounddevice.py
andimarafioti: This is a good idea, but honestly with the gradium implementation we were getting tons of this disruptions, and the annoying part was the silence more than the discontinuity itself. It might be a characteristic of the speaker in the mini that makes it not very disruptive.

CarolinePascal: Keeping this as a future improvement then !

src/reachy_mini/media/audio_sounddevice.py
andimarafioti: Added the change, thank you!

PR #548, opened 2025-12-15
Update README.md

clmnt:
remove the tag line that is not really aligned with what we've been saying


clmnt:
mind approving @RemiFabre @matthieu-lapeyre

PR #552, opened 2025-12-15
fix(apps): prevent app restart race condition and improve cleanup

tfrere:
## Problem

When stopping an app via the desktop app, rapidly restarting it before the previous instance fully terminates causes:
- Force kill (SIGKILL) of the app mid-cleanup
- Robot receiving conflicting commands from residual threads
- Erratic robot movements on subsequent app launches

## Root cause

1. `is_app_running()` returned `False` during `STOPPING` state, allowing premature restarts
2. 5s timeout was too short for apps with cleanup routines (e.g., conversation app resets robot to neutral in ~4s)
3. Force-killed processes could leave orphan child processes

## Changes

- Include `STOPPING` in `is_app_running()` to block restarts during shutdown
- Increase stop timeout from 5s to 10s
- Add `_kill_process_tree()` to recursively terminate child processes on force kill

### DIFF
--- a/src/reachy_mini/apps/manager.py
+++ b/src/reachy_mini/apps/manager.py
@@ -9,6 +9,7 @@ from enum import Enum
 from typing import TYPE_CHECKING, Any, Optional
 
 import numpy as np
+import psutil
 from pydantic import BaseModel
 
 from . import AppInfo, SourceKind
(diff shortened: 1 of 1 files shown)


apirrone (approved):
Tested on @tfrere's mac, and on my linux (doesn't break anything)

PR #562, opened 2025-12-16
Adding entrypoint script to easily reflash the motors

apirrone:
just run `reachy-mini-reflash-motors`

### DIFF
--- a/src/reachy_mini/tools/reflash_motors.py
+++ b/src/reachy_mini/tools/reflash_motors.py
@@ -36,13 +36,13 @@ def main() -> None:
     motors = list(config.motors.keys())
     if args.serialport is None:
         console.print(
-            "Are you running this script on the wireless version of Reachy Mini?"
+            "Which version of Reachy Mini are you using?",
         )
         wireless_choice = questionary.select(
             ">",
             [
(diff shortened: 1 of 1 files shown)


src/reachy_mini/tools/reflash_motors.py
```diff
@@ -0,0 +1,95 @@
...
+    console = Console()
+
+    config_file_path = str(files(reachy_mini).joinpath("assets/config/hardware_config.yaml"))
+    config = parse_yaml_config(config_file_path)
+    motors = list(config.motors.keys())
+    if args.serialport is None:
+        console.print(
+            "Are you running this script on the wireless version of Reachy Mini?"
+        )
+        wireless_choice = questionary.select(
```
pierre-rouanet, line 41: Maybe we should make it more explicit to avoid saying yes without really knowing.

Which versions are you running on:
> lite
> wireless

What do you think @apirrone?

apirrone: yes

pierre-rouanet:
Apart from the comment above, it works on my side (tested on a lite)!

PR #568, opened 2025-12-16
Claude/fix hume prosody audio d ej at

MrPhussy:


FabienDanieau:
Thanks @MrPhussy, and sorry for the long silence on this one.

Two separate things here:

- **The Hume example**: we'd rather not host third-party integrations in `examples/` — we want to keep the core repo free of external services and API keys. A Hume app is a great fit as its own repo or a Hugging Face Space, and we'd be happy to link it from the docs.
- **The audio fix**: we're definitely open to a PR that unblocks your app on our side. But `src/reachy_mini/media/audio_sounddevice.py` no longer exists — media now goes through GStreamer (`media/audio_gstreamer.py`). If you still hit the channel/shape issue, could you open a small PR against current `main` targeting the GStreamer path?

Two notes for that one: `self.ensure_chunk_shape()` isn't defined anywhere in this branch, and the description is empty. Please follow our PR template (what's the problem, how to reproduce, what's the fix) and keep it to one logical change — the `uv.lock` regeneration is ~99% of this diff.

Closing this for now, but happy to review the follow-up.

PR #571, opened 2025-12-17
Tentative fix

RemiFabre:
This patch is a partial fix to the problem.

This patch only guards moves that the daemon itself runs. Anything that calls backend.play_move() (including /move/play/..., /move/goto, /move/wake_up) now acquires the RLock, so:
- A daemon play_move or goto_target cannot overlap with another daemon play_move. New requests are
    rejected with a warning.
- While that daemon-managed move is running, direct set_target traffic from HTTP or the Zenoh server is ignored => so a client spamming set_target() cannot fight a daemon playback. (maybe the warning spams too much?)

What this fix does not cover is any situation where multiple SDK clients drive motions purely through the SDK (e.g. each calling mini.play_move() or spamming set_target).

A fix for all cases is a bit more complicated. One way of doing it is for clients to acquire a lock from the daemon (given only of the daemon is idle). But this means identifying the client.

### DIFF
--- a/src/reachy_mini/daemon/app/routers/move.py
+++ b/src/reachy_mini/daemon/app/routers/move.py
@@ -230,6 +230,10 @@ async def set_target(
     backend: Backend = Depends(get_backend),
 ) -> dict[str, str]:
     """POST route to set a single FullBodyTarget."""
+    if backend.is_move_running:
+        # Avoid fighting with the daemon while a trajectory is running
+        backend.logger.warning("Ignoring set_target request: move already running.")
+        return {"status": "ignored", "reason": "move_running"}
     backend.set_target(
         head=target.target_head_pose.to_pose_array()
         if target.target_head_pose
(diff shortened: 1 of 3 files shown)


RemiFabre:
@CarolinePascal what do you think of this? Could you please test an confirm this works? This should at least cover the (common) case of people clicking on a dashboard emotion while something else is running.

@pierre-rouanet I know you've thought about this issue before, maybe the fix is clear for you?

RemiFabre:
Note: the diff on github is misleading, the changes are relatively minimal, but need to be super well tested before merging. (I use kdiff3 for this type of diffs)

CarolinePascal:
@RemiFabre - Sorry for the delay, it works perfectly ! Waiting for @pierre-rouanet opinion to merge ;)

pierre-rouanet:
If it solves an urgent issue for the dashboard I'm fine with it. Not sure about the general problem, but it will probably require deeper changes anyways.

Sorry I don't really have time to test it now. If it's fine for you @CarolinePascal then go!

PR #573, opened 2025-12-17
fix(stream samplerate): fix sample rate issue when using streaming on the wireless

CarolinePascal:
The PR fixes pitch issues when using streaming by adding proper recovery of the channels number in webrtc backend.


FabienDanieau (approved):
Great!

PR #574, opened 2025-12-17
add a method to clear the output buffer with audio sound device. Help…

andimarafioti:
Add a method to clear the output buffer with audio sound device. Helpful to make interruptions work in other apps like the conversation app. 
The context here is that the audio sounddevice has this output queue which might be populated, and to really make the robot stop speaking we need to empty it. Without this, the conversation app feels less responsive. Plus, the code was there already for gstreamer, so this should be rather straight forward to merge.

### DIFF
--- a/src/reachy_mini/media/audio_sounddevice.py
+++ b/src/reachy_mini/media/audio_sounddevice.py
@@ -171,7 +171,7 @@ class SoundDeviceAudio(AudioBase):
 
     def start_playing(self) -> None:
         """Open the audio output stream."""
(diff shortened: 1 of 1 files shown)


CarolinePascal:
All good, but a small request : could you replace the (very unsafe) `self._output_buffer.clear()` at the beginning of `def start_playing(self):` with your new method ? 

Ultimate nitpicking : I would also clear the buffer after checking whether we should call `stop_playing` !

andimarafioti:
good comments! i made the first replacement you suggested, and added a clear buffer to stop playing 💡

PR #594, opened 2025-12-20
Updating doc with missed deps required by a brand new MBP

aguitto:
On a brand new mbp it is required to install some missed deps in order to let it run. 
Added the bash command in the doc to install those deps (cmake, pkg-config, cairo).


jyvet:
I was about to create a similar PR for Linux. With Debian, I had to run:

`sudo apt install cmake libcairo2-dev pkg-config python3-dev`

Cheers,

aguitto:
Would you like I add it here so we can group the changes togheter ?

Issue #596, opened 2025-12-21
Issues are poorly formatted and lack debug info


RemiFabre:
=> Provide a template for GitHub issues

PR #597, opened 2025-12-21
Provides an issue template

RemiFabre:
We're starting to have issues that are poorly formatted and lack critical info to be able to debug (like robot's version, OS, etc).

I created an issue template heavily inspired from LeRobot's.


RemiFabre:
Closes https://github.com/pollen-robotics/reachy_mini/issues/596

PR #608, opened 2025-12-22
reflash motor on start

apirrone:
Temporary : Always run reflash motors when starting the daemon. This adds ~5s to the startup sequence on a wireless


apirrone:
Closed by #609

PR #611, opened 2025-12-22
Support recorded move datasets with optional data/ folder and updated…

RemiFabre:
… example script

Minimal changes to read move datasets with or without the extra "data" folder.

You can test both work like this:
python recorded_moves_with_url.py -l emotions
python recorded_moves_with_url.py --dataset RemiFabre/dances

These changes makes the SDK able to read datasets created with the app Marionette.

### DIFF
--- a/docs/SDK/gstreamer-installation.md
+++ b/docs/SDK/gstreamer-installation.md
@@ -1,0 +1,181 @@
+# 📡 GStreamer Installation for Wireless Reachy Mini
+
+> This guide will help you install [GStreamer](https://gstreamer.freedesktop.org) to receive video and audio streams from your wireless Reachy Mini.
+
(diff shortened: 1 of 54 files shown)


examples/recorded_moves_example.py
```diff
@@ -2,14 +2,20 @@
 
 Run :
 
-python3 recorded_moves_example.py -l [dance, emotions]
+python3 recorded_moves_with_url.py -l [dance, emotions]
```
apirrone: This should ne re renamed to `recorded_moves_examples.py`, `recorded_moves_with_url.py` doesn't exist

apirrone (changes requested):
Apart from the very minor rename request, it works !

PR #612, opened 2025-12-22
Added flag to let an app start when robot startup

aguitto:
image

### 1. Startup Configuration Module (src/reachy_mini/apps/startup_config.py)
Created a utility module to manage app startup preferences
Preferences are stored in a JSON file (app_startup_config.json) in the venv parent directory
Functions to load, save, get, and set startup preferences for apps

### 2. API Endpoints (updated src/reachy_mini/daemon/app/routers/apps.py)
GET /api/apps/startup-preference/{app_name} - Get the startup preference for an app
POST /api/apps/startup-preference/{app_name} - Set the startup preference for an app

### 3. UI Updates (updated src/reachy_mini/daemon/app/dashboard/static/js/apps.js)
Added a checkbox next to each app's enable/disable toggle
The checkbox is labeled "Start at startup"
Checkbox state is loaded from the backend when the app list is displayed
Changes are saved to the backend when the checkbox is toggled
Updated the grid layout to accommodate the new checkbox column

### 4. Auto-start Logic (updated src/reachy_mini/daemon/app/main.py)
Added logic in the daemon startup lifecycle to automatically start apps that are marked to start at startup
Only the first app in the list is started (since only one app can run at a time)
Includes error handling if the auto-start fails
The checkbox appears next to each app's toggle in the dashboard. When checked, that app will automatically start when the daemon starts up. 

The preference is persisted in a JSON configuration file.

### DIFF
--- a/src/reachy_mini/apps/startup_config.py
+++ b/src/reachy_mini/apps/startup_config.py
@@ -1,0 +1,141 @@
+"""Utility functions for managing app startup preferences."""
+
+import json
+import logging
+import sys
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+def _get_venv_parent_dir() -> Path:
+    """Get the parent directory of the current venv (OS-agnostic)."""
+    executable = Path(sys.executable)
+    
+    # Determine expected subdirectory based on platform
+    import platform as platform_module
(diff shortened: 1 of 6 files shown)


FabienDanieau:
we're finally working on this feature here #1238 . Let's discuss if you're still motivated by this! We can still use the conf file feature.

PR #622, opened 2025-12-24
examples: Add a simple example on how to play emotions

cverna:
Emotions are a really fun part of using Reachy, this is a simple example for anyone wanted to start playing with emotions in their own application.


apirrone:
Thanks ! But we already have this example script that pretty much does the same thing ^^ https://github.com/pollen-robotics/reachy_mini/blob/develop/examples/recorded_moves_example.py

cverna:
Oh great I missed it, I guess because of the name of the file, it was not obvious that emotions or dances are recorded moves.
But naming is hard 👍

PR #627, opened 2025-12-26
fix(windows): resolve pwd import error and Unicode console encoding issues

CharlesCNorton:
## Summary
- Fix `pwd` module import error on Windows by making it conditional
- Replace Unicode emoji with ASCII alternatives for Windows console compatibility

## Problem
Windows users encounter two errors when using the SDK:

1. `ModuleNotFoundError: No module named 'pwd'` when importing the daemon
2. `UnicodeEncodeError: 'charmap' codec can't encode character '\u2705'` when the motor setup runs

## Solution
1. `startup_check.py`: Conditionally import `pwd` only on non-Windows systems (same pattern as PR #299 for `fcntl`)
2. `setup_motor.py`: Detect console encoding at startup; use ✅/❌ on UTF-8 consoles, fall back to `[OK]`/`[FAIL]` on Windows

## Test plan
- [x] Verified daemon imports successfully on Windows
- [x] Verified motor communication works on Windows (COM5)
- [x] Robot responds correctly (wake up, head movement)

### DIFF
--- a/src/reachy_mini/tools/setup_motor.py
+++ b/src/reachy_mini/tools/setup_motor.py
@@ -10,12 +10,26 @@ Please note that all values given in the configuration file are in the motor's r
 """
 
 import argparse
+import sys
 import time
 from pathlib import Path
 
 from rustypot import Xl330PyController
 
 from reachy_mini.utils.hardware_config.parser import MotorConfig, parse_yaml_config
+
+
+def _supports_unicode() -> bool:
+    """Check if the console supports Unicode output."""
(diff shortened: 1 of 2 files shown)


CarolinePascal:
Hi @CharlesCNorton,

Thank you for your contribution, though I think we spotted it a bit too late - The solved issues have been tackled in our 1.2.6 release.

I'm closing this PR, but feel free to open a new one if you spot another issue !

Best,

Caroline.

PR #621, opened 2025-12-24
fix(webrtc): Use H264 Level 3.1 for Safari/WebKit compatibility

tfrere:
## Problem

Safari and WebKit-based browsers (including Tauri on macOS) only support H264 Level 3.1 for WebRTC. The current configuration uses Level 4.0, which is rejected by these browsers during SDP negotiation, causing the video stream to fail.

## Root Cause

- **Current**: `level=(string)4` (Level 4.0) - supports 1080p@30fps but incompatible with Safari
- **Safari supports**: `profile-level-id=42e01f` (Constrained Baseline Level 3.1) - max 720p

### H264 Level Specifications

| Level | Max Macroblocks/s | Max Resolution @ 30fps |
|-------|-------------------|------------------------|
| 3.1   | 108,000           | 1280×720 (720p)        |
| 4.0   | 245,760           | 1920×1080 (1080p)      |

## Solution

1. Change default wireless resolution from 1080p@30fps to 720p@30fps (within Level 3.1 limits)
2. Set H264 level to 3.1 with constrained-baseline profile

## Changes

- `camera_constants.py`: Change `default_resolution` to `R1280x720at30fps`
- `webrtc_daemon.py`: Set `level=(string)3.1,profile=(string)constrained-baseline`

## Impact

- ✅ WebRTC stream now works on Safari, Tauri/macOS, and all other browsers
- ⚠️ Resolution reduced from 1080p to 720p (acceptable for camera preview)

## Testing

- [ ] Safari macOS
- [ ] Tauri app (WebKit)
- [ ] Chrome
- [ ] Firefox

PR #641, opened 2025-12-27
feat(webrtc): Add configurable H264 profile for cross-platform compatibility

tfrere:
## Summary

Add CLI argument `--h264-profile` to configure H264 encoding profile for WebRTC streaming, allowing users to choose between compatibility and quality.

## Motivation

Safari/WebKit only supports H264 Level 3.1 with Constrained Baseline profile, which limits resolution to 720p. However, users on Chrome/Firefox/Edge don't need this limitation and can benefit from 1080p streaming.

This PR adds a configurable option instead of hardcoding one profile.

### New CLI argument

```bash
# Default: Safari compatible (720p)
reachy-mini-daemon --wireless-version

# High quality mode (1080p, no Safari support)
reachy-mini-daemon --wireless-version --h264-profile quality
```

### Available profiles

| Profile | H264 Level | H264 Profile | Resolution | Safari | Chrome/Firefox |
|---------|------------|--------------|------------|--------|----------------|
| `compatible` (default) | 3.1 | constrained-baseline | 720p@30fps | ✅ | ✅ |
| `quality` | 4.0 | main | 1080p@30fps | ❌ | ✅ |

### Files changed

- `media/webrtc_daemon.py` - Add `H264_PROFILES` config and `h264_profile` parameter
- `daemon/daemon.py` - Pass `h264_profile` to WebRTC constructor
- `daemon/app/main.py` - Add `--h264-profile` CLI argument
- `media/camera_constants.py` - Keep 1080p default (WebRTC overrides based on profile)

## Testing

- [ ] Test with `--h264-profile compatible` on Safari/macOS
- [ ] Test with `--h264-profile quality` on Chrome
- [ ] Verify default behavior (compatible) works on all platforms

## Related

Supersedes #621 - Instead of hardcoding Level 3.1, this makes it configurable.

### DIFF
--- a/src/reachy_mini/media/webrtc_client_gstreamer.py
+++ b/src/reachy_mini/media/webrtc_client_gstreamer.py
@@ -56,7 +56,7 @@ class GstWebRTCClient(CameraBase, AudioBase):
         self._pipeline_record.add(self._appsink_audio)
 
         self.camera_specs = cast(CameraSpecs, ReachyMiniWirelessCamSpecs)
-        self.set_resolution(CameraResolution.R1920x1080at30fps)
+        self.set_resolution(self.camera_specs.default_resolution)
 
         self._appsink_video = Gst.ElementFactory.make("appsink")
         caps_video = Gst.Caps.from_string("video/x-raw,format=BGR")


FabienDanieau:
For now the resolution of the client is hardcoded [see here](https://github.com/pollen-robotics/reachy_mini/blob/9c3839637dbdc0031c1a51727468a7c3da6a79c3/src/reachy_mini/media/webrtc_client_gstreamer.py#L59). So the look_at example won't work with the `compatible` profile for instance.
We can set the compatible res by default, but the problem is that this resolution is cropped. In this end we would like the user to be able to set a target resolution as [we do here](https://github.com/pollen-robotics/reachy_mini/blob/9c3839637dbdc0031c1a51727468a7c3da6a79c3/tests/test_video.py#L69). IMO the --profile-h264 args should be removed.

I think we have to set 1280x720@30 by default otherwise the osx users won't have the view in the app. Then it would be up to the user to change the resolution and framerate depending on the needs. Either we do the resolution control here, or we open another issue. I'm fine with both. What do you think @apirrone ?

apirrone:
I'm fine with both too, but we have to make sure to not break stuff, changing the default resolution can indeed make the image more cropped and break the hand tracking app for exmaple

FabienDanieau:
> I'm fine with both too, but we have to make sure to not break stuff, changing the default resolution can indeed make the image more cropped and break the hand tracking app for exmaple

Ok, since the hand tracking is not available on the wireless for now we can move forward.
Let for go option 1 and open an issue to allow the change of resolution and the calibration of all different resolutions.

tfrere:
Here are the modifications !

PR #649, opened 2025-12-29
Fix or greater bug : pause webrtc for better restart

cdussieux:


FabienDanieau:
Ok it seems to work, but it might be a bug on gstreamer side. I'll investigate. Meanwhile we can push this fix.
Why not simply calling pause here ? The function exists
https://github.com/pollen-robotics/reachy_mini/blob/ff36fadc0cda43558351cdea1e5f6289eeae7d1a/src/reachy_mini/daemon/daemon.py#L356
\+ comment to explain why not stop here

CarolinePascal:
Added @FabienDanieau comment in #655 !

PR #656, opened 2025-12-29
feat: add --mockup-sim mode (lightweight simulation without MuJoCo)

tfrere:
## Summary

Adds a new `--mockup-sim` flag to run the daemon in lightweight simulation mode without requiring MuJoCo.

## Motivation

- MuJoCo is a heavy dependency (~hundreds of MB)
- Some users just want to test the API/SDK without physics simulation
- Useful for CI/CD, quick prototyping, or running on systems where MuJoCo is hard to install

## Usage

```bash
# Lightweight simulation (no MuJoCo required)
reachy-mini-daemon --mockup-sim
```

## What works

- ✅ Full kinematics (FK/IK with AnalyticalKinematics)
- ✅ Zenoh publishing (joint positions, head pose)
- ✅ All REST API endpoints
- ✅ WebSocket streaming (`/api/state/ws/full`)
- ✅ Audio playback (wake_up, go_sleep sounds)
- ✅ Desktop app 3D viewer (via Zenoh/WebSocket)
- ✅ Webcam/microphone (apps open them directly, like with real robot)

## What's different from `--sim` (MuJoCo)

| Feature | --sim (MuJoCo) | --mockup-sim |
|---------|----------------|--------------|
| Kinematics (FK/IK) | ✅ | ✅ |
| Physics simulation | ✅ | ❌ |
| Collision detection | ✅ | ❌ |
| Native 3D viewer | ✅ | ❌ |
| Dependencies | mujoco | none |
| Motion dynamics | realistic | instant |
| Media access | UDP streaming | direct (like real robot) |

## Key design decision

In `--mockup-sim` mode, `simulation_enabled=False` is reported to apps. This makes apps behave exactly as they would with a real robot (opening webcam directly instead of expecting UDP streams). A separate `mockup_sim_enabled=True` flag allows internal distinction if needed.

## Changes

- New `MockupSimBackend` class (~150 lines)
- CLI argument `--mockup-sim` in `main.py`
- Backend selection logic in `daemon.py`
- Camera fallback to generic webcam (`cv2.VideoCapture(0)`) when no Reachy Mini camera found

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -17,12 +17,15 @@ jobs:
       - name: Install system dependencies (Ubuntu)
         if: matrix.os == 'ubuntu-latest'
         run: sudo apt-get update && sudo apt-get install -y libportaudio2
-      - name: Install dependencies
+      - name: Create virtual environment and install dependencies
         run: |
+          python -m venv venv
+          source venv/bin/activate
           python -m pip install --upgrade pip
           pip install .[dev]
       - name: Run tests
         run: |
+          source venv/bin/activate
           pytest -vv -m 'not audio and not video and not audio_gstreamer and not video_gstreamer and not wireless and not wireless_gstreamer' --tb=short
         env:
           MUJOCO_GL: disable
--- a/examples/recorded_moves_example.py
+++ b/examples/recorded_moves_example.py
@@ -10,6 +10,12 @@ import argparse
 from reachy_mini import ReachyMini
 from reachy_mini.motion.recorded_move import RecordedMove, RecordedMoves
 
+# Keep compatibility with the original library flag while allowing custom datasets.
+LIBRARY_DATASETS = {
+    "dance": "pollen-robotics/reachy-mini-dances-library",
+    "emotions": "pollen-robotics/reachy-mini-emotions-library",
+}
+
 
 def main(dataset_path: str) -> None:
     """Connect to Reachy and run the main demonstration loop."""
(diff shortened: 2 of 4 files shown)


src/reachy_mini/daemon/daemon.py
```diff
@@ -579,8 +597,14 @@ def _setup_backend(
         websocket_uri: Optional[str],
         hardware_config_filepath: str | None = None,
         reflash_motors_on_start: bool = True,
-    ) -> "RobotBackend | MujocoBackend":
-        if sim:
+    ) -> "RobotBackend | MujocoBackend | MockupSimBackend":
+        if mockup_sim:
```
pierre-rouanet, line 601: If we set both --mockup-sim and --sim it will start the mockup version. Maybe we can raise an error in such case to explicitly say to the user that only one should be set?
It's no big deal as we should re-work this whole parameters stuff to better reflect only the current version (maybe only a --version {lite, wireless, mujoco, mockup} ?)

src/reachy_mini/daemon/backend/mockup_sim/backend.py
```diff
@@ -0,0 +1,178 @@
...
+                self._head_joint_positions = self.target_head_joint_positions.copy()
+            if self.target_antenna_joint_positions is not None:
+                self._antenna_joint_positions = self.target_antenna_joint_positions.copy()
+
+            # Update current states
+            self.current_head_joint_positions = self._head_joint_positions.copy()
+            self.current_antenna_joint_positions = self._antenna_joint_positions.copy()
+
+            # Update kinematics model (computes FK)
+            self.update_head_kinematics_model(
```
pierre-rouanet, line 95: We probably can do some optimizations here as we know both the target and present. But let's keep it this way as   this will only be called in context where we have plenty of CPU power.

pierre-rouanet (approved):
I've added some minor comments, more to keep track for future refactoring.

So it looks ok for me! Or should I say "zog zog".

PR #664, opened 2025-12-30
659 auto detect lite vs wireless when using python sdk

RemiFabre:
Added a string-based `connection_mode` that can be either of these: 
- "auto" (default)
- "localhost_only" 
- "network"

In auto mode, the SDK client will first try to connect to a local daemon and if it fails, will try for a distant daemon. All our examples should now work out of the box regardless of Lite vs Wireless (currently we need to manually put localhost_only=False for examples to work on a Wireless). 

The old localhost_only parameter still works, but it now just logs a deprecation warning and defers to connection_mode when both are present.

To test this I used a Wireless unit (final product, on v1.2.6rc2) and a Beta Lite (on develop).
I ran "sequence.py" from my laptop and tried killing one or the other Mini (or having both alive at the same time).  It works as expected:
- If only the Lite is alive, the Lite moved.
- If only the Wireless alive, the Wireless moved.
- If both are alive, the Lite moved.

Then installed and ran the conversation app from the dashboard on the Wireless (just in case), everything worked as expected.

Examples that use media don't work on the wireless yet, WIP

### DIFF
--- a/docs/SDK/gstreamer-installation.md
+++ b/docs/SDK/gstreamer-installation.md
@@ -170,7 +170,7 @@ When installing Reachy Mini Python package, you will also need to add the `gstre
 ### Install from PyPI
 
 ```bash
-uv add reachy-mini --extra gstreamer
+uv pip install "reachy-mini[gstreamer]"
 ```
 
 ### Install from source


RemiFabre:
Media examples (look_at_image.py and recorded_moves_example.py) don't work from my computer on the wireless from develop either. What is the current way of doing this?

Edit: this seems to be a problem on my computer only. Can someone please test this?

src/reachy_mini/reachy_mini.py
```diff
@@ -49,12 +49,17 @@
     ]
 )
 
+ConnectionMode = Literal["auto", "localhost_only", "network"]
```
pierre-rouanet, line 52: You could have used an enum instead to avoid the manual check/norm, no?

src/reachy_mini/reachy_mini.py
```diff
@@ -181,6 +196,73 @@ def _configure_mediamanager(
...
+
+        return "localhost_only" if legacy_localhost_only else "network"
+
+    def _initialize_client(
+        self, requested_mode: ConnectionMode, timeout: float
+    ) -> tuple[ZenohClient, ConnectionMode]:
+        """Create a client according to the requested mode, adding auto fallback."""
+        if requested_mode == "auto":
+            try:
+                client = self._connect_single(localhost_only=True, timeout=timeout)
```
pierre-rouanet, line 239: maybe add a log here as well to always log the selected mode.

pierre-rouanet (approved):
Except for the minor comments, lgtm!

Works fine!
Yet, I've tested with a mujoco sim daemon running on my windows computer and the client on my Mac. So, not with a real robot (lite or wireless).

apirrone:
I ran `look_at_image.py` on my computer with my wireless running (both on this branch), and it worked without having to specify `localhost_only=False`. 🚢

RemiFabre:
> I ran `look_at_image.py` on my computer with my wireless running (both on this branch), and it worked without having to specify `localhost_only=False`. 🚢

Yes it seems that issue was on my computer only!

PR #670, opened 2025-12-31
Make get_ip_adress work on windows

pierre-rouanet:


FabienDanieau (approved):
Tested with
```
from reachy_mini.daemon.utils import get_ip_address
get_ip_address("Wi-Fi")
```
Is it used somewhere?

pierre-rouanet:
When spawning the daemon with --no-localhost-only.

PR #672, opened 2025-12-31
enhancement #6741: reduce usage of CI

FabienDanieau:
inspired from https://github.com/huggingface/lerobot/blob/main/.github/workflows/fast_tests.yml


cdussieux (approved):
lgtm

PR #673, opened 2025-12-31
654 bind imu in sdk for wireless

apirrone:
# Tests : 
- [x] wireless local 
- [x] wireless remote 
- [x] lite (should not break anything, but will not work)

# To test this PR : 

```bash
python3 examples/imu_example.py
```


FabienDanieau (approved):
Works nicely!
For the example I would check if the reachy mini is wireless before running the loop. If it's a lite just print a warning and exit.

cdussieux (approved):
tested on wireless

PR #678, opened 2026-01-02
fix(backend choice): relaxing constraints on backend choice for Lite and Wireless versions

CarolinePascal:
This PR adds back partial media backend choice, especially on the Lite version (where only the webrtc backend is not available) and for `no_video` backends.


CarolinePascal:
Tested with Wireless and Lite on macOS - merging !

PR #680, opened 2026-01-02
679 wireless apps always have a cpu at 100

FabienDanieau:
The CPU usage is due to the conversion of the camera frame from YUV2 to BGR when a gstreamer media backend is created. Even if the frame is not used, it is processed and then dropped.

I change the gstreamer element videoconvert, to v4l2convert which is better adapted for the rpi. A cpu overhead is still present though. -> check if a demanding app (conversation?) is still performing correctly

If the application doesn't need video, it is better to rely on the `gstreamer_no_video` backend. (this [PR should be merged](https://github.com/pollen-robotics/reachy_mini/pull/678) to use this feature).
-> set the `gstreamer_no_video `for an app that requires sound only (radio). No extra CPU usage should be reported.

### DIFF
--- a/src/reachy_mini/apps/templates/main.py.j2
+++ b/src/reachy_mini/apps/templates/main.py.j2
@@ -10,7 +10,8 @@ class {{ class_name }}(ReachyMiniApp):
     # Optional: URL to a custom configuration page for the app
     # eg. "http://localhost:8042"
     custom_app_url: str | None = "http://0.0.0.0:8042"
-    # Optional: specify a media backend ("gstreamer", "default", etc.)
+    # Optional: specify a media backend ("gstreamer", "gstreamer_no_video", "default", etc.)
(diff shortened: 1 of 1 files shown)


RemiFabre:
"check if a demanding app (conversation?) is still performing correctly"
-> done:
I tested this on a wireless unit. Ran the conversation app that worked fine, although the wobbling was laggy/non smooth still. I also started the emotion app and I still had a core at 100% CPU. I don't have a clear/easy way to see if this had a positive impact.

"I change the gstreamer element videoconvert, to v4l2convert which is better adapted for the rpi. A cpu overhead is still present though." --> Since you tested this, can you share some numbers so we can quantify the improvement?

Still this is a simple change that should have a positive impact, let's merge.

One question: is it a good idea to request a media backend instead of "default"? Like I wanted to request "gstreamer_no_video" in the emotions app, but that's only a good idea for a wireless unit no? As gstreamer is not always installed on laptops (and is a PITA to install, although this should change soon).

PR #688, opened 2026-01-06
Wireless connection mode

vkublytskyi:
This PR intend to improve developer experience of Reachy Mini Wireless owners.

### 🚀 Description

This PR introduces a new `wireless` connection mode to improve connectivity reliability, particularly on networks where UDP broadcast/multicast used for network discovery is blocked or unreliable (e.g., standard Google Fiber routers).

It also updates the `auto` connection logic in `src/reachy_mini/reachy_mini.py` to attempt a direct mDNS connection based on robot name (e.g. `reachy-mini.local` for Reachy Mini Wireless) before falling back to network discovery.

### 📝 Changes

- **New "Wireless" Mode**: Added `connection_mode="wireless"` to `ReachyMini` in `src/reachy_mini/reachy_mini.py`. This attempts a direct TCP connection to the robot using its mDNS hostname (`reachy-mini.local`).
- **Improved Auto-Connect Logic**: The `auto` mode fallback sequence in `src/reachy_mini/reachy_mini.py` has been updated to:
  - `localhost` (local daemon)
  - `wireless` (direct mDNS connection)
  - `network` (network discovery)
- **Zenoh Client Update**: Updated `ZenohClient` in `src/reachy_mini/io/zenoh_client.py` to accept a specific `host` string, enabling direct TCP connections to any target.

### ⚠️ Deprecations

- **`ZenohClient.__init__` in `src/reachy_mini/io/zenoh_client.py`**:
    - **Deprecated Argument**: `localhost_only`
    - **Reason**: Replaced by the more generic `host` argument to support specific remote targets.
    - **Behavior**: Usage will trigger a `logging.warning` but remains functional.

- **`ReachyMini._connect_single` (internal) in `src/reachy_mini/reachy_mini.py`**:
    - **Deprecated Argument**: `localhost_only`
    - **Reason**: Replaced by `host` to align with `ZenohClient` updates.
    - **Behavior**: Usage will trigger a `logging.warning` but remains functional.

### 🔗 Backward Compatibility

- **Fully Backward Compatible**: Existing code relying on the default `auto` mode will benefit from improved connectivity without changes.
- **Legacy Arguments**: Calls explicitly using `localhost_only=True` will continue to function correctly, emitting a warning to encourage migration.

### DIFF
--- a/src/reachy_mini/io/zenoh_client.py
+++ b/src/reachy_mini/io/zenoh_client.py
@@ -5,7 +5,6 @@ robot. It subscribes to joint positions updates and allows sending commands to t
 """
 
 import json
-import logging
 import threading
 import time
 from dataclasses import dataclass
@@ -24,27 +23,27 @@ from reachy_mini.io.protocol import AnyTaskRequest, TaskProgress, TaskRequest
 class ZenohClient(AbstractClient):
     """Zenoh client for Reachy Mini."""
 
-    def __init__(self, prefix: str, localhost_only: Optional[bool] = None, host: Optional[str] = None):
+    def __init__(self, prefix: str, localhost_only: Optional[bool] = None, host_addr: Optional[str] = None):
         """Initialize the Zenoh client.
 
         Args:
             prefix: The Zenoh prefix to use for communication (used to identify multiple robots).
-            localhost_only: (Deprecated) If True, connect to localhost only. Use 'host="localhost"' instead.
-            host: If specified, connect to this host only
+            localhost_only: If True, connect to localhost only.
+            host_addr: If specified, connect to this host only.
 
         """
-        self.logger = logging.getLogger(__name__)
         self.prefix = prefix
-        if host:
+        
+        if localhost_only and host_addr and host_addr not in ["localhost", "127.0.0.1"]:
+            raise ValueError("host_addr must be 'localhost' or '127.0.0.1' when localhost_only is True.")
+                    
+        if host_addr:
             c = zenoh.Config.from_json5(
                 json.dumps(
-                    {"mode": "client", "connect": {"endpoints": [f"tcp/{host}:7447"]}}
+                    {"mode": "client", "connect": {"endpoints": [f"tcp/{host_addr}:7447"]}}
                 )
             )
         elif localhost_only:
-            self.logger.warning(
-                "The 'localhost_only' argument is deprecated. Please use 'host=\"localhost\"' instead."
-            )
             c = zenoh.Config.from_json5(
                 json.dumps(
                     {"mode": "client", "connect": {"endpoints": ["tcp/localhost:7447"]}}
--- a/src/reachy_mini/io/zenoh_client.py
+++ b/src/reachy_mini/io/zenoh_client.py
@@ -29,13 +29,15 @@ class ZenohClient(AbstractClient):
 
         Args:
             prefix: The Zenoh prefix to use for communication (used to identify multiple robots).
-            localhost_only: If True, connect to localhost only.
-            host_addr: If specified, connect to this host only.
+            localhost_only: If True or None with no host_addr, connect to localhost only.
+            host_addr: If specified, connect to this host only. If None and localhost_only is False, connect to the first robot found on the network.   
 
         """
         self.prefix = prefix
         
-        if localhost_only and host_addr and host_addr not in ["localhost", "127.0.0.1"]:
+        if localhost_only is None and host_addr is None:
+            localhost_only = True # Preserve old behavior when called with only prefix
+        elif localhost_only and host_addr and host_addr not in ["localhost", "127.0.0.1"]:
             raise ValueError("host_addr must be 'localhost' or '127.0.0.1' when localhost_only is True.")
                     
         if host_addr:
--- a/src/reachy_mini/io/zenoh_client.py
+++ b/src/reachy_mini/io/zenoh_client.py
@@ -5,6 +5,7 @@ robot. It subscribes to joint positions updates and allows sending commands to t
 """
 
 import json
+import logging
 import threading
 import time
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


vkublytskyi:
Related discussions on Discord:
- https://discord.com/channels/519098054377340948/1452322061547475006/1452322065356029972
- https://discord.com/channels/519098054377340948/1452173450314780718/1452792539919945729
- https://discord.com/channels/519098054377340948/1451166295436296274/1451172994104950857

vkublytskyi:
Updated with upstream and fixed merge conflict.

pierre-rouanet:
Thanks @vkublytskyi, I'll try to have a look at it this week!

src/reachy_mini/io/zenoh_client.py
```diff
@@ -23,17 +24,27 @@
...
         """Initialize the Zenoh client.
 
         Args:
             prefix: The Zenoh prefix to use for communication (used to identify multiple robots).
-            localhost_only: If True, connect to localhost only
+            localhost_only: (Deprecated) If True, connect to localhost only. Use 'host="localhost"' instead.
+            host: If specified, connect to this host only
 
         """
+        self.logger = logging.getLogger(__name__)
```
andimarafioti: why instantiate a logger that you don't use? :S

src/reachy_mini/io/zenoh_client.py
```diff
@@ -23,17 +24,27 @@
...
     """Zenoh client for Reachy Mini."""
 
-    def __init__(self, prefix: str, localhost_only: bool = True):
+    def __init__(self, prefix: str, localhost_only: Optional[bool] = None, host: Optional[str] = None):
         """Initialize the Zenoh client.
 
         Args:
             prefix: The Zenoh prefix to use for communication (used to identify multiple robots).
-            localhost_only: If True, connect to localhost only
+            localhost_only: (Deprecated) If True, connect to localhost only. Use 'host="localhost"' instead.
```
andimarafioti: I wouldn't deprecate localhost, instead you can keep it and add the host_address

src/reachy_mini/io/zenoh_client.py
```diff
@@ -23,17 +24,27 @@
...
-
-        if localhost_only:
+        if host:
+            c = zenoh.Config.from_json5(
+                json.dumps(
+                    {"mode": "client", "connect": {"endpoints": [f"tcp/{host}:7447"]}}
+                )
+            )
+        elif localhost_only:
+            self.logger.warning(
```
andimarafioti: oh, here you use the logger. still, I would remove it to simplify

src/reachy_mini/reachy_mini.py
```diff
@@ -285,34 +286,70 @@ def _initialize_client(
...
+        # Backward compatibility for localhost_only argument passed as keyword
+        if "localhost_only" in kwargs:
+            self.logger.warning(
+                "The 'localhost_only' argument is deprecated. Please use 'host=\"localhost\"' instead."
+            )
+            if kwargs["localhost_only"]:
+                host = "localhost"
+
+        # Ensure host is treated as str or None for ZenohClient
+        host = host if isinstance(host, str) else None
```
andimarafioti: Here you are raising the deprecation warning twice. host should never be a bool, so you are assumming an intent in the user, which might not be correct

andimarafioti (commented):
some observations :)

src/reachy_mini/io/zenoh_client.py
vkublytskyi: Logger is used to send a warning about the `localhost_only` parameter deprecation. 
Deprecation, in turn, was introduced for backward compatibility. Implementation would be clearer without it, but from my experience, backward compatibility practices are highly important for open source projects with the open ecosystem like Reachy Mini apps.

An alternative is to use `warnings` libs or only comments, but the rest of the code uses logger for deprecation messages.

src/reachy_mini/io/zenoh_client.py
vkublytskyi: I deprecated `localhost_only` because, with the `host` or `host_address` parameter, client code can achieve the same functionality as `localhost_only`, making it appear excessive. IMO "lite" and "wireless" type should be first class concpet on ReachyMini class level (I cannot connect my wireless version in `localhost_only` mode), and Zenoh client should operate with the hosts and does not care what exactly that host is.

What is the reason for the Zenoh client to treat localhost in a different way than any other host?

src/reachy_mini/io/zenoh_client.py
vkublytskyi: If backward compatibility is not needed, I think, remove the old parameter, and the logger would be much cleaner.
If backward compatibility is needed, this is the way how deprectaion message is communicated in other parts of the code.

Please let me know what would be better for the code base from your perspective?

src/reachy_mini/reachy_mini.py
vkublytskyi: If there is existing code (e.g. somebody inherited ReachyMini class in their code) that calls `self._connect_single(True, 5.0)` that is a valid call, and with the new signature, it will result in `host` has bool value.

Please let me know if backward compatibility is not concern and I should simplify code.

andimarafioti: But `self._connect_single` is (per convention) a private method, so having this call would be hacky. On top of that, you are describing a posible situation, another one is that a suer does `self._connnect_single(host=True)` and gets this warning about `localhost_only` which is probably unrelated, the user is confused about what the host parameter does, and now you send it to `localhost` which won't work for wireless 😵. That's what I meant with "you're assumming the intent", your failure mode is real, but it's not the only one. I think this warning would be better with something like: `"The 'host' argument should be a string. Did you mean `localhost_only=True`? That parameter is deprecated and will be removed in the next major version, use "host='localhost'" instead."`

src/reachy_mini/io/zenoh_client.py
andimarafioti: fair

src/reachy_mini/io/zenoh_client.py
andimarafioti: The other issue I have with this (and I should have mentioned this earlier today, I'm sorry) is that if you search for "localhost_only" in the project you find 63 results. But I'm pretty sure that you didn't change it everywhere where it should be changed for this to make sense. Hence my point of, maybe it's easier to not deprecate it, or to work around it. Maybe if localhost_only -> we look for the host with the robot name or for "localhost" explictly

src/reachy_mini/reachy_mini.py
vkublytskyi: @andimarafioti yes, this make sense. Probably backward compatibility for private method indeed not needed. I also rewrote ZenohClient without deprecation and with throwing exception if values of `localhost_only` and `host_addr` conflicts with each other.

Please review if such changes more appropriate for the project.

vkublytskyi:
Closing this out as it's gone stale. Happy to revisit or open a fresh PR if there's interest down the line.

PR #689, opened 2026-01-06
fix(webrtc): Recursively search for webrtcbin to handle re-instantiation

vkublytskyi:
## Description
This PR addresses an issue where `webrtcbin` could not be found by name ("webrtcbin0") when the `ReachyMini` client is re-instantiated, such as when restarting a connection in a Jupyter notebook cell without restarting the kernel.

## Changes
- Modified `reachy_mini/src/reachy_mini/media/webrtc_client_gstreamer.py`:
  - Added `_iterate_gst` helper to iterate over GStreamer iterators.
  - Updated `_configure_webrtcbin` to include a fallback mechanism: if `get_by_name("webrtcbin0")` fails, it now performs a recursive search within the `webrtcsrc` bin to find an element with the factory name "webrtcbin".

## Motivation
In scenarios like re-running a notebook cell that establishes a connection to the robot, the internal GStreamer element names might change (e.g., incrementing to `webrtcbin1`), causing the hard-coded lookup for `webrtcbin0` to fail. This fix ensures robust retrieval of the `webrtcbin` element regardless of its dynamically assigned name.

## Testing
- Verified that connection can be established successfully on the first run.
- Verified that connection can be re-established by re-running the instantiation cell multiple times without errors.

### DIFF
--- a/src/reachy_mini/daemon/app/services/bluetooth/bluetooth_service.py
+++ b/src/reachy_mini/daemon/app/services/bluetooth/bluetooth_service.py
@@ -278,7 +278,7 @@ class CommandCharacteristic(Characteristic):
 
 
 class ResponseCharacteristic(Characteristic):
-    """Response Characteristic.""" ""
+    """Response Characteristic."""
 
     def __init__(self, bus, index, service):
         """Initialize the Response Characteristic."""


vkublytskyi:
@FabienDanieau I fixed linter issues.

PR #698, opened 2026-01-07
Report gstreamer latency

FabienDanieau:
Log the latencies of the gstreamer pipeline
Add few optimisations

### DIFF
--- a/src/reachy_mini/apps/sources/local_common_venv.py
+++ b/src/reachy_mini/apps/sources/local_common_venv.py
@@ -415,7 +415,7 @@ def _find_metadata_for_entry_point(ep_name: str) -> dict:  # type: ignore
 
             # Check 1: Normalized name match
             if normalize(space_name) == ep_normalized:
-                return file_metadata # type: ignore
+                return file_metadata  # type: ignore
 
             # Check 2: Entry point name appears in siblings (package structure)
             siblings = file_metadata.get("siblings", [])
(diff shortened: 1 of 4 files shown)


src/reachy_mini/daemon/app/services/wireless/launcher.sh
```diff
@@ -1,8 +1,4 @@
 #!/bin/bash
-# Get the directory where this script is located
-SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-
-"$SCRIPT_DIR/generate_asoundrc.sh"
```
CarolinePascal, line 5: Should we keep it just in case there is an issue and the .asoundrc file is not there ?

src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -183,6 +183,7 @@ def _init_pipeline_playback(self, pipeline: Gst.Pipeline) -> None:
         else:
             audiosink = Gst.ElementFactory.make("alsasink")
             audiosink.set_property("device", f"hw:{self._id_audio_card},0")
+            audiosink.set_property("buffer-time", 20000)
```
CarolinePascal, line 186: Can we also set `latency_time` at this stage ? This parameter really made a difference on macOS !

CarolinePascal: 5000 for 5ms worked nicely :)

src/reachy_mini/daemon/app/services/wireless/launcher.sh
FabienDanieau: the asoundrc is generated by the daemon, in audio_utils. Actually I've got an issue. After changing audio_utils this script overwrote old parameters. IMO it's better to keep only one method

src/reachy_mini/media/audio_gstreamer.py
FabienDanieau: done

PR #702, opened 2026-01-08
700 propagate motor controller stats to backendstatus

pierre-rouanet:
Add stats from the rust motor-controller in real robot backend status.


pierre-rouanet:
Yes, good idea!

pierre-rouanet:
Added a section on troubleshooting @FabienDanieau

Issue #687, opened 2026-01-05
Daemon fails to gracefully shut down on Wireless


alozowski:
### Ticket Type

🐛 Bug Report

### Robot Type

Wireless

### Operating System

Linux

### Browsers Used

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] Opera
- [ ] Other

### Troubleshooting

- [x] I have read https://github.com/pollen-robotics/reachy_mini/blob/develop/docs/troubleshooting.md

### Update Status

- [ ] My robot is up-to-date via the dashboard.
- [x] I'm running the latest Reachy Mini release from this repo.

### Description

The daemon encounters a race condition during shutdown when the backend's motor controller connection is already closed or encounters a "session closed" error. This results in multiple cascading errors and prevents graceful shutdown.

### Context & Reproduction

1. Start the daemon normally on RPI with `reachy-mini-daemon --wireless-version`
2. Wait until the robot is up and running
4. Press Ctrl+C to stop the daemon
5. Observe the cascading errors during shutdown

### Relevant logs or stack trace

```Shell
2026-01-05 23:02:27,402 - reachy_mini.daemon.daemon - INFO - Waking up Reachy Mini...
INFO:reachy_mini.daemon.daemon:Daemon started successfully.
2026-01-05 23:02:29,405 - reachy_mini.daemon.daemon - INFO - Daemon started successfully.
INFO:uvicorn.error:Application startup complete.
2026-01-05 23:02:29,406 - uvicorn.error - INFO - Application startup complete.
INFO:uvicorn.error:Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
2026-01-05 23:02:29,406 - uvicorn.error - INFO - Uvicorn running on http://0.0.0.0:8000 (Press CTRL+C to quit)
INFO:uvicorn.error:10.101.100.118:49648 - "WebSocket /api/move/ws/updates" [accepted]
2026-01-05 23:02:29,598 - uvicorn.error - INFO - 10.101.100.118:49648 - "WebSocket /api/move/ws/updates" [accepted]
INFO:uvicorn.error:connection open
2026-01-05 23:02:29,600 - uvicorn.error - INFO - connection open
^CINFO:uvicorn.error:Shutting down
2026-01-05 23:02:45,882 - uvicorn.error - INFO - Shutting down
INFO:uvicorn.error:connection closed
[...]

PR #703, opened 2026-01-08
fix: add null check in stop_recording to prevent daemon error state

tfrere:
## Problem
When the daemon tries to stop (e.g., after clicking Power Off on a WiFi-connected Reachy), it can get stuck in an ERROR state with the message:
```
'NoneType' object has no attribute 'close'
```

This happens because `_is_recording` can be `True` while `_input_stream` is `None`, causing the `close()` call to fail.

Once in ERROR state, the daemon cannot recover without a manual restart, preventing subsequent WiFi connections.

## Solution
- Add null check for `_input_stream` before calling `stop()`/`close()`
- Reset `_is_recording` flag after successful cleanup

## Testing
1. Connect to Reachy via WiFi
2. Put robot to sleep
3. Click Power Off
4. Verify daemon state is `stopped` (not `error`)
5. Reconnect via WiFi - should work

### DIFF
--- a/src/reachy_mini/media/audio_sounddevice.py
+++ b/src/reachy_mini/media/audio_sounddevice.py
@@ -228,8 +228,8 @@ class SoundDeviceAudio(AudioBase):
         See AudioBase.stop_recording() for complete documentation.
         """
         if self._input_stream is not None:
-            self._input_stream.stop()  # type: ignore[attr-defined]
-            self._input_stream.close()  # type: ignore[attr-defined]
+            self._input_stream.stop()
+            self._input_stream.close()
             self._input_stream = None
             self.logger.info("SoundDevice audio stream closed.")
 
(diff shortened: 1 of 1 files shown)


erwamartin:
I just ran into this issue today: 'NoneType' object has no attribute 'close'.

After that, I couldn’t put it into sleep mode and had to turn it back on, which caused the motors to force their position before returning to the initial position.

I recorded a video:

https://github.com/user-attachments/assets/4dbd3c2b-a137-47ae-874e-270dc9c26703

I assume this PR would fix this issue as well?

FabienDanieau:
> I just ran into this issue today: 'NoneType' object has no attribute 'close'.

Could you track down which object is None? Sounddevice is not supposed to be called on the wireless version.

We had a similar issue with None objects during the go to sleep which is solved and merged to develop. See https://github.com/pollen-robotics/reachy_mini/issues/687

tfrere:
Should be fixed in the last release !

erwamartin:
@FabienDanieau @tfrere the issue that I reported and recorded actually happened in version 1.2.8. Is it supposed to be fixed in this one?

PR #705, opened 2026-01-09
Custom install

cdussieux:
add a field to install using tag/branch from reachy-mini repo

enhance current version display (show version + source)

install it from source in your venv with `pip install --force-reinstall  git+https://github.com/pollen-robotics/reachy_mini.git@custom-install`
then you whell see correct source in settings (after restart)

you can try to install a specific branch/tag, and see that... well on another branch there is not this custom display so you may not know perfectly it worked unless you installed a tagged version, but at least you can try to rollback with it

### DIFF
--- a/docs/source/platforms/reachy_mini/install_daemon_from_branch.md
+++ b/docs/source/platforms/reachy_mini/install_daemon_from_branch.md
@@ -3,6 +3,8 @@
 > **⚠️ For Developers and Testers Only**
 >
 > This guide explains how to install the Reachy Mini daemon from a specific GitHub branch before it is officially released. Use this for testing new features or bug fixes.
+>
+> **💡 Tip:** Starting from v1.2.14, you can install a branch directly from **Settings → Reachy Mini Update** in the web dashboard.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


pierre-rouanet:
Installed it on my RM wireless and used it to change to develop. Works fine! ✅ 

Maybe add a message in the modal at the end of the install to remind people to reboot?

CarolinePascal:
I realize the issue I linked is not exactly solved by the PR : would it be possible to extend the PR to the [check_and_sync_apps_venv_sdk()](https://github.com/pollen-robotics/reachy_mini/blob/2b4d8c8f13660a0c07a75aaf361f3e8de64e8270/src/reachy_mini/utils/wireless_version/startup_check.py#L255) method ?

cdussieux:
@CarolinePascal in its current state, this PR does cover part of it. If the custom branch has been installed using this process, the app venv will also be updated.

What is not covered is if a user does a pip install from branch by hand, this is specifically something you would need ?
Did that, and seems ok. https://github.com/pollen-robotics/reachy_mini/pull/783 to cover the by-hand usecase

PR #706, opened 2026-01-09
Bump up huggingface hub

alozowski:
This PR migrates Hugging Face authentication to the huggingface_hub v1.x API (to the recent huggingface-hub 1.3.0 if specific), replacing deprecated token management utilities with the current, supported mechanisms.

> Note: only transformers>5.0.0 supports huggingface_hub>1.x, thus it's necessary to verify projects that require transformers (ex, [conversation app](https://github.com/pollen-robotics/reachy_mini_conversation_app))


alozowski:
I checked, only conversation app currently uses transformers, so we're safe to merge this PR and [PR 178](https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/178) on conversation app side

Issue #717, opened 2026-01-12
Apps don't show as Installed when running daemon directly when entry point name differs from space name


yaseminozkut:
### Ticket Type

🐛 Bug Report

### Robot Type

Wireless

### Operating System

macOS

### Browsers Used

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] Opera
- [ ] Other

### Troubleshooting

- [x] I have read https://github.com/pollen-robotics/reachy_mini/blob/develop/docs/troubleshooting.md

### Update Status

- [ ] My robot is up-to-date via the dashboard.
- [x] I'm running the latest Reachy Mini release from this repo.

## Problem                                                                                                                               
  When running the daemon directly (not the desktop app), some apps installed via entry points don't show as "Installed" in the app store UI at http://localhost:8000. The Install button remains clickable even after the app is already installed.  

This happens when the HuggingFace space name differs from the Python entry point name:
- Example 1: `reachy-mini-gemini` (HuggingFace space name) vs `reachy_mini_gemini_app` (Python entry point name) 
- Example 2: `reachy_mini_phone_teleop` (HuggingFace space name) vs `reachy-mini-phone-teleop` (Python entry point name)  

## Steps to reproduce: 
1- Run `python -m reachy_mini.daemon.app.main --sim`
2- Go to http://localhost:8000
3- Install an app that has different Hugging Face space name and pyproject.toml name (e.g., reachy_mini_gemini or reachy_mini_phone_teleop) 
4- After installation, the button still shows "Install" instead of "Installed"

## Expected Behavior
The Install button should toggle to "Installed" after installing an app.

### Relevant logs or stack trace

```Shell

```

PR #718, opened 2026-01-12
fix: apps not showing as Installed when entry point name differs from hugging face space name

yaseminozkut:
Fixes issue #717 

  - Save and load app metadata for entry-point based installs
  - Match apps by HuggingFace space ID (extra.id) instead of name
  - Add _find_metadata_for_entry_point() helper with multiple strategies


yaseminozkut:
> Thank you for your PR! This looks good for me.
> 
> Adding @cdussieux as reviewer as he is the one with deep understanding of the app install process!

Thank you for the review!

PR #724, opened 2026-01-14
651 improve error messaging when mini reachy is off

pierre-rouanet:
Make connection fail more explicit.

If localhost-only=True --> most likely the daemon is not running.
if auto --> either the daemon is not running or not accessible.

### DIFF
--- a/src/reachy_mini/reachy_mini.py
+++ b/src/reachy_mini/reachy_mini.py
@@ -298,7 +298,7 @@ class ReachyMini:
                 )
                 try:
                     client = self._connect_single(localhost_only=False, timeout=timeout)
-                except TimeoutError:
+                except (zenoh.ZError, TimeoutError):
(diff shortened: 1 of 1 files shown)


RemiFabre (approved):
Tested this on both lite and wireless and works as expected. My only concern is that in the case of localhost_only, you catch any exception and always raise TimeoutError. There could be some edge cases where this is problematic.

pierre-rouanet:
You're right! Is this new version better for you @RemiFabre ?

RemiFabre:
> You're right! Is this new version better for you @RemiFabre ?

Hum, I think still no because this:
```
self._connect_single
```
Can raise 2 types of exceptions: zenoh.ZError and TimeoutError, but you catch only one (you seem to catch the more likely tho).
Pushing a mini fix that covers both cases.

Please merge directly if this is good for you.

PR #731, opened 2026-01-15
GStreamer as default media backend

FabienDanieau:
Windows and MacOS are now supported. The new gstreamer wheels are automatically installed.
Linux users should still rely on a manual installation (see https://github.com/pollen-robotics/reachy_mini/blob/572-port-gstreamer-backend-to-windows/docs/SDK/gstreamer-installation.md).

Using the Lite should be similar as before.
Accessing the Wireless media stream is now possible with Windows and MacOs. 
`python examples/look_at_image.py `

@CarolinePascal I've made a [minor change](https://github.com/pollen-robotics/reachy_mini/blob/eef7e566d0543b079946100709f984e46032ba6d/src/reachy_mini/media/camera_gstreamer.py#L331) requested by mypy. Could you check that the camera is still well detected ? (for the Lite)

### DIFF
--- a/docs/source/API/media.mdx
+++ b/docs/source/API/media.mdx
@@ -11,7 +11,7 @@
 
 [[autodoc]] reachy_mini.media.audio_gstreamer.GStreamerAudio
 
-[[autodoc]] reachy_mini.media.audio_sounddevice.SoundDeviceAudio
+<!-- [[autodoc]] reachy_mini.media.audio_sounddevice.SoundDeviceAudio -->
 
 ### Audio Utils Functions
 


CarolinePascal:
- [x] Media CI on macOS (test_audio.py and test_video.py, except known resolution issues with webrtc)
- [x] Media CI on Linux
- [x] Media CI on Windows 
- [x] Update documentation !
- [x] `gstreamer` on pypi ⚠️ **mandatory** - This could prevent any further release of `reachy_mini` !
- [ ] Investigate timing (init) and cleanup (del) issues
- [ ] Remove no longer necessary sleeps after `get_audio_sample`
- [x] Rework examples !
- [ ] Consider starting playing/recording at startup ?

scienceetonnante:
I ran the conversation app (commit 7089ae6) on wireless on MacOS using this branch instead of Reachy-mini SDK.

I got an error trying to launch the app.

[error.txt](https://github.com/user-attachments/files/25078064/error.txt)

The error seems specific to MacOS
After a lot of tests and clauding, I have a fix on the app (but I don't know if this is the best place for it, so here is the fixed main.py from the conversation app

[main.py](https://github.com/user-attachments/files/25078067/main.py)

After that fix everything works both gradio and console.

Attached is description of the error and the fix written by Claude

[gstreamer-macos.md](https://github.com/user-attachments/files/25078134/gstreamer-macos.md)

alozowski:
small note: we solved the error with @scienceetonnante, it was the gstreamer installation on macos

PR #732, opened 2026-01-15
small improvement to the readme

andimarafioti:
Reordered the code snippet to come after the installation of the SDK instead of before. That should be a bit more clear.


CarolinePascal (approved):
LGTM !

PR #736, opened 2026-01-17
Remove early return optimization in set_motor_control_mode()

### LINKED ISSUE #735: enable_motors() silently fails after selective disable_motors
After selective disable_motors(ids=[...]), enable_motors() doesn't work.
The enable_motors() call is silently ignored due to a state tracking issue.

RemiFabre:
Test script to reproduce the bug:
```
#!/usr/bin/env python3
"""
Minimal reproduction of the motor enable/disable bug in reachy_mini SDK.

BUG: After selective disable_motors(ids=[...]), enable_motors() doesn't work.
     The enable_motors() call is silently ignored due to a state tracking issue.

EXPECTED: enable_motors() should re-enable all motors regardless of prior state.
ACTUAL: enable_motors() does nothing if motor_control_mode is already "Enabled",
        even though some motors were selectively disabled.

ROOT CAUSE (in reachy_mini/daemon/backend/robot/backend.py):
- set_motor_torque_ids() disables specific motors but doesn't update motor_control_mode
- set_motor_control_mode() has early return: if mode == self.motor_control_mode: return
- So enable_motors() -> set_motor_control_mode(Enabled) returns early (already "Enabled")

WORKAROUND: Call disable_motors() (all) before enable_motors()
"""

import time
from reachy_mini.reachy_mini import INIT_HEAD_POSE

# Same motors that fire_nation_attacked disables during gameplay
SELECTIVE_MOTORS = [
    "body_rotation",
    "stewart_1",
    "stewart_2",
    "stewart_3",
    "stewart_4",
    "stewart_5",
    "stewart_6",
]


def test_bug():
    print("=== Motor Enable/Disable Bug Reproduction ===\n")

    with ReachyMini() as mini:
        # Step 1: Start with all motors enabled
        print("[1] Enabling all motors...")
        mini.enable_motors()
        mini.goto_target(head=INIT_HEAD_POSE, antennas=[0.0, 0.0], duration=1.0)
        print("    Motors enabled, robot at INIT pose.\n")

        # Step 2: Selectively disable some motors (like fire_nation does)
        print(f"[2] Selectively disabling motors: {SELECTIVE_MOTORS}")
        mini.disable_motors(ids=SELECTIVE_MOTORS)
        print("    Head should now be compliant (you can move it by hand).\n")

        input("    >>> Press Enter to try enabling motors again...")

        # Step 3: Try to enable all motors - THIS IS THE BUG
        print("\n[3] Calling enable_motors() to re-enable all...")
        mini.enable_motors()
        print("    Trying to move to INIT pose...")
        mini.goto_target(head=INIT_HEAD_POSE, antennas=[0.0, 0.0], duration=1.0)

        print("\n[!] BUG: If the head is still compliant, enable_motors() didn't work!")
        print("    The robot should have moved to INIT pose but likely didn't.\n")

        input("    >>> Press Enter to apply the workaround...")

        # Step 4: Workaround - disable all first, then enable
        print("\n[4] WORKAROUND: disable_motors() then enable_motors()...")
        mini.disable_motors()  # <-- This resets motor_control_mode to Disabled
        mini.enable_motors()   # <-- Now this actually enables because mode changed
        mini.goto_target(head=INIT_HEAD_POSE, antennas=[0.0, 0.0], duration=1.0)
        print("    Robot should now move to INIT pose correctly.\n")

        print("[5] Test complete. Disabling motors...")
        mini.disable_motors()


if __name__ == "__main__":
    test_bug()

```

### DIFF
--- a/src/reachy_mini/daemon/backend/robot/backend.py
+++ b/src/reachy_mini/daemon/backend/robot/backend.py
@@ -544,10 +544,6 @@ class RobotBackend(Backend):
 
     def set_motor_control_mode(self, mode: MotorControlMode) -> None:
         """Set the motor control mode."""
-        # Check if the mode is already set
-        if mode == self.motor_control_mode:
-            return
-
         if mode == MotorControlMode.Enabled:
             if self.motor_control_mode == MotorControlMode.GravityCompensation:
                 # First, make sure we switch to position control


RemiFabre:
Root cause: set_motor_torque_ids() doesn't update motor_control_mode, so after selective disable the mode is still "Enabled".                              

So when we call enable_motors(), the call early-returns thinking motors were already on.

Here I went for a simple fix were we disable the optimization and always perform set_motor_control_mode (this makes it a state less call).

We could also make a more complex fix where there is an "intermediate" between "Enabled" and "Disabled" (when a subset of motors is enabled).
Opinions @pierre-rouanet ?

pierre-rouanet:
Yes I'm afraid your fix could result in some strange behavior. Imo, if we want to really provide the possibility to change torque for some motors, we should have this reflected with a more complex value (all on, all off, mixed) as you suggested.

RemiFabre:
Ok I'll look into this

RemiFabre:
After exploring this and discussing with Pierre, this won't be fixed right now as it will be covered in a future refacto.

PR #741, opened 2026-01-20
Fix/Improve camera calibration 

apirrone:
- [x] re calibrate camera for lite and wireless
- [x] Take into account crop factors for different camera resolutions
- [x] Implement set_resolution for webrtc streaming backend

This should also fix `examples/look_at_image.py` for all platforms and resolutions

### DIFF
--- a/src/reachy_mini/media/camera_utils.py
+++ b/src/reachy_mini/media/camera_utils.py
@@ -203,7 +203,7 @@ def scale_intrinsics(
         K_scaled: Adjusted camera matrix for target resolution
 
     """
-    K_scaled = K_original.copy()
+    K_scaled: npt.NDArray[np.float64] = K_original.copy()
 
     orig_w, orig_h = original_size
     target_w, target_h = target_size
--- a/src/reachy_mini/tools/camera_calibration/analyze_crop.py
+++ b/src/reachy_mini/tools/camera_calibration/analyze_crop.py
@@ -47,11 +47,13 @@ def analyze_image(
     for i, marker_id in enumerate(marker_ids):
         corners = marker_corners[i][0]
         center = corners.mean(axis=0)
-        marker_centers_pixels[int(marker_id[0])] = center  # type: ignore
+        marker_centers_pixels[int(marker_id[0])] = center
         # Normalize to image dimensions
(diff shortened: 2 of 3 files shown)


examples/look_at_image.py
```diff
@@ -32,6 +34,9 @@ def main(backend: str) -> None:
     print("Click on the image to make ReachyMini look at that point.")
     print("Press 'q' to quit the camera feed.")
     with ReachyMini(media_backend=backend) as reachy_mini:
+        # reachy_mini.media.camera.close()
+        # reachy_mini.media.camera.set_resolution(CameraResolution.R3072x1728at10fps)
```
FabienDanieau, line 41: this is for example purposes right? you may add a word like uncomment to change the resolution

src/reachy_mini/media/webrtc_client_gstreamer.py
```diff
@@ -156,7 +156,18 @@ def __del__(self) -> None:
...
         super().set_resolution(resolution)
+
+        # Check if pipeline is not playing before changing resolution
+        if self._pipeline_record.get_state(0).state == Gst.State.PLAYING:
+            raise RuntimeError(
+                "Cannot change resolution while the camera is streaming. Please close the camera first."
+            )
+
         self._resolution = resolution
+        caps_video = Gst.Caps.from_string(
```
FabienDanieau, line 167: this actually doesn't not change the resolution because it is set by the sender. The original resolution will be rescaled locally.
It's ok I'll open another issue for that since it doesn't break anything

src/reachy_mini/tools/camera_calibration/.gitignore
```diff
@@ -0,0 +1,4 @@
+images/*
```
FabienDanieau: I'd prefer to use the .gitignore at the root

src/reachy_mini/media/webrtc_client_gstreamer.py
apirrone: okay

src/reachy_mini/tools/camera_calibration/.gitignore
apirrone: okay

PR #743, opened 2026-01-21
Plugin hook system, that would allow external libraries access to hoo…

darkseed:
### Overview

This PR introduces a minimal, non-breaking plugin system that allows external packages to receive notifications when moves (emotions, dances, etc.) start and end. This enables community extensions like LED control, logging, or custom behaviors without modifying the SDK code.

### Motivation

Community members want to extend Reachy Mini's behavior (e.g., LED eye colors during emotions, custom logging, home automation triggers) without forking the SDK. This plugin system provides a clean, Pythonic way to add extensibility while maintaining backward compatibility.

### Changes

**New Files:**
- `src/reachy_mini/motion/hooks.py` - Core hook system with entry point discovery
- `tests/test_hooks.py` - Comprehensive test suite (13 test cases)
- `docs/motion_hooks_plugin_guide.md` - Complete guide for plugin authors
- `MOTION_HOOKS_IMPLEMENTATION.md` - Technical implementation summary

**Modified Files:**
- `src/reachy_mini/daemon/backend/abstract.py` - Integrated hooks into `Backend.play_move()`
- `src/reachy_mini/reachy_mini.py` - Integrated hooks into `ReachyMini.async_play_move()`
- `src/reachy_mini/motion/recorded_move.py` - Added `dataset_name` attribute for context
- `src/reachy_mini/motion/__init__.py` - Exported `MoveHook` for plugin authors

### How It Works

1. **Plugin Discovery**: Plugins register via Python entry points in their `pyproject.toml`
2. **Automatic Loading**: Hooks load automatically during SDK initialization
3. **Event Notifications**: Plugins receive `on_move_start()` and `on_move_end()` callbacks
4. **Graceful Errors**: Plugin failures are logged but don't crash the application

### Example Usage

**Plugin Author:**

`my_led_plugin/hook.py`

```python
class LEDHook:
  def on_move_start(self, move_name, dataset):
    if dataset and "emotions" in dataset:
      self.set_led_color(move_name)
      
  def on_move_end(self, move_name, dataset):
    self.reset_led()
```

**Plugin Registration:**
```text
# pyproject.toml
[project.entry-points."reachy_mini.motion.hooks"]
led_eyes = "my_led_plugin.hook:LEDHook"
```

**End User:**

```bash
pip install reachy-mini-led-eyes
# Plugin automatically loads - zero configuration needed!
```

## Key Features
- Zero-config: Automatic plugin discovery via entry points
- Non-breaking: No changes to existing APIs
[...]

PR #755, opened 2026-01-22
Component Runtime

brainwavecoder9:
A dynamic component runtime that gives 3rd parties the option to bring their own hardware (or software tbh). Think: broad extensibility with external modules, without modifying daemon or SDK code. This is first-class integration for third-party devices. 

The framework is laid out BUT I'm broadly flexible and happy to make further refinements/decisions and any necessary changes to align with the team's expectations. 

I've established some baseline requirements:  
- 	**The robot always wins** - Component issues will never impact daemon startup
- 	**There is a single source of truth** - Loading decisions are explicitly declared 
- 	**Components cannot break each other** - Isolated loading, no cross-dependencies, no direct interactions
- 	**Provenance must be known** - Must be able to trace the origin of every component
- 	**The daemon owns the lifecycle** - Components are always passive

From there, we can layer on some conceptual expectations:

- 	**Components are libraries** - They only expose APIs
- 	**One isolated namespace** - Explicit optionality, no collisions, clear delineation between 3rd party and official
- 	**Applications orchestrate** - All the things happen through apps, components simply expose the device
	
That keeps the implementation simple, with the entire process looking something like this: 
1.  Devs publish components to HuggingFace Spaces
2.  Users declare in a manifest
3.  Daemon loads into an isolated namespace
4.  Apps access components VIA a dedicated robot interface (e.g., robot.component["third-party-thing"])

For Example: 

	### Minimal component definition (published Hugging Face space)
	my-component/
	├── __init__.py
	└── (component owner's implementation files)

	### Daemon's manifest (e.g., components.txt, refs HF Space repo tag): 
	my-component=username/my-component:v1.0.0
	
	### Dynamic auto-loader provides via component namespace 
	robot.component["my-component"]

### Design rationale:

While we could support other sources or local folders, this type of inclusion can introduce additional challenges (management overhead, troubleshooting and debugging, security). A single, simple, manifest-driven source of truth, means we can turn things off/on very easily without having to know or understand any of the details. A text file as a control point means less-technical people have a simple installation process (also easy to add to dashboard as toggles).

The dedicated namespace means we don't have to be concerned about collisions or confusion, and the overhead is never a burden for the team. Dynamic loading is automatic and super straightforward (I've added placeholder comments to the daemon).  If there's a problem it's always a component problem... the *entire robot.component namespace is the responsibility of the corresponding 3rd party*. For issues Pollen can always simply defer with something like **_"We are not involved with 3rd party components, you will need to speak with the developer of that component"_** 

Component owners having to maintain this independently is a small price to pay. Devs get fully integrated access to their devices, the robot is never impacted, and Pollen doesn't have to know or care about those external integrations.

PyPi / GitHub / etc. are also possible, but given the low level access I think Hugging Face exclusivity makes sense, because it's the only place that provides a direct lever for HF (of all options it's the only one where you can pull the plug directly). That said, happy to enable whatever you want.

As for developers, requiring a component factory to be returned means they can implement as they please. This gives broad flexibility and control for 3rd party inclusions, without compromising the robot, and still maintaining a control point for HF. This unlock allows the community to freely extend Reachy without requiring Pollen's involvement.

Example: A researcher publishes alice/reachy-imu:v1.0.0 to add IMU support. Users add one line to their manifest. No review, no merge, no waiting, and no additional setup; apps then have direct access to the device through the robot interface. The contract is simple device exposure without any coupling.

### Next steps:

**Seeking additional thoughts and confirmation on approach.** The code is pretty simple (see included changes) but I'll complete the draft when we've confirmed how we want to go about this.

Let me know if you have any questions, would like further clarification or changes, or whatever else I can do to make this easier to explore.

### DIFF
--- a/src/reachy_mini/daemon/daemon.py
+++ b/src/reachy_mini/daemon/daemon.py
@@ -12,6 +12,7 @@ import time
 from dataclasses import asdict, dataclass
 from enum import Enum
 from importlib.metadata import PackageNotFoundError, version
+from pathlib import Path
 from threading import Event, Thread
 from typing import Any, Optional
 
@@ -59,6 +60,11 @@ class Daemon:
         self.desktop_app_daemon = desktop_app_daemon
 
         self.backend: "RobotBackend | MujocoBackend | MockupSimBackend | None" = None
+        
+        # === Storage for component registry -==
+        
+        self.components: dict[str, Any] = {}
+        
         # Get package version
         try:
             package_version = version("reachy_mini")
@@ -281,6 +287,14 @@ class Daemon:
             self._status.error = self.backend.error
             return self._status.state
 
+
+        # === Load All Components ===        
+        
+        self.components = self._load_external_components(
+            manifest_path=Path.home() / ".reachy-mini" / "components.txt",
+            simulation=sim or mockup_sim,
+        )
+        
         if wake_up_on_start:
             try:
                 self.logger.info("Waking up Reachy Mini...")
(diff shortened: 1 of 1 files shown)


brainwavecoder9:
**TL;DR: The component defines its own API, an the daemon exposes it through the existing robot interface.**

The contract:
- Factory supplied with **context** = {"simulation": bool}.
- A **Component** must be returned, but it's composition is arbitrary.
- **Version** must be available as a string.
- **cleanup()** is the only expected method, and is optional.

brainwavecoder9:
FYI since it came up, this approach would support any source, including non-Python apps, additional languages, etc. Component maintainers are able to bind whatever they want in init, including other languages or sources.

A less obvious unlock is that this also means components are infinitely composable. There is no requirement to supply a complete library. A component can be a granular module, and an app can be a collection of calls to robot/integrated component/multiple standalone components.

FabienDanieau:
Thanks, that's an interesting discussion to have. What do you think of the plugin system proposed here #743 ?

brainwavecoder9:
> Thanks, that's an interesting discussion to have. What do you think of the plugin system proposed here #743 ?

I can see how it could be useful for motion-based events. It’s not a problem I'm focused on, and I wasn’t involved in the related discussion or development. #743 addresses a narrower need in a different problem space. I would consider it separately if there is a desire for tighter component/robot integration.

I'm envisioning something far more ambitious and less restrictive. Creating an extensible surface would allow external developers to explore freely without asking Pollen to take on any additional burdens.

PR #763, opened 2026-01-23
App assistant now also checks the "src" layout & allows an app to be named differently from the package

scienceetonnante:
**BEFORE:** 
`reachy_mini_app_assistant check` was checking only a flat layout, with the package named similar to the app. 

For instance "my-app" had to have the following structure.
```
my_app/
└── my_app/
    └── main.py
```

There was two problems with :
   - the Conversation App was not checked as valid because it had a "src" layout
   - forking the conversation app was cumbersome since the app had to be named after the package

**NOW:**
The app assistant check will also check and validate a layout like

```
my_app/
└── src/
    └── my_package/
        └── main.py
```

The first commit introduces checking for the src/ layout
The second commit allows the app name to be different from the package name. Note that the class name is still expected to be a CamelCase version of the package name.

(1) Now the conversation app is validated by the assistant app check because the src/ layout is also checked (first commit)
(2) We can clone it, naming the app differently but keeping the package name

This has been checked on the reachy_mini_astronomer app which is a simple clone of the conversation app see https://huggingface.co/spaces/dlouapre/reachy_mini_astronomer/tree/main

The astronomer app has the following structure

```
reachy_mini_astronomer/
└── src/
    └── reachy_mini_conversation_app/
        └── main.py
```
(class name in main.py is still ReachyMiniConversationApp)

### DIFF
--- a/src/reachy_mini/apps/assistant.py
+++ b/src/reachy_mini/apps/assistant.py
@@ -280,7 +280,20 @@ def check(console: Console, app_path: str) -> None:
             sys.exit(1)
 
     entrypoint_name = app_name
-    pkg_name = app_name.replace("-", "_")
+
+    # Extract pkg_name from entry point (allows package name != project name)
+    entry_points = (
+        pyproject_content.get("project", {})
+        .get("entry-points", {})
+        .get("reachy_mini_apps", {})
+    )
+    app_name_normalized = app_name.replace("-", "_")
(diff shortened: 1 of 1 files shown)


apirrone (approved):
LGTM, tested: 
- checked a few older apps
- installed reachy mini astronomer, confirmed that it installs/uninstalls and shows up no problem in the dashboard (changing the package structure/names could break this)

PR #772, opened 2026-01-27
Reduce number of write calls by checking last send position

haixuanTao:
This PR makes it possible to check for last set **target** head and antenna position to avoid resetting target position despite it not changing. This reduces overall load of CPU / energy.

This also makes it possible to use write raw method for directly interacting with the motors. 

This also makes the behavior of the motor more consistent whether the daemon is on or off.

### DIFF
--- a/src/reachy_mini/daemon/backend/robot/backend.py
+++ b/src/reachy_mini/daemon/backend/robot/backend.py
@@ -109,6 +109,10 @@ class RobotBackend(Backend):
         self.target_antenna_joint_current = None  # Placeholder for antenna joint torque
         self.target_head_joint_current = None  # Placeholder for head joint torque
 
+        # Track last sent positions to avoid redundant writes
+        self._last_sent_head_positions = None
+        self._last_sent_antenna_positions = None
+
         if hardware_error_check_frequency <= 0:
             raise ValueError(
                 "hardware_error_check_frequency must be positive and non-zero (Hz)."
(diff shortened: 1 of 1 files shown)


haixuanTao:
This should also reduce the risk of poisoned error and having a failure on read/write

haixuanTao:
@apirrone @pierre-rouanet @FabienDanieau just wondering is there any issue with this PR?

pierre-rouanet:
Hi @haixuanTao, I'd assume the CPU reduction would be pretty marginal no? But I guess it's mainly interesting for you raw write approach.
We are starting a "dry-january" re-work (cc by @FabienDanieau)  to simplify the interface you can use for the daemon. We will have to see what suits you best for what you are doing but to be honest I'm not sure how I feel about multiplying the possible approaches in the core at the moment.

haixuanTao:
@pierre-rouanet Yes, it's more ot fix an issue with being able to control it raw.

But i genuinely believe that each write carry its risk of read/write failure as well as I/O overhead. 

For your information, the idea of the raw method is to be able to have the exact same API to a webserial API if for whatever reason we can't use the daemon on the lite version. 

Reason for not being able to use the daemon is that there might be some security restriction,  hardware issue,  python installation issue, driver issue, connection issue ( for example in china ). This creates at the very least a plan B that we can use. 

There's is fundamentally no issue with the raw method nor the daemonless webserial version.

There is just this PR conflict on read/write when the daemon is set to a position.

haixuanTao:
raw method is also the only method for anyone not just me that want to build something that could work on lite or wireless without having to use the daemon

haixuanTao:
@pierre-rouanet Could be great to have clarity on if this is ever going to be merged as we also have https://github.com/pollen-robotics/reachy-mini-desktop-app/pull/123 depending on it and if not we should probably close both.

pierre-rouanet:
I don't think we should focus on adding new ways of communicating with the robot but more focusing on making sure the existing ones are working really well! And are really well integrated into the desktop app.
For the pure js apps, @cdussieux made a really neat example here: https://huggingface.co/spaces/cduss/webrtc_example

Closing the issue for now and we will make a roadmap project soon to explain more.

Issue #777, opened 2026-01-28
Error to start apps that has name mismatch between hugging face name and entry point name (seperate venvs)


yaseminozkut:
### Ticket Type

🐛 Bug Report

### Robot Type

Lite

### Operating System

macOS

### Browsers Used

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] Opera
- [ ] Other

### Troubleshooting

- [ ] I have read https://github.com/pollen-robotics/reachy_mini/blob/develop/docs/troubleshooting.md

### Update Status

- [ ] My robot is up-to-date via the dashboard.
- [ ] I'm running the latest Reachy Mini release from this repo.

### Description

in `get_app_module`, when its using separate venvs, to get the app, it is using the app name from hugging face space name to match it with the entry point. if the HF space name does not match the entry point, then the app fails to start even though its installed
#### Logs from terminal: 
```
Sidecar stderr:   File "/Users/yaseminozkut/Dev/Reachy/reachy_mini/src/reachy_mini/apps/manager.py", line 103, in start_app
Sidecar stderr:     module_name = local_common_venv.get_app_module(
Sidecar stderr:   File "/Users/yaseminozkut/Dev/Reachy/reachy_mini/src/reachy_mini/apps/sources/local_common_venv.py", line 719, in get_app_module
Sidecar stderr:     raise ValueError(f"No entry point found for app '{app_name}'")
Sidecar stderr: ValueError: No entry point found for app 'reachy_mini_phone_teleop'
```

#### Logs from desktop app:

<img width="852" height="583" alt="Image" src="https://github.com/user-attachments/assets/f300f302-7d14-4c52-bc86-6479358f4e5e" />

### Context & Reproduction

To reproduce, from reachy mini desktop app:
- install reachy_mini_phone_teleop or reachy-mini-gemini app
- try to start them (it will install but fail to start because it won't find entry point)

### Relevant logs or stack trace

```Shell

```

PR #778, opened 2026-01-28
fix: app startup failure for name mismatches in separate venvs

yaseminozkut:
Fixes issue #777 
- added `_extract_package_info_from_pyproject` function to extract the real entry point from `pyproject.toml` during installation 
- use the stored `entry_point_name` instead of the space name when loading apps from separate venvs
- add fallback to app name for the apps installed before this fix so that users won't have problem (at least for the apps that don't have mismatch. otherwise, they will fail after this fix and they would need to uninstall and install the app because we are checking different variable from the metadata)

### DIFF
--- a/src/reachy_mini/apps/sources/local_common_venv.py
+++ b/src/reachy_mini/apps/sources/local_common_venv.py
@@ -442,7 +442,7 @@ def _find_metadata_for_entry_point(ep_name: str) -> dict:  # type: ignore
 
 def _extract_package_info_from_pyproject(target_path: str) -> dict[str, str | None]:
     """Extract package name and entry point name from pyproject.toml file."""
-    import toml  # type: ignore
+    import toml
 
     pyproject_path = Path(target_path) / "pyproject.toml"
 
--- a/src/reachy_mini/apps/sources/local_common_venv.py
+++ b/src/reachy_mini/apps/sources/local_common_venv.py
@@ -440,6 +440,38 @@ def _find_metadata_for_entry_point(ep_name: str) -> dict:  # type: ignore
     return {}
 
 
+def _extract_package_info_from_pyproject(target_path: str) -> dict[str, str | None]:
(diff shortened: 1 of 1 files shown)


CarolinePascal:
Hi @yaseminozkut,

Thank you for the contribution ! However, after discussing with the team, we would rather enforce that the hugging face space name always matches the entry point name. It's already the case when generating an app with `reachy-mini-app-assistant create` and we will make the app checks more strict to ensure the names are the same.

Let us know if you still think there is an issue to solve there.

Best,

Caroline.

Jacid23:
problem is there needs to be a name entry point registry because now the ui has no clue what one install the main or the fork. What a nightmare trying to figure out why diff space names showed up as different space install in the ui. People fork these things and honestly shouldn't be able to publish with the same name or something gotta change

yaseminozkut:
Hi @CarolinePascal,
Thanks for your comment! Yes checking the Hugging Face space name make sense. But if the entry point name for the package and the space name does not match, then even though the desktop app installs, it cannot find it and fails to start. I saw few apps had this error when I tried to start them. Maybe they are old (made before the enforcing) and the apps itself need to be fixed anyway. 

@Jacid23 also raised a good point about a different problem for forks sharing the same entry point names. I haven't tested this scenario myself, maybe @Jacid23 can give more explanation. Does it fail to start or start the wrong app, or is it just a UI issue (not clear who's app is it)? If its just a UI issue, maybe usernames from Hugging Face can be added in the UI. 

I also saw this discussion https://huggingface.co/datasets/pollen-robotics/reachy-mini-official-app-store/discussions/20 that @pierre-rouanet mentioned (maybe its before the enforcing with the `reachy-mini-app-assistant check` ?)

Best,
Yasemin

CarolinePascal:
Also pinging @apirrone and @RemiFabre for visibility !

PR #781, opened 2026-01-29
Simplify the SDK by moving the remote control code to an app.

andimarafioti:
I moved all remote control code to an app, we tested it and it's working, so we can now remove the code from the SDK to streamline it.


haixuanTao:
Ok so should people use Webrtc if they want to connect through the web to the robot?

FabienDanieau:
> Ok so should people use Webrtc if they want to connect through the web to the robot?
Using the SDK with gstreamer should do that for you. And we'll provide examples with a js app

PR #782, opened 2026-01-29
779 allow for motor data streaming via webrtc statecommand

pierre-rouanet:
WIP, more to open discussion/testing rather than merging now.

I add a DataChannel to the existing GstWebrtc. 

I made some really basic connections with the backend (only head_pose in state, only set_target head in command). My idea was
- first, we can use this for testing/validating the process
- it will make more sense to me, to make the proper integration with the backend as we clean up the daemon/backend. Right now, it's completely tangled between backend/zenoh/routers.

What do you say @FabienDanieau? @cdussieux if you want to play as well, you should have all the basics needed for your space.

--- 

How to test/use

* install this branch on your wireless
* restart daemon service
* start daemon for the webrtc loop to start
* run the examples/debug/gstreamer_client.py (it will make the robot do a gentle sinus on yaw and print current state)

### DIFF
--- a/examples/debug/gstreamer_client.py
+++ b/examples/debug/gstreamer_client.py
@@ -1,6 +1,7 @@
 """Simple gstreamer webrtc consumer example."""
 
 import argparse
+import time
 from threading import Thread
 
 import gi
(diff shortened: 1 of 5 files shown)


cdussieux:
> I made some really basic connections with the backend (only head_pose in state, only set_target head in command). My idea was
> 
> * first, we can use this for testing/validating the process
> * it will make more sense to me, to make the proper integration with the backend as we clean up the daemon/backend. Right now, it's completely tangled between backend/zenoh/routers.

This step-by-step approach does make sense; for me this is the right way to do this. I'll let you know as soon as u have the whole pipe with space

pierre-rouanet:
Closed in favor of #797

PR #785, opened 2026-01-30
Give forked conversation apps unique Python package names

RemiFabre:
- Rename package folder from `reachy_mini_conversation_app` to `{app_name}` after cloning
- Update all imports in `.py` files to use the new package name
- Rename `ReachyMiniConversationApp` class to `{AppName}` (PascalCase)
- Update pyproject.toml entry points with correct package and class names
- Update profile templates to use Jinja variable for package name

### DIFF
--- a/skills/create-app.md
+++ b/skills/create-app.md
@@ -119,3 +119,79 @@ When planning the app, consider which patterns apply:
 | Control loop | Real-time reactivity needed | `reachy_mini_conversation_app/moves.py` |
 | Head as controller | Games, recording | `fire_nation_attacked`, `marionette` |
 | LLM integration | AI-powered behavior | `reachy_mini_conversation_app` |
+
+---
+
+## Creating a Beautiful Landing Page (index.html)
+
+The root `index.html` is the landing page shown on Hugging Face Spaces. A well-designed landing page makes your app look professional and helps users understand what it does.
+
+**Reference template:** Use the [Marionette app](https://huggingface.co/spaces/RemiFabre/marionette) as a template for the structure and styling.
+
+### Structure
+
+A good landing page has three sections:
+
+1. **Hero Section** - Video/image + title + description
+2. **Technical Section** - "How it works" steps + features
+3. **Footer** - Resources, links, social media
+
+### Key Elements
+
+```
+index.html
+├── Hero Section
+│   ├── Demo video (autoplay, loop, muted)
+│   ├── App emoji + title
+│   ├── Tags (categories)
+│   └── Short description (1-2 sentences)
+│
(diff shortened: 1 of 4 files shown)


RemiFabre:
Copilot and Claude having a heated argument over regexes, just like us humans back in the day :heart: 

  1. Regex word boundary (\b) with underscores                                                      
                                                                                                    
  This is a non-issue. In Python regex, _ is a word character, so \breachy_mini_conversation_app\b  
  correctly matches the full package name and won't match partial strings like                      
  reachy_mini_conversation_app_extra. The suggested pattern is overly complex.                      
                                                                                                    
  2. Class name over-replacement in comments/docstrings                                             
                                                                                                    
  Valid theoretical concern, but acceptable in practice:                                            
  - If a comment says "replaces ReachyMiniConversationApp", it should be updated for the forked app 
  - The string is specific enough that false positives are unlikely                                 
  - Consistency throughout the codebase is desirable                                                
                                                                                                    
  My take: both suggestions are overly cautious. The current implementation works correctly for the 
  use case. The Copilot suggestions would add complexity without meaningful benefit.

RemiFabre:
@scienceetonnante I retested this from scratch and I believe this implementation is fine (I don't think doing what copilot says is worth it in this case)

scienceetonnante:
Thanks for taking care of this ! I thought changing the package name would cause too much side effects, but apparently you managed to do it :)

Tested here and it works.

My only minor remark is that the pyproject.toml still contains the following line

```
[tool.setuptools.package-data]
reachy_mini_conversation_app = [ "images/*", "static/*", ".env.example", "demos/**/*.txt", "prompts_library/*.txt", "profiles/**/*.txt", "prompts/**/*.txt",]
```

should `reachy_mini_conversation_app` be changed to the new package name here ?

RemiFabre:
Good catch! The pushed version should fix this

RemiFabre:
I also added an improvement to the skill of app creation to create a beautiful front page for the space (based on the template made by AnneCha, like this https://huggingface.co/spaces/RemiFabre/marionette )

And I couldn't make the exact instructions printed in the "Next steps" work for me, but it works when I use pip, so I changed it to:

```
Next steps:
  cd /home/remi/reachy_mini_apps/tempconv6
  pip install -e .
  reachy-mini-app-assistant check .

To test your app locally:
  python src/tempconv6/main.py --gradio
  # Then open: http://127.0.0.1:7861/
```

I'm merging now because I think the bug mentioned here https://discord.com/channels/519098054377340948/1467973428697891108/1467974128677163182 is our fault and will be fixed with this PR. But feel free to revert/improve any of this

PR #786, opened 2026-02-01
feat: add range for supported sounddevice versions

yeungadrian:
# Add range for supported sounddevice versions

I'm trying to build a voice agent with my reachy mini using only local models and running into compatibiltiy issues with mlx-audio.

Unfortunately the reachy-mini package is currently incompatible with mlx-audio due to the sounddevice dependency (mlx-audio forces 0.5.3 and reachy-mini currently forcing 0.5.1). Proposing to allow a slightly more flexible sounddevice version to solve this.

Reading the latest changes for sounddevice https://github.com/spatialaudio/python-sounddevice/releases and it seems quite safe to allow versions greater than 0.5.1


CarolinePascal (approved):
Looks good to me, thanks for the contribution !

PR #798, opened 2026-02-05
fix(gstreamer): adding gstreamer in daemon's update pip install command

CarolinePascal:
This PR adds the missing `gstreamer` dependency in the `pip install` command called when updating the daemon's venv `reachy_mini` package.


FabienDanieau (approved):
Good catch thanks!

PR #799, opened 2026-02-05
fix(git lfs): fixing uv's git lfs issue by setting the env variable UV_GIT_LFS to 1 when installing an app

CarolinePascal:
This PR fixes `uv` issues with `git lfs` when calling `uv pip install` commands. It simply sets the environment variable `UV_GIT_LFS` to 1 (c.f. [related doc](https://docs.astral.sh/uv/concepts/projects/dependencies/#index))

⚠️ This PR depends on #830 !

### DIFF
--- a/src/reachy_mini/utils/wireless_version/startup_check.py
+++ b/src/reachy_mini/utils/wireless_version/startup_check.py
@@ -260,10 +260,10 @@ def check_and_sync_apps_venv_sdk() -> None:
 
     """
     import json
+    import os
 
     from .update_available import get_install_source
     from .utils import build_install_command
-    import os
 
     # Get daemon install info
     try:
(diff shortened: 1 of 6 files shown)


FabienDanieau:
Shall we set UV_GIT_LFS=1 in the ISO has well?

src/reachy_mini/utils/wireless_version/startup_check.py
```diff
@@ -329,14 +329,16 @@ def check_and_sync_apps_venv_sdk() -> None:
...
         print(f"Syncing apps_venv to git ref: {daemon_info['git_ref']}")
     else:
         pkg = f"reachy-mini[gstreamer]=={daemon_info['version']}"
         extra = []
+        extra_env = []
         print(f"Syncing apps_venv to version: {daemon_info['version']}")
 
     if use_uv:
-        cmd = ["uv", "pip", "install", "--python", str(apps_venv_python), pkg] + extra
+        cmd = extra_env + ["uv", "pip", "install", "--python", str(apps_venv_python), pkg] + extra
```
FabienDanieau: don't you need a space here?

FabienDanieau:
Do we still want this? Best would be to update the ISO right (see https://github.com/pollen-robotics/reachy-mini-os/issues/56 )?

PR #801, opened 2026-02-06
fix(gst init): fixing Gst.init() call by substituing [] to None

CarolinePascal:
This PR fixes a bug introduced by the latest GStreamer version (1.28). Until now, `Gst.init(None)` was working correctly (initialization without arguments), but starting from version 1.28, this syntax will make the initialization crash. 

Using `Gst.init([])` instead fixes the issue while achieving the same result.

Tested on macOS (GStreamer@1.28) and Ubuntu (GStreamer@1.24).


alozowski:
Tested as well and it works, thanks @CarolinePascal!
Let's merge, because otherwise it's impossible to start, for example, the conversation app locally.

PR #818, opened 2026-02-11
fix(apps): cleanup orphaned app processes on daemon restart

realkim93:
## Summary

- When the daemon restarts, app subprocesses spawned by the previous `AppManager` remain running as orphans
- The new daemon's `AppManager` initializes with `current_app = None`, unaware of these orphans
- Both the orphaned app and the new daemon send commands to the robot simultaneously, causing conflicts

This PR adds `_cleanup_orphaned_apps()` to `AppManager.__init__()` that detects and terminates orphaned app processes on startup.

## Detection criteria (all must match)

- Executable path is under `_get_venv_parent_dir()`
- Matches `*_venv/bin/python` pattern (`.venv` excluded to protect the daemon itself)
- Command line contains `-m` (module execution pattern)
- PID is not the current daemon or any of its children

## Termination strategy

1. `SIGTERM` for graceful shutdown (3-second timeout)
2. `_kill_process_tree()` + `kill()` for force termination of survivors

## Changes

- `src/reachy_mini/apps/manager.py`: Added `_cleanup_orphaned_apps()` method + 1-line call in `__init__`

## Test plan

- [ ] Start daemon → start an app → restart daemon only → verify orphaned process is cleaned up
- [ ] Verify daemon's own process is not falsely terminated
- [ ] Verify no errors when no orphaned processes exist

### DIFF
--- a/src/reachy_mini/apps/manager.py
+++ b/src/reachy_mini/apps/manager.py
@@ -101,7 +101,10 @@ class AppManager:
                 if not cmd_exe.startswith(prefix) and not exe.startswith(prefix):
                     continue
 
-                # Must match *_venv/bin/python pattern (but NOT .venv — that's the daemon itself)
+                # Must match *_venv/{bin,Scripts}/python pattern
+                # (but NOT .venv — that's the daemon itself)
+                # Linux/macOS: *_venv/bin/python
+                # Windows:     *_venv/Scripts/python.exe
                 check_path = cmd_exe if cmd_exe.startswith(prefix) else exe
                 path_parts = check_path.split(os.sep)
                 is_app_venv_python = False
@@ -110,7 +113,7 @@ class AppManager:
                         part.endswith("_venv")
                         and part != ".venv"
                         and i + 1 < len(path_parts)
-                        and path_parts[i + 1] == "bin"
+                        and path_parts[i + 1] in ("bin", "Scripts")
                     ):
                         is_app_venv_python = True
                         break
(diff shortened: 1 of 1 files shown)


realkim93:
## Test Results

Tested locally on macOS (Darwin 25.2.0, Python 3.12.9).

### Test 1: Orphaned app process cleanup
- Spawned a fake orphan process (`*_venv/bin/python -m http.server`) under the venv parent directory, detached from the current process tree via `start_new_session=True`
- Instantiated `AppManager` → `_cleanup_orphaned_apps()` detected and terminated the orphan via `SIGTERM`
- **PASS** ✅

### Test 2: Daemon self-protection
- Instantiated `AppManager` and verified the current process (simulating the daemon) was NOT terminated
- `.venv` exclusion rule correctly protected the daemon's own process
- **PASS** ✅

### Test 3: No orphans — no errors
- Instantiated `AppManager` with no orphaned processes running
- Initialization completed without errors, `current_app` correctly set to `None`
- **PASS** ✅

### Note on second commit
During testing, discovered that `psutil.exe()` resolves symlinks (e.g., `venv/bin/python` → `/usr/bin/python3.12`), which caused the path-based detection to miss orphans. Fixed by also checking `cmdline[0]`, which preserves the original invocation path.

FabienDanieau:
Hey, do you still have the issue with the latest version of the daemon? If yes, I think we should track down the process that are not properly clean up in the first place. Thanks!

PR #830, opened 2026-02-12
feat(wireless update): installation improvements on the wireless version 

CarolinePascal:
This PR introduces several improvements for the `reachy_mini` update process on the wireless version :

- Introduction of an unique function `build_install_command()`to generate all the installation commands
- Improvement of the installation speed with git references with a 3 steps approach : 
1. Force reinstall `reachy_mini` without dependencies
2. Check for broken dependencies
3. Fix broken dependencies if needed
- Improvement of the robustness by calling installation commands from `HOME` instead of the current dir
- Extended support for environment variables when running a bash command with `subprocess`

### DIFF
--- a/src/reachy_mini/utils/wireless_version/startup_check.py
+++ b/src/reachy_mini/utils/wireless_version/startup_check.py
@@ -330,7 +330,7 @@ def check_and_sync_apps_venv_sdk() -> None:
 
     # Build install command
     cmd, extra_env = build_install_command(
-        extras="gstreamer",
+        extras="",
         git_ref=daemon_info.get("git_ref") if daemon_info["source"] == "git" else None,
         version=daemon_info["version"] if daemon_info["source"] != "git" else None,
         python=apps_venv_python,
--- a/src/reachy_mini/utils/wireless_version/update.py
+++ b/src/reachy_mini/utils/wireless_version/update.py
@@ -22,7 +22,7 @@ async def update_reachy_mini(
     # Update daemon venv
     logger.info("Updating daemon venv...")
     cmd, extra_env = build_install_command(
-        extras="wireless-version,gstreamer",
+        extras="wireless-version",
         git_ref=git_ref, 
         pre_release=pre_release, 
         upgrade=True,
@@ -34,7 +34,7 @@ async def update_reachy_mini(
     if apps_venv_python.exists():
         logger.info("Updating apps_venv SDK...")
         cmd, extra_env = build_install_command( 
-            extras="gstreamer",
+            extras="",
             git_ref=git_ref, 
             pre_release=pre_release,
             python=apps_venv_python, 


src/reachy_mini/utils/wireless_version/utils.py
```diff
@@ -2,21 +2,124 @@
...
+
+    # --- Package, extra args & env ---
+    if git_ref:
+        print(f"Installing from git ref: {git_ref}")
+        git_url = f"git+https://github.com/{GITHUB_REPO}.git@{git_ref}"
+        git_package = f"reachy-mini[{extras}] @ {git_url}"
+        # Step 1: force reinstall the package without the dependencies
+        step1 = shlex.join(base + [git_package, "--force-reinstall", "--no-deps", "--no-cache-dir"])
+        # Step 2: check if dependencies need to be updated
+        check_base = [arg if arg != "install" else "check" for arg in base]
```
pierre-rouanet, line 84: I think that ```uv pip check``` does not exist. Shall we force the check to rely on pip only?

src/reachy_mini/utils/wireless_version/startup_check.py
```diff
@@ -328,24 +329,17 @@ def check_and_sync_apps_venv_sdk() -> None:
...
-        git_url = f"git+https://github.com/pollen-robotics/reachy_mini.git@{daemon_info['git_ref']}"
-        pkg = f"reachy-mini @ {git_url}"
-        extra = ["--force-reinstall"]
-        print(f"Syncing apps_venv to git ref: {daemon_info['git_ref']}")
-    else:
-        pkg = f"reachy-mini=={daemon_info['version']}"
-        extra = []
-        print(f"Syncing apps_venv to version: {daemon_info['version']}")
+    cmd, extra_env = build_install_command(
+        extras="",  #TODO: replace "" with "gstreamer" when the wheel is released
```
pierre-rouanet: Should the be replaced now?

src/reachy_mini/utils/wireless_version/update.py
```diff
@@ -22,57 +19,31 @@ async def update_reachy_mini(
...
-            install_cmd = [
-                str(Path("/venvs/apps_venv/bin/pip")),
-                "install",
-                "--upgrade",
-                apps_pkg,
-            ] + extra_args
-
-        await call_logger_wrapper(install_cmd, logger)
+        cmd, extra_env = build_install_command( 
+            extras="",  #TODO: replace "" with "gstreamer" when the wheel is released
```
pierre-rouanet: same

src/reachy_mini/utils/wireless_version/utils.py
```diff
@@ -2,21 +2,124 @@
...
 
-async def call_logger_wrapper(command: list[str], logger: logging.Logger) -> None:
-    """Run a command asynchronously, streaming stdout and stderr to logger in real time.
+
+def _check_uv_available() -> bool:
+    """Check if uv is available on the system."""
+    return shutil.which("uv") is not None
+
+
+def build_install_command(
```
pierre-rouanet, line 21: It'd be better to add logger as a parameter and use log instead of prints, otherwise the info will not be propagated to the dashboard.
I can take care of it if you don't have time.

pierre-rouanet:
I managed to install your branch and then install another one. So the whole process worked but it's hard to make sure we cover all the cases here.

From what I've seen it fails to detect uv (even if it's indeed installed on my reachy_mini).

```bash
Updating daemon venv...
Running: pip install 'reachy-mini[wireless-version] @ git+https://github.com/pollen-robotics/reachy_mini.git@main' --force-reinstall --no-deps --no-cache-dir && pip check || pip install 'reachy-mini[wireless-version]' --upgrade --upgrade-strategy only-if-needed
...
```

CarolinePascal:
@pierre-rouanet I added an `uv` check outside vents as well ;) I also switched to `logging` instead of `print` (which should be avoided in production code 🙈)

pierre-rouanet:
Great, thanks!

src/reachy_mini/utils/wireless_version/utils.py
CarolinePascal: I'm pretty sure it does 🤔 Random reference for proof : https://github.com/astral-sh/uv/issues/10398

src/reachy_mini/utils/wireless_version/startup_check.py
CarolinePascal: It is actually ! I did not get the news.

src/reachy_mini/utils/wireless_version/utils.py
CarolinePascal: Done in [e5c14a0](https://github.com/pollen-robotics/reachy_mini/pull/830/commits/e5c14a06aae575964aa1e93c2e88b7f0cb4c7620)

src/reachy_mini/utils/wireless_version/update.py
CarolinePascal: Done in [64fb8d9](https://github.com/pollen-robotics/reachy_mini/pull/830/commits/64fb8d94441b4367c642a52af68da9b44f51d347)

src/reachy_mini/utils/wireless_version/startup_check.py
CarolinePascal: Done in [64fb8d9](https://github.com/pollen-robotics/reachy_mini/pull/830/commits/64fb8d94441b4367c642a52af68da9b44f51d347)

src/reachy_mini/utils/wireless_version/utils.py
CarolinePascal: Though I did not pass `logger` as a parameter but used a global variable.

PR #841, opened 2026-02-15
feat(volume control): Adding support for volume control on all OS

CarolinePascal:
This PR adds support for volume control on Linux, macOS and Windows. The different implementations rely on :
- **macOS**: CoreAudio framework.
- **Linux**: `amixer`, `aplay` and `arecord`
- **Windows**: `pycaw` library

Of course tested across all OSes.

Testing procedure :
- `pytest -vv tests/unit_tests/test_volume_control.py`
- Run the daemon, open the web dashboard, change volume. The changes should work even if Reachy Mini is not selected as default.


RemiFabre:
So I pip installed this branch from a clean env with `pip install -e .[dev]`

I'm testing on a lite but no sound is coming out of the robot at all, even if I put the sound in the dashboard at maximum. Same behavior regardless of the default output source in ubuntu (I restarted the daemon after changing the source).

The pytest command crashes on weird import problems, I'm investigating if it could come from my install

RemiFabre:
Ah sorry I'm dumb, I thought these were integration tests so I ran them with the daemon alive, tests are passing;

But still no sound coming out from the robot when I run the daemon. Here the wrong output is selected:
image

RemiFabre:
I confirm that with the exact same setup, sound works on develop

CarolinePascal:
~~Looks like the volume control defaults to the system default audio device 🤔 Could you please copy-paste the output of `aplay -l` and `pactl list` here ? Thanks !~~

There was a bug indeed (I expected Reachy Mini sound card names to match exactly, but they sometimes don't), it should be fixed now !

N.B. The current implementation looks for Reachy Mini Audio and respeaker, so there is a chance it will pick the respeaker first if one respeaker card is plugged !

RemiFabre:
Ok retesting, first in simulation (first image) I can't play any sound although a viable speaker is selected this time.
image

And then on a Lite:
image

Still no sound.

But this is also currently the case on develop, so maybe this isn't this branch's fault?

alozowski:
It works for me!

PR #843, opened 2026-02-16
Better mypy workflow

FabienDanieau:
I tried to clarify the mypy workflow. We often had differences between local and CI. This is due to a difference of venv. Mypy gives different results depending on the found dependencies. I've changed the CI a bit to use all deps (the more the better).

 I explained that in the contributing.md Is is clear enough?
 
@alozowski do you think we can extend this contributing with more info?
@RemiFabre what do you think about telling the agent that this file exists? Having them using ruff and mypy will help to generate better code

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,6 +12,14 @@ on:
       - "tests/**"
       - ".github/workflows/**"
       - "pyproject.toml"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   ruff:
     runs-on: ubuntu-latest
@@ -26,17 +34,14 @@ jobs:
     steps:
       - name: Install system dependencies
         run: sudo apt-get update && sudo apt-get install -y libportaudio2 libcairo2-dev libgirepository1.0-dev
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
       - name: Install uv
         uses: astral-sh/setup-uv@v7
-
-      - name: Set up Python
-        uses: actions/setup-python@v5
         with:
-          python-version: "3.12"
+          python-version: 3.12
 
       - name: Install the project
-        run: uv sync --locked --all-extras
+        run: uv sync --frozen --all-extras
 
       - name: Lint with mypy
         run: uv run mypy
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -2,12 +2,12 @@
 
 ## Code quality
 
-the quality of code is insured by ruff and mypy. Please make sure to run them before pushing your code.
+The quality of code is insured by ruff and mypy. Please make sure to run them before pushing your code.
 
 All the tools are available in the [dev] dependency group, so you can install them with:
 
 ```bash
-pip install -e ".[dev]"
+uv sync --group dev
 ```
 
 ### Ruff
@@ -32,7 +32,7 @@ To check the type annotations of the code, you can run mypy with the following c
 mypy
 ```
 
-Please note that mpy results depends on the installed package. The [CI](../.github/workflows/lint.yml) runs mypy with the full installation with uv on linux. Any differences between the CI and the local environment is probably due to missing dependencies. To ensure you have the same environment as the CI, you can install the package with all the extras with uv:
+Please note that mpy results depend on the installed package. The [CI](../.github/workflows/lint.yml) runs mypy with the full installation with uv on linux. Any differences between the CI and the local environment are probably due to missing dependencies. To ensure you have the same environment as the CI, you can install the package with all the extras with uv:
 
 ```bash
 uv sync --all-extras
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,7 @@ dev = [
     "mypy==1.18.2",
     "types-requests",
     "types-toml",
-    "types-PyYAML"
+    "types-PyYAML",
 ]
 examples = ["pynput", "soundfile"]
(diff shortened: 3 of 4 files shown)


.github/workflows/lint.yml
```diff
@@ -24,16 +24,19 @@ jobs:
   mypy:
     runs-on: ubuntu-latest
     steps:
+      - name: Install system dependencies
+        run: sudo apt-get update && sudo apt-get install -y libportaudio2 libcairo2-dev libgirepository1.0-dev
       - uses: actions/checkout@v4
```
alozowski: I found out recently that it's possible to use `actions/checkout@v6`

.github/workflows/lint.yml
```diff
@@ -24,16 +24,19 @@ jobs:
...
         with:
           python-version: "3.12"
-          cache: "pip" # caching pip dependencies
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install .[dev]
+
+      - name: Install the project
+        run: uv sync --locked --all-extras
```
alozowski: Do you have a strong opinion on `--locked` here? For example, I use `--frozen` in [the conv app typecheck.yml](https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/f933d9f89478e43ebb50a1f9a2337f2f8bcf2a9a/.github/workflows/typecheck.yml#L26). I tend to avoid `--locked` in CI only because it verifies the lockfile is up-to-date with `pyproject.toml`, then installs from it. If they diverge, it errors out, when `--frozen` installs exactly what's in `uv.lock` without any validation against `pyproject.toml` and it's better for reproducibility

.github/workflows/lint.yml
```diff
@@ -24,16 +24,19 @@ jobs:
...
     steps:
+      - name: Install system dependencies
+        run: sudo apt-get update && sudo apt-get install -y libportaudio2 libcairo2-dev libgirepository1.0-dev
       - uses: actions/checkout@v4
-      - name: Set up Python 3.12
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+
+      - name: Set up Python
         uses: actions/setup-python@v5
```
alozowski: I found out that `setup-uv` can manage python too, so you can try to test it with:
```
  uses: astral-sh/setup-uv@v7
  with:
    python-version: 3.12
```
and if it works, it's possible to drop `setup-python`. I'll try the same for the conv app

.github/workflows/lint.yml
```diff
@@ -24,16 +24,19 @@ jobs:
   mypy:
     runs-on: ubuntu-latest
     steps:
+      - name: Install system dependencies
+        run: sudo apt-get update && sudo apt-get install -y libportaudio2 libcairo2-dev libgirepository1.0-dev
```
alozowski, line 36: it doesn't really matter, but should we checkout first and then install apt packages? `apt-get` install doesn't need the repo to be checked out, so both orderings work, but checking out first looks more logical (feel free to disagree)

.github/workflows/lint.yml
alozowski, line 14: It's possible to add here these sections:
```
permissions:
  contents: read

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

`concurrency` section is mostly a quality-of-life improvement, because without it, if you push, for example, three commits in a row to the same branch, github will spin up three separate lint runs, even though only the last one matters. This setting groups runs by workflow name + branch, and automatically cancels the older ones when a new push comes in

docs/contributing.md
```diff
@@ -0,0 +1,54 @@
+# Contributing
+
+## Code quality
+
+the quality of code is insured by ruff and mypy. Please make sure to run them before pushing your code.
```
alozowski: It's better to start with the capital "T" and fixed "ensured" typo:
```
The quality of code is ensured by ruff and mypy.
```

docs/contributing.md
```diff
@@ -0,0 +1,54 @@
...
+
+All the tools are available in the [dev] dependency group, so you can install them with:
+
+```bash
+pip install -e ".[dev]"
```
alozowski: Since CI now uses uv, worth showing `uv sync --group dev` as the primary method here (and pip as fallback)

docs/contributing.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Mypy
+
+To check the type annotations of the code, you can run mypy with the following command:
+
+```bash
+mypy
+```
+
+Please note that mpy results depends on the installed package. The [CI](../.github/workflows/lint.yml) runs mypy with the full installation with uv on linux. Any differences between the CI and the local environment is probably due to missing dependencies. To ensure you have the same environment as the CI, you can install the package with all the extras with uv:
```
alozowski: Small typos:
```
Please note that mypy results depend on the installed packages.
```

alozowski: Also "Any differences between the CI and the local environment are probably"... (changing is to are)

pyproject.toml
```diff
@@ -50,6 +50,7 @@ dev = [
     "mypy==1.18.2",
     "types-requests",
     "types-toml",
+    "types-PyYAML"
```
alozowski: let's add a trailing comma, they always make the code nice

.github/workflows/lint.yml
FabienDanieau: oh yeah that's what I want

.github/workflows/lint.yml
FabienDanieau: yes these packages are needed by some pip dependencies

alozowski:
About the contributing guide – I can extend it after this PR is merged as a separate PR

Issue #846, opened 2026-02-17
Async call will fail when call from the webrtc handler.


pierre-rouanet:
```bash
2026-02-17 14:52:15,766 - reachy_mini.media.webrtc_daemon - INFO - Data channel message from peer 8cbc2082-27b4-402f-8b46-9e13a66e1003: {"goto_target":{"head":[[1,0,0,0],[0,1,0,0],[0,0,1,0],[0,0,0,1]]}}
2026-02-17 14:52:15,766 - reachy_mini.daemon.backend.robot.backend - ERROR - WebRTC command error: no running event loop
/Users/pierrerouanet/Dev/mini/reachy_mini/src/reachy_mini/daemon/backend/abstract.py:872: RuntimeWarning: coroutine 'Backend._webrtc_goto' was never awaited
```

PR #847, opened 2026-02-17
Run webrtc async call in threaded loop.

pierre-rouanet:
As mentioned in #846, the async call will fail from the webrtc handler.

```2026-02-17 14:52:15,766 - reachy_mini.media.webrtc_daemon - INFO - Data channel message from peer 8cbc2082-27b4-402f-8b46-9e13a66e1003: {"goto_target":{"head":[[1,0,0,0],[0,1,0,0],[0,0,1,0],[0,0,0,1]]}}
2026-02-17 14:52:15,766 - reachy_mini.daemon.backend.robot.backend - ERROR - WebRTC command error: no running event loop
/Users/pierrerouanet/Dev/mini/reachy_mini/src/reachy_mini/daemon/backend/abstract.py:872: RuntimeWarning: coroutine 'Backend._webrtc_goto' was never awaited
```

I've added a run loop to make sure they are always scheduled.
Don't hesitate to change it if you have a better/cleaner fix.

### DIFF
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -850,7 +850,13 @@ class Backend:
             gst_webrtc (Any): The GstWebRTC instance to setup.
 
         """
-        gst_webrtc.set_message_handler(self._handle_webrtc_message)
+        _loop = asyncio.new_event_loop()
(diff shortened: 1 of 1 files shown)


cdussieux:
looks good

PR #849, opened 2026-02-17
Fix Windows UnicodeEncodeError when creating app templates with emoji

LaureenFabre:
On Windows, when I tried to follow the documentation to create and publish an app (https://huggingface.co/blog/pollen-robotics/make-and-publish-your-reachy-mini-apps), I hit a Unicode error: 
image

The template files seem to be written with _**write_text**_ without an explicit encoding. My fix is to set **_encoding="utf-8"_** everywhere **_write_text_** is used.

I only tested this fix on Windows, but it should not introduce regressions on Linux or macOS.


RemiFabre:
Tested on Linux, works

RemiFabre:
Thanks :)

PR #851, opened 2026-02-17
848 microphone fpc cable change tutorial

augustin-crampette:
Add a tutorial to help people changing the mic's FPC cable.
Also add a note about the microphone issue in the "frequent issues" section.


FabienDanieau (approved):
All good! I finally moved only motor_diagnosis and change_mic_fps_cable to the new folder, so external links to troubleshooting.md are preserved.
They will show up in the same Help & Support section in the doc anyway. You can merge whenever you want!

PR #852, opened 2026-02-17
fix: support PulseAudio device detection on Linux (non-PipeWire)

andimarafioti:
## Problem

When running Reachy Mini on a Jetson AGX Orin (which uses PulseAudio, not PipeWire), the GStreamer `DeviceMonitor` device lookup fails silently. The code checks for the PipeWire-specific `node.name` property, which doesn't exist under PulseAudio, causing it to fall through all conditional branches and return `None`.

This results in:
```
WARNING - No Source audio card found.
WARNING - No Sink audio card found.
```

...and no audio playback despite the USB audio device being correctly detected by ALSA and PulseAudio.

## Fix

Added a Linux fallback path that constructs the PulseAudio device name from `udev.id` + `device.profile.name` (e.g. `alsa_output.usb-Pollen_Robotics_Reachy_Mini_Audio_...-00.analog-stereo`), with an additional `device.string` ALSA fallback.

Applied consistently to both `audio_gstreamer.py` and `webrtc_daemon.py`.

## Impact

- **No change** to PipeWire, Windows, or macOS code paths (checked first)
- Fixes audio on any Linux system running PulseAudio instead of PipeWire
- Tested on Jetson AGX Orin (JetPack 6, Ubuntu 22.04, PulseAudio)

### DIFF
--- a/src/reachy_mini/media/webrtc_daemon.py
+++ b/src/reachy_mini/media/webrtc_daemon.py
@@ -322,56 +322,32 @@ class GstWebRTC:
         monitor.add_filter("Audio/Source")
         monitor.start()
 
-        target = "Reachy Mini Audio"
-        try:
-            for device in monitor.get_devices():
-                name = device.get_display_name() or ""
-                props = device.get_properties()
+        snd_card_name = "Reachy Mini Audio"
 
-                if target not in name:
-                    continue
+        devices = monitor.get_devices()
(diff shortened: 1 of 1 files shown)


FabienDanieau:
Looks good! Maybe @OriNachum wants to test on his setup, he had also issues with gstreamer and orin

PR #853, opened 2026-02-17
Fix Windows temp directory cleanup failure in reachy-mini-app-assistant check

LaureenFabre:
On Windows, when I tried to follow the documentation to create and publish an app (https://huggingface.co/blog/pollen-robotics/make-and-publish-your-reachy-mini-apps), I hit a _**PermissionError.**_

In the **_check()_** function, we switch the current working directory to a temporary one (while creating the test environment), but after the checks complete we do not switch back to the original directory. 
On Windows, this keeps the temporary directory “in use” and prevents cleanup.

My fix is to save the original working directory before switching to the temporary one, then restore it after the checks.

Tested on Windows only, but this should not impact other operating systems.


RemiFabre:
The changes are minimal but appear big because a try/finally block was added and the indentation changes many lines.

Tested on Linux, works. Waiting for a test on Mac before merging.

Thank you for this!

CarolinePascal:
Tested on macOS as well - No errors, and the temporary folder is correctly deleted at the end. I'm merging the PR.

PR #860, opened 2026-02-18
859 add gstreamer to pyprojectyml

FabienDanieau:
Gstreamer provided a temp bundle name so we can add gstreamer deps to the pyproject.toml
Could you check if everything is still ok with Mac?

Thanks

### DIFF
--- a/docs/source/API/media.mdx
+++ b/docs/source/API/media.mdx
@@ -11,7 +11,7 @@
 
 [[autodoc]] reachy_mini.media.audio_gstreamer.GStreamerAudio
 
-<!-- [[autodoc]] reachy_mini.media.audio_sounddevice.SoundDeviceAudio -->
+[[autodoc]] reachy_mini.media.audio_sounddevice.SoundDeviceAudio
 
 ### Audio Utils Functions
 


alozowski:
I tested this on macos inside the conversation app and `uv sync --all-extras --group dev` fails during dependency resolution.
uv attempts to resolve for `sys_platform == 'win32'` and reports:
```
❯ uv sync --all-extras --group dev --reinstall
  × No solution found when resolving dependencies for split (markers: python_full_version == '3.12.*' and sys_platform == 'win32'):
  ╰─▶ Because gstreamer-cli==1.28.0 requires Python >=3.13 and only gstreamer-cli<=1.28.0 is available, we can conclude that gstreamer-cli>=1.28.0
      cannot be used.
      And because gstreamer-bundle==1.28.0 depends on gstreamer-cli>=1.28.0, we can conclude that gstreamer-bundle==1.28.0 cannot be used.
      And because reachy-mini==1.3.1 depends on gstreamer-bundle{sys_platform != 'linux'}==1.28.0 and only reachy-mini==1.3.1 is available, we can
      conclude that all versions of reachy-mini cannot be used.
      And because your project depends on reachy-mini and your project requires reachy-mini-conversation-app[all-vision], we can conclude that your
      project's requirements are unsatisfiable.

      hint: The resolution failed for an environment that is not the current one, consider limiting the environments with `tool.uv.environments`.

      hint: `gstreamer-cli` was requested with a pre-release marker (e.g., gstreamer-cli>1.28.0,<1.29.dev0), but pre-releases weren't enabled (try:
      `--prerelease=allow`)
```

The failure occurs while resolving for `sys_platform == 'win32'`, even though the environment is macos.

I tried adding metadata into the conversation app pyproject:
```
[[tool.uv.dependency-metadata]]
name = "gstreamer-cli"
requires-python = ">=3.9"
```

(and similar overrides for the other gstreamer-* packages). While this can sometimes make resolution succeed, it doesn't reliably fix the problem and can result in inconsistent or broken environments depending on platform markers and extras.

From my side, this looks like a metadata / platform constraint issue in the gstreamer packaging rather than a uv configuration problem.

alozowski:
I also tested with Python 3.13 to satisfy the `gstreamer-cli>=3.13` constraint. The resolution issue changes, but the overall installation is still not stable due to platform-specific constraints and gstreamer packaging interactions. So simply bumping Python to 3.13 does not resolve the problem in a clean way.

nirbheek:
This issue was fixed with https://gitlab.freedesktop.org/gstreamer/cerbero/-/merge_requests/2116, and will be in the 1.28.1 wheels release. Probably next week or so.

nirbheek:
> installation is still not stable due to platform-specific constraints

Forgot to mention here, but we released 1.28.1 last week, so this bug should be fixed now.

pyproject.toml
```diff
@@ -37,11 +37,44 @@ dependencies = [
...
+requires-python = ">=3.9"
+
+[[tool.uv.dependency-metadata]]
+name = "gstreamer-plugins-restricted"
+requires-python = ">=3.9"
+
+[[tool.uv.dependency-metadata]]
+name = "gstreamer-plugins-gpl-restricted"
+requires-python = ">=3.9"
+
```
nirbheek: All this can go away now.

alozowski:
Hi @nirbheek! I just tried to substitute `gstreamer-bundle==1.28.0` to `gstreamer-bundle==1.28.1` and remove the whole patch section, but I still can't install because it seems 1.28.1 isn't published – could you please check?

nirbheek:
Haaa only that wheel got forgotten :sweat_smile:  fixed now, sorry!

FabienDanieau:
> Haaa only that wheel got forgotten 😅 fixed now, sorry!

All good but we had to do this
```
[tool.uv]
dependency-metadata = [
    # Upstream metadata currently marks `gstreamer-msvc-runtime` as unconditional.
    # It should only be required on Windows.
    { name = "gstreamer-libs", version = "1.28.1", requires-dist = ["gstreamer-msvc-runtime; sys_platform == 'win32'", "setuptools"] },
]
```
for some reason` uv pip install gstreamer-bundle `tries to download gstreamer-msvc-runtime

nirbheek:
> for some reason `uv pip install gstreamer-bundle` tries to download gstreamer-msvc-runtime

Hm, so uv also requires dependencies to be invariant across all wheels. We would need to ship a dummy gstreamer-msvc-runtime on other platforms or something like that. Thanks.

PR #861, opened 2026-02-18
Clean   simplify  webrtc audio management

cdussieux:


src/reachy_mini/media/webrtc_client_gstreamer.py
```diff
@@ -137,22 +137,19 @@ def __init__(
         webrtcsrc = self._configure_webrtcsrc(signaling_host, signaling_port, peer_id)
         self._pipeline_record.add(webrtcsrc)
 
+        self._webrtcbin = None
+        self._webrtcsrc = webrtcsrc
```
FabienDanieau: Can we use self._webrtcsrc directly in l137? I'm not sure the local variable is useful here

src/reachy_mini/media/webrtc_client_gstreamer.py
```diff
@@ -354,42 +375,92 @@ def stop_recording(self) -> None:
...
+        appsrc.set_property("is-live", True)
         caps = Gst.Caps.from_string(
             f"audio/x-raw,format=F32LE,channels={self.CHANNELS},rate={self.SAMPLE_RATE},layout=interleaved"
         )
-        self._appsrc.set_property("caps", caps)
+        appsrc.set_property("caps", caps)
 
         audioconvert = Gst.ElementFactory.make("audioconvert")
         audioresample = Gst.ElementFactory.make("audioresample")
+        opusenc = Gst.ElementFactory.make("opusenc")
```
FabienDanieau, line 428: minor optim, you can specify the `restricted-lowdelay` profile
https://github.com/pollen-robotics/gstreamer_webrtc/blob/9bda95d8e22cbfabcdc7d551313c5440edccd228/src/gstreamer/avpipeline.py#L248

src/reachy_mini/media/webrtc_client_gstreamer.py
```diff
@@ -354,42 +375,92 @@ def stop_recording(self) -> None:
...
-            )
+        elems = (appsrc, audioconvert, audioresample, opusenc, rtpopuspay)
 
-        meta_structure = Gst.Structure.new_empty("meta")
-        meta_structure.set_value("name", "reachymini-client")
-        webrtcsink.set_property("meta", meta_structure)
+        # Add elements to webrtcbin's parent bin so they share the same hierarchy
+        target_bin = webrtcbin_parent if webrtcbin_parent else self._webrtcsrc
+        for elem in elems:
+            if not target_bin.add(elem):
```
FabienDanieau, line 439: Shouldn't we remove the elements if the client leaves?

FabienDanieau (approved):
Way more elegant than my solution, love it!
I've added optional optimization suggestions

PR #865, opened 2026-02-18
docs(uv active flag): adding` uv --active` flag on windows

CarolinePascal:
This PR adds in the docs the necessary `--active` flag on `uv` commands running on Windows.

### DIFF
--- a/docs/source/SDK/installation.md
+++ b/docs/source/SDK/installation.md
@@ -237,7 +237,8 @@ uv pip install --upgrade --index-url https://gitlab.freedesktop.org/api/v4/proje
 
 In your terminal, run:
 ```bash
-git clone https://github.com/pollen-robotics/reachy_mini && cd reachy_mini
+git clone https://github.com/pollen-robotics/reachy_mini
+cd reachy_mini
 uv pip install -e .
 uv pip install --upgrade --index-url https://gitlab.freedesktop.org/api/v4/projects/1340/packages/pypi/simple gstreamer==1.28.0
 ```


RemiFabre:
Thx! We tested something similar yesterday with mixed results but we'll retry this evening using exactly your commands.

LaureenFabre:
Hi!

Thanks for editing the documentation, it's all good on that part.
I still had one issue when following this documentation: https://huggingface.co/blog/pollen-robotics/make-and-publish-your-reachy-mini-apps

I had to run **_uv pip install -e ._** before running **_uv run --active reachy-mini-app-assistant create_**
otherwise, the program is not found

RemiFabre:
Ok so we still had some issues. The problem was that we followed "Option B: Install from Source", but when we run:
`uv sync --active` it doesn't install from the current local branch, instead the uv.lock forces v1.3.1 for reachy_mini, and the recent windows fixes are not there yet.

Doing `uv pip install -e .` after the uv sync forces a local (and editable) install. Imo this would be a better default behavior when installing from source. What do you think @CarolinePascal ?

After this, we were able to do everything (simulation, app creation, app run from dashboard, app check, app publish) on the windows machine!

RemiFabre:
Also all the commands on this guide:
https://huggingface.co/blog/pollen-robotics/make-and-publish-your-reachy-mini-apps

Require "uv run --active" to work with this setup. Two things:
- uv is annoying no?
- Wanna update the blog post to add this and become first author? :)

LaureenFabre:
Hi,
In the end, what works for me is mainly never running uv sync. So I have to force --no-sync.

Here is a procedure on Windows that works on my side:

```
git clone <repository>

powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"

uv --version

powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"

git lfs install

uv venv reachy_mini_env --python 3.12

powershell Set-ExecutionPolicy RemoteSigned

reachy_mini_env\Scripts\activate

cd reachy_mini

uv pip install -e ".[mujoco]"

uv pip install --upgrade --index-url https://gitlab.freedesktop.org/api/v4/projects/1340/packages/pypi/simple gstreamer==1.28.0

uv run --active --no-sync reachy-mini-daemon --sim
```
Then for the app part:
```
uv run --active --no-sync reachy-mini-app-assistant create

uv run --active --no-sync reachy-mini-app-assistant check

uv pip install -e .

uv run --active --no-sync reachy-mini-daemon --sim

uv run --active --no-sync reachy-mini-app-assistant publish
```

There is also a Windows specific issue with reachy-mini-app-assistant check and reachy-mini-app-assistant publish, but that should be handled in another PR. I have a fix that works on Windows for this, but I’m not sure whether it has any impact on Linux or macOS.

CarolinePascal:
Hi !

Thanks for the combined efforts on this issue :) I was also considering removing `uv sync` as it has inconsistent behaviors across OS (e.g. `--active` seems necessary on macOS as it silently creates a local `.venv`otherwise...). 

Then for the commands, I'm wondering if we need `uv run` at all... I remember starting the daemon with the executable `reachy_mini_daemon.exe` and no `uv run` on Windows. In the meantime, do you know why we need the `--no sync` flag with `uv run` ? I don't really get it 🤔 

Thanks again for the testing efforts !

Best,

Caroline.

CarolinePascal:
A new potential issue with `uv run` - https://github.com/astral-sh/uv/issues/14585 😬

LaureenFabre:
Hi,
It also seems to me that during my installation battles, I managed to run the commands without uv run. But since I wasn’t familiar with uv and the documentation clearly said to always use it, I kept looking for solutions that included it.

For --no-sync, I need to use it because otherwise it automatically performs a --sync, which causes errors: it tries to reinstall things with sync, and that leads to file access errors.

Have a good day

CarolinePascal:
Ok, looks like `uv` tries to sync whatever happens 🤔 

I double checked, and sourcing the virtual env indeed updates the `PATH` meaning that `uv run` is not needed 🎉  I'll update the docs trying to only keep the simplest commands (i.e. with a minimum of `--active`, `--no-sync` flags to minimize errors). I'm keeping you updated !

CarolinePascal:
I modified the commands as follows:
- `uv sync` ➡️ `uv pip install -e .` - Ensures a local and editable install of the package in the currently active virtual environment.
- `uv run my_exec` ➡️ `my_exec` - Ensures the currently active virtual environment is used, without messing up with the current installation.

I'm waiting for your green flag to merge 👍

P.S. Both changes were tested on macOS + Windows.

LaureenFabre:
The whole procedure works fine on Windows, thanks for the changes :-)

The only thing that doesn’t work is the **_&&_** on Windows, at least with my PowerShell. It has to be done in two separate steps:

```
git clone https://github.com/pollen-robotics/reachy_mini
cd reachy_mini
```
image


Otherwise, just a small and not very important remark: you include the GStreamer installation command in option A, in option B, and then again in the specific GStreamer section. It’s not a problem, but I’m thinking that people who aren’t used to this might end up running the command twice.

Thanks again for the quick response on this topic :-)

CarolinePascal:
Once again a Windows oddity - `&&` works in the regular command line tool but not in PowerShell (at least in older versions of Powershell) ! 

I'll update the documentation to match this behavior. 

I'll also add a work about the redundant GStreamer installation, but better twice than not at all 👍

CarolinePascal:
Update : I decided not to touch the second GStreamer mention, as it is in the Linux section and should not impact Windows and macOS users.

RemiFabre:
Good, glad we're improving all of this. Merging!

PR #872, opened 2026-02-19
Add mDNS service registration and discovery for Reachy Mini robots

pierre-rouanet:
Register a _reachy-mini._tcp DNS-SD service when the daemon starts so clients can discover robots on the local network automatically via find_robots(). Includes a TCP liveness check to filter stale cache entries and a CLI entry point (python -m reachy_mini.utils.discovery).

Add discovery abilities to make it easy to work with multiple robots.
Not sure if we should directly include it as it could have some side effects (zombies ReachyMini, add delay before connection). What do you think @cdussieux?

- [x] Needed some workaround on macOS but works on my side
- [x] Works on windows as well
- [x] Test on linux

### DIFF
--- a/src/reachy_mini/utils/discovery.py
+++ b/src/reachy_mini/utils/discovery.py
@@ -43,7 +43,8 @@ def _get_local_ip() -> str:
         s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
         try:
             s.connect(("224.0.0.1", 0))
-            return s.getsockname()[0]
+            ip: str = s.getsockname()[0]
+            return ip
         finally:
             s.close()
     except OSError:


cdussieux:
Linux lite, discovery went well even though reported IP is weird (docker bridge IP ?)

wireless worked both locally and from linux

could detect both while running at the same time

(only tested the discovery script)

src/reachy_mini/daemon/app/main.py
```diff
@@ -157,6 +160,9 @@ async def dataset_updater(interval_hours: float) -> None:
                     hardware_config_filepath=args.hardware_config_filepath,
                 )
 
+            # Register mDNS service only after the daemon is ready
+            mdns.register()
```
alozowski, line 164: `mdns.register()` is called unconditionally, even when `autostart=False` – in that case, the service may be advertised while the daemon hasn't been started yet. The comment says "register only after the daemon is ready", so should we move registration inside the `if args.autostart` block, or otherwise gate it on daemon state? something like:
```
mdns_registered = False

try:
    if args.autostart:
        await app.state.daemon.start(...)
        mdns.register()
        mdns_registered = True

    yield
finally:
    if mdns_registered:
        mdns.unregister()
    ...
```

pierre-rouanet: You're right! Yet as the no autostart option will be deprecated soon, I'll just leave it this way to not make the code more complex.

PR #874, opened 2026-02-20
Issues on the reachy-mini-app-assistant check/publish

LaureenFabre:
Hi,

I had two issues on the reachy-mini-app-assistant check/publish :
- Fix app check isolation in temporary venv :
Improve isolation when checking an app inside a temporary virtual environment.
It replaces direct **_pip_** calls with **_python -m pip_** from the target venv and runs subprocesses with a cleaned environment, notably removing PYTHONPATH and explicitly setting VIRTUAL_ENV.
This prevents the installation or entrypoint check from accidentally using modules from the parent environment, which could lead to inconsistent behavior or false positives.

- Make app publish git flow robust on Windows :
Replace shell-based Git commands (os.system, shell=True, and cd && ...) with structured **_subprocess.run_** calls using explicit argument lists and _**cwd**_, along with proper error handling.
It also makes the space remote configuration idempotent by updating it if it already exists instead of failing.
This should improve reliability, cross-platform consistency, and error visibility, especially in cases where shell behavior, quoting, or partial previous runs caused fragile or silent failures.

Only tested on Windows.

Best

### DIFF
--- a/.github/workflows/upload_pr_documentation.yml
+++ b/.github/workflows/upload_pr_documentation.yml
@@ -5,6 +5,7 @@ on:
     workflows: ["Build HF PR Documentation"]
     branches:
       - main
+      - develop
     types:
       - completed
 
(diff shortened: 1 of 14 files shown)


RemiFabre (approved):
Hello,
I think the changes are good and safe. This has been tested successfully on Windows and Ubuntu but not yet on Mac. Test procedure is e.g.:
```
reachy-mini-app-assistant create fakeapp4 .
reachy-mini-app-assistant check fakeapp4/
reachy-mini-app-assistant publish fakeapp4/
```

Waiting for a test on Mac to be extra safe, then merging. Thank you for your work.

RemiFabre:
Thanks @CarolinePascal for the Mac test, merging!

Issue #878, opened 2026-02-21
mujoco sim crashes on macos: UDP frame too large


alozowski:
I got the following error starting the robot in simulation:
```
❯ mjpython -m reachy_mini.daemon.app.main --sim
<...>
2026-02-21 19:35:25,098 - reachy_mini.daemon.daemon - INFO - Waking up Reachy Mini...
Exception in thread Thread-8 (rendering_loop):
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.12/3.12.12_2/Frameworks/Python.framework/Versions/3.12/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "/opt/homebrew/Cellar/python@3.12/3.12.12_2/Frameworks/Python.framework/Versions/3.12/lib/python3.12/threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/lozowski/Developer/reachy/reachy_mini/src/reachy_mini/daemon/backend/mujoco/backend.py", line 146, in rendering_loop
    streamer.send_frame(im)
  File "/Users/lozowski/Developer/reachy/reachy_mini/src/reachy_mini/daemon/backend/mujoco/video_udp.py", line 47, in send_frame
    self.sock.sendto(data, self.addr)
OSError: [Errno 40] Message too long
```
macos has a default UDP datagram size limit of 9216 bytes:
```
❯ sysctl net.inet.udp.maxdgram
net.inet.udp.maxdgram: 9216
```

The jpeg-encoded frames from the 1280×720 camera at quality 80 easily exceed this, causing `sendto()` to fail. Linux and Windows have higher defaults (65535), so this affects macos only.

The `max_packet_size` parameter exists in `UDPJPEGFrameSender` but is never actually used to constrain the sent data.

PR #857, opened 2026-02-18
use gstreamer to stream the mujoco camera

FabienDanieau:
make it easier to remove opencv dep in the near future

This should work as usual

```
reachy-mini-daemon --sim
```

```
GST_DEBUG=3 python examples/look_at_image.py 
```

PR #879, opened 2026-02-21
Fix mujoco sim crash on macos

alozowski:
Fixes issue https://github.com/pollen-robotics/reachy_mini/issues/878.

macos limits udp datagrams to 9216 bytes by default, causing `OSError: [Errno 40] Message too long` when streaming camera frames from the mujoco sim.

Changes:
- `backend.py:` detect platform and use 640×360 on macos, 1280×720 on linux/windows
- `video_udp.py`: default `max_packet_size` to 8192, auto-reduce jpeg quality to fit

Let's test if it works and makes sense


alozowski:
Solved by https://github.com/pollen-robotics/reachy_mini/pull/857.

PR #880, opened 2026-02-21
Use dedicated GMainContext per GStreamer component to prevent GIL deadlock

fuzzie360:
# Summary

- Each GStreamer component (GStreamerAudio, GStreamerCamera, GstWebRTCClient, GstWebRTC) now creates its own dedicated GLib.MainContext instead of using the shared default context
- Bus watches and timeout sources are attached to the per-component context, preventing GIL deadlocks that occur when multiple GStreamer components compete for the global default GMainContext from different Python threads
- GLib.timeout_add_seconds() calls replaced with explicit GLib.timeout_source_new_seconds() + source.attach(context) to target the correct context

# Problem

When multiple GStreamer components (camera, audio, WebRTC) each run their own GLib.MainLoop on separate threads, they all implicitly share the process-wide default GMainContext. GLib's context acquisition interacts poorly with Python's GIL, causing deadlocks under concurrent usage.

As a result, a deadlock can occur randomly, which causes:

- uvicorn api server does not respond (waiting forever)
- no zenoh topics are published so nothing can be read
- no errors are logged so very hard to debug

# Solution

https://developer.gnome.org/documentation/tutorials/main-contexts.html#default-contexts describes how thread-default contexts should be used instead for these I/O operation tasks.

Give each component its own GMainContext by:
1. Creating a new context via GLib.MainContext.new()
2. Constructing the MainLoop bound to that context via GLib.MainLoop.new(context, False)
3. Pushing the context as thread-default (push_thread_default / pop_thread_default) when registering bus watches and when running the loop
4. Attaching timeout sources explicitly to the component's context instead of using the global GLib.timeout_add_seconds()

### DIFF
--- a/src/reachy_mini/media/webrtc_daemon.py
+++ b/src/reachy_mini/media/webrtc_daemon.py
@@ -281,9 +281,13 @@ class GstWebRTC:
         appsrc.set_property("caps", caps)
 
         play_bus = playback_pipe.get_bus()
-        play_bus.add_watch(
-            GLib.PRIORITY_DEFAULT, self._on_playback_bus_message, peer_id
-        )
+        self._context.push_thread_default()
+        try:
+            play_bus.add_watch(
+                GLib.PRIORITY_DEFAULT, self._on_playback_bus_message, peer_id
+            )
+        finally:
+            self._context.pop_thread_default()
 
         playback_pipe.set_state(Gst.State.PLAYING)
 


fuzzie360:
Sorry, my mistake. This was a misdiagnosis.

The symptoms were the same so I thought it was a deadlock like my other issue.

I found the actual error log here:

```
Feb 21 23:53:19 reachy-mini launcher.sh[153892]: ERROR:reachy_mini.daemon.daemon:Backend encountered an error: unpack requires a buffer of 2 bytes
Feb 21 23:53:19 reachy-mini launcher.sh[153892]: 2026-02-21 23:53:19,938 - reachy_mini.daemon.daemon - ERROR - Backend encountered an error: unpack requires a buffer of 2 bytes
Feb 21 23:53:20 reachy-mini launcher.sh[153892]: Exception in thread Thread-6 (_publish_status):
Feb 21 23:53:20 reachy-mini launcher.sh[153892]: Traceback (most recent call last):
Feb 21 23:53:20 reachy-mini launcher.sh[153892]:   File "/home/pollen/.local/share/uv/python/cpython-3.12.12-linux-aarch64-gnu/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
Feb 21 23:53:20 reachy-mini launcher.sh[153892]:     self.run()
Feb 21 23:53:20 reachy-mini launcher.sh[153892]:   File "/home/pollen/.local/share/uv/python/cpython-3.12.12-linux-aarch64-gnu/lib/python3.12/threading.py", line 1012, in run
Feb 21 23:53:20 reachy-mini launcher.sh[153892]:     self._target(*self._args, **self._kwargs)
Feb 21 23:53:20 reachy-mini launcher.sh[153892]:   File "/home/pollen/reachy_mini/src/reachy_mini/daemon/daemon.py", line 459, in _publish_status
Feb 21 23:53:20 reachy-mini launcher.sh[153892]:     self.zenoh_server.pub_status.put(json_str)
Feb 21 23:53:20 reachy-mini launcher.sh[153892]: zenoh.ZError: session closed
```

Clearly the problem was  backend/robot/backend.py:639:
```py
resp_bytes = self.c.async_read_raw_bytes(id, 144, 2)
resp = struct.unpack("h", bytes(resp_bytes))[0] # exception occurred here
```

Closing this because this is unnecessary.

PR #886, opened 2026-02-23
Catch exceptions in read_hardware_errors to prevent backend crash

pierre-rouanet:
## Summary

- [x] Wrap per-motor health check in `read_hardware_errors()` with try/except
- [x] This is linked to https://github.com/pollen-robotics/reachy-mini-motor-controller/pull/40 which fix a bug where reading_hardware_error could hang forever (but now can raise an error).
- [x] Still need to be tested


RemiFabre:
Tested for 6h+ continuously, never had the crash that I kept having on develop!

PR #887, opened 2026-02-23
fix(audio tests): fixing audio tests

CarolinePascal:
This PR fixes various bugs in the audio tests (record and play).

### DIFF
--- a/examples/sound_play.py
+++ b/examples/sound_play.py
@@ -22,7 +22,9 @@ def play_wav(mini: "ReachyMini", wav_path: str) -> None:
     wav_path = os.path.abspath(wav_path)
     print(f"Playing {wav_path}...")
     mini.media.play_sound(wav_path)
-    time.sleep(2)  # wait a bit for the sound to finish
+    
+    wav_duration = os.path.getsize(wav_path) / mini.media.get_output_audio_samplerate()
+    time.sleep(wav_duration)
     print("Playback finished.")
 
 
--- a/examples/sound_record.py
+++ b/examples/sound_record.py
@@ -15,6 +15,7 @@ import soundfile as sf
 from reachy_mini import ReachyMini
 from reachy_mini.media.media_manager import MediaBackend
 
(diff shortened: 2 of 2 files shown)


examples/sound_play.py
```diff
@@ -22,7 +22,9 @@ def play_wav(mini: "ReachyMini", wav_path: str) -> None:
     wav_path = os.path.abspath(wav_path)
     print(f"Playing {wav_path}...")
     mini.media.play_sound(wav_path)
-    time.sleep(2)  # wait a bit for the sound to finish
```
RemiFabre: pfff criminal

examples/sound_record.py
```diff
@@ -26,24 +26,21 @@ def main(backend: str) -> None:
...
+        print(f"Recording for {DURATION} seconds...")
+
+        start_time = time.time()
+        while time.time() - start_time < DURATION:
             sample = mini.media.get_audio_sample()
             if sample is not None:
                 audio_samples.append(sample)
-                current_nb_samples += sample.shape[0]
-
-            if mini.media.backend == MediaBackend.DEFAULT_NO_VIDEO:
```
RemiFabre: If my understanding is correct, here it doesn't use gstreamer, it uses sounddevice, that returns instant after checking a queue? Then I'd still add a small time.sleep(0.005)

RemiFabre:
The dashboard seems to show a good selection of hardware (I'm using a lite):
image


But the recording of audio using the example still fails on my end. Now the duration is correct (5s) but the file contains only 0s. This is true with both commands:
python sound_record.py --backend gstreamer_no_video

python sound_record.py

RemiFabre:
Retested on the wireless and then everything works fine.

Maybe it's only my laptop that has all these audio issues?

FabienDanieau:
ongoing. Lite is good but it doesn't work with the wireless at the moment

FabienDanieau (approved):
All good!

Issue #894, opened 2026-02-25
images of change mic fpc cable page are not loaded


FabienDanieau:
https://huggingface.co/docs/reachy_mini/troubleshooting/change_mic_fpc_cable

PR #895, opened 2026-02-25
bug #894: fix images links

FabienDanieau:
The image links were pointing to the asset from within the github repo. Since HF documentation website is located elsewhere we need to use actual http links.


augustin-crampette:
I fixed also the link of a mic cable image in the troubleshooting.md page.
You can merge @FabienDanieau if it's all right :)

Issue #896, opened 2026-02-25
Move dev dependencies in pyproject.toml


alozowski:
Currently, dev dependencies are declared under `[project.optional-dependencies]` in `pyproject.toml` – [L46](https://github.com/pollen-robotics/reachy_mini/blob/078e3e18f8a74b77e8ff8486a0b52b86d233e136/pyproject.toml#L46)

This means dev tools are treated as extras. PEP 735 introduced dependency groups, which are specifically designed for development-time dependencies. They are local-only and never published with the package. `uv` supports them natively via the `[dependency-groups]` table.

Let's move dev dependencies to:
```
[dependency-groups]
dev = [
    "pytest",
    "pytest-asyncio",
    "ruff==0.12.0",
    "onshape-to-robot==1.7.6",
    "pre-commit",
    "mypy==1.18.2",
    "types-requests",
    "types-toml",
]
```

[Example in conv app](https://github.com/pollen-robotics/reachy_mini_conversation_app/blob/5770077a3b528f68cb7f0edcd82399376457b38b/pyproject.toml#L60)

PR #898, opened 2026-02-25
Move dev dependencies to dependency-groups

alozowski:
Moves dev dependencies from `[project.optional-dependencies]` to `[dependency-groups]`. Solves https://github.com/pollen-robotics/reachy_mini/issues/896


FabienDanieau (approved):
Great!

One update is missing in the contributing.md in the mypy section

> ensure you have the same environment as the CI, you can install the package with all the extras with uv:

```bash
uv sync --all-extras --group dev
```

PR #906, opened 2026-02-28
Fix SSH hostname in quickstart docs

glutamatt:
## Summary

- Append `.local` suffix to SSH hostname in quickstart docs (`reachy-mini` → `reachy-mini.local`)

The Reachy Mini is discovered via mDNS, so the `.local` suffix is required for name resolution. Without it, `ssh pollen@reachy-mini` fails with `Could not resolve hostname`.

## Test plan

- [ ] Verify `ssh pollen@reachy-mini.local` resolves and connects successfully


FabienDanieau (approved):
Good catch thx!

PR #914, opened 2026-03-02
Fix/volume control wireless

CarolinePascal:
This PR fixes issues with audio volume control on the Wireless version :
- Defaults to `ALSA` instead of `pulsectl` when the import of `pulsectl` succeeds but the instantiation of a `Pulse` fails (happens when the daemon is run with `sudo`)
- Extend output volume control to the case when the daemon is off and no media backend is available (no sound is played)


FabienDanieau (approved):
Works on wireless. Thanks!

PR #919, opened 2026-03-03
Use main as main branch

FabienDanieau:
We'll get rid of develop to match the HF workflow

Needs this to be merged
https://huggingface.co/datasets/hf-doc-build/doc-build/discussions/56

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: 'Version tag (e.g. v1.4.0) - Leave empty for standard main build'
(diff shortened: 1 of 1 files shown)


.github/workflows/build_documentation.yml
```diff
@@ -2,22 +2,38 @@ name: Build HF documentation
...
     branches:
-      - develop
+      - main
     paths:
       - "docs/**"
       - "examples/**"
 
 jobs:
   build:
+    if: github.event_name != 'pull_request'
```
CarolinePascal: This is not required as we do not trigger the workflow on pull requests

CarolinePascal (changes requested):
Just a small comment, but all good otherwise !

PR #858, opened 2026-02-18
WIP Feat/websocket transport

pierre-rouanet:
##  Replace zenoh transport with WebSocket for SDK communication                                                                                                                                                       
                                                                                                                                                                                                                     
###  Problem                                                
                                                                                                                                                                                                                     
  Zenoh's multicast discovery causes command cross-talk when multiple Reachy Mini robots share the same WiFi network (all use default prefix reachy_mini). It also fails on WSL and some routers. The actual usage   
  pattern is simple 1:1 JSON pub/sub that doesn't need multicast.

(Most of the logic is actually shared with the WebRTC data handling).
                                                                                                                                                                                                                     
  ### Solution                                                                                                   

Replace zenoh with a WebSocket endpoint (/ws/sdk) on the existing FastAPI server. Each robot is addressed by host:port — no multicast, no topic namespacing, no collisions.

 ### Changes

- Transport-agnostic Publisher wrapping a callback, so backends call .put(data) without knowing the transport (see src/reachy_mini/io/publisher.py) and use pydantic BaseModel instead of raw dict and json dumps.
- Update tests so we start both Daemon and FastAPI app required to have the SDK client working.
- API Breaking Changes -  ReachyMini.__init__ - new parameters (backward compatible)
  - Added host: str = "reachy-mini.local" and port: int = 8000 keyword arguments
  - Existing code using positional or keyword args (robot_name, connection_mode, etc.) continues to work unchanged
  - robot_name is kept but no longer used for transport (was zenoh prefix); it's still stored and available as self.robot_name

### Tests
 - [x] Lite SDK
 - [x] Lite Conversation App
 - [x] Wireless
 - [x] Wireless Conversation App
 - [x] WebRTC demo (works on my fork here: https://huggingface.co/spaces/PierreRouanet/webrtc_example) - only slightly modify command protocol changes to make parsing cleaner daemon side.

### Optional Bonus

- [x] mDNS service discovery to find all available robots : https://github.com/pollen-robotics/reachy_mini/pull/872

PR #920, opened 2026-03-03
897 cleanup daemon vs backend start

### LINKED ISSUE #897: Cleanup daemon vs backend start
Cleanup start daemon procedure such that:

- daemon always spawns fastapi server (at __init__)
- daemon start also start backend (but allows backend to fail)

pierre-rouanet:
Make sure the backend can fail during daemon start up. 
Also add the correct error in backend status.

cc @tfrere.

---

This integrates #872 and #858 to avoid duplicate diffs. It will be squashed once they are merged.


pierre-rouanet:
Closed in favor of #929

PR #927, opened 2026-03-04
bug #926: fix logic when looking for audiocard for wireless

FabienDanieau:
when self._get_audio_device returned None, the system was using audiotest[src|sink] instead of defaulting to the reachymini_audio_[src|sink] for the wireless version.


CarolinePascal:
All good ! One small addition - It would be nice to change this log in `_get_audio_device` (`audio_gstreamer.py`) :
```
self.logger.warning(f"No {device_type} audio card found.")
```
It's misleading, we actually did not find a sound card matching "Reachy Mini Audio".

PR #958, opened 2026-03-11
Add a guide to write app ui to fit in the desktop app.

pierre-rouanet:
Generated by Claude for Claude. Can you try Remi to see if it works to update the UI or some of your apps? Or for a brand new one.

### DIFF
--- a/docs/embedded-app-ui-guide.md
+++ b/docs/embedded-app-ui-guide.md
@@ -1,0 +1,168 @@
+# Building a Web UI for the Reachy Mini Desktop App
+
+When your app defines a `custom_app_url`, its web interface can be displayed directly inside the Reachy Mini desktop app's right panel. This guide covers how to build a web UI that looks great in this embedded context.
+
+## Layout constraints
+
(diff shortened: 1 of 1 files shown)


RemiFabre:
Hello,
I think this should be reviewed by @tfrere as my front end skills are 0. I can test this, but my judgement of "is it good or not" will be very superficial. Also I've done quite a bit of front end for Reachy Mini apps with Claude lately and I still struggle a lot go the something with 0 bugs that works everywhere. It will work on my machine/browser, but users will report weird bugs of menus not selecting properly for example.

I think it would be great if Thibaud you wrote an app that serves as a great frontend example for other apps to copy. Marionette is a good example of an app that has to interact with the backend with tight timings (to synchronize stuff, to cancel moves), creates local folders/files, and communicates with the HF hub for dataset creation and download. It manages moves and sound but no camera though.

No way, have we converged on Thibaud redoing my lousy front end? Nice talk! We should do this more.
xoxo

tfrere:
Ok i'll take a look soon !

PR #962, opened 2026-03-12
Add SetFullTargetCmd to batch head/antennas/body_yaw in one WS message

RemiFabre:
Introduces a new set_full_target command that carries all three fields (head, antennas, body_yaw) as optional in a single WebSocket message, reducing serialization + send overhead from 3 round-trips to 1 in the control loop. Existing set_target() is unchanged for backwards compat.

Tested using the conversation app, there is an average reduction of 44% of call time:
Avg time to call set_target before: 4.49ms
Avg time to call set_target with these changes: 2.53ms


src/reachy_mini/reachy_mini.py
```diff
@@ -813,6 +814,68 @@ def set_target_body_yaw(self, body_yaw: float) -> None:
         """
         self.client.send_command(SetBodyYawCmd(body_yaw=body_yaw))
 
+    def set_target_full(
```
pierre-rouanet: We should directly replace the current set_target with your implementation of the set_target_full, no? Or do you see a good reason to have both?

RemiFabre: The reason I did it this way was to make it ultra safe and merge fast, have some test time with this version, then replace it completely.

I'll replace it directly

PR #964, opened 2026-03-13
Unify GStreamer media architecture across Wireless and Lite

FabienDanieau:
## Summary

The daemon now **always** starts a `GstMediaServer` that owns camera and audio hardware, regardless of platform (Wireless CM4, Lite USB, macOS, Windows, simulation). Clients choose between `LOCAL` (IPC), `WEBRTC`, or `NO_MEDIA` backends — auto-detected by default.

This replaces the previous split where Lite used OpenCV+SoundDevice and Wireless used GStreamer+WebRTC with different code paths.

### What changed

**Architecture**
- Daemon always owns camera + audio via `GstMediaServer` (renamed from `GstWebRTC`)
- Local clients read camera frames via IPC (`unixfdsink`/`unixfdsrc` on Linux/macOS, `win32ipcvideosink`/`win32ipcvideosrc` on Windows)
- Remote clients stream via WebRTC (unchanged)
- `MediaBackend` enum: `LOCAL`, `WEBRTC`, `NO_MEDIA` (old names like `gstreamer`, `sounddevice_opencv` are deprecated with warnings)

**Client-side media**
- Removed `CameraBase`/`AudioBase` ABCs, `OpenCVCamera`, `SoundDeviceAudio` — all replaced by `GStreamerCamera` and `GStreamerAudio`
- `GStreamerCamera` (IPC reader) now includes `v4l2convert` (hw-accelerated on RPi) or `videoconvert` (fallback) to handle any daemon-side frame format (YUY2 from RPi, BGR from Lite, I420 from sim)
- Audio device selection checks `.asoundrc` **first** (wireless CM4 AEC routing), only falls back to `DeviceMonitor` when absent — both daemon-side and client-side
- Added queue in audio playback pipeline to fix GStreamer buffering warnings
- Bidirectional WebRTC audio: all transceivers set to SENDRECV, `provide-clock=False` only on non-ALSA sources
- DoA extracted into standalone `audio_doa.py` module
- OpenCV is now an optional dependency (`pip install reachy_mini[opencv]`); `cv2.undistortPoints` replaced with pure numpy

**Runtime media release/acquire**
- New `release_media()` / `acquire_media()` on both `Daemon` and `ReachyMini` — lets clients tell the daemon to stop its GStreamer pipeline for direct camera/mic access (OpenCV, sounddevice, etc.)
- `ReachyMini(media_backend="no_media")` auto-releases, `__exit__` auto-acquires
- REST API: `POST /api/media/release`, `POST /api/media/acquire`, `GET /api/media/status`
- New example: `examples/custom_media_manager.py`

**Camera specs propagation**
- Daemon detects camera model and exposes specs via `GET /api/camera/specs` and `DaemonStatus.camera_specs_name`
- Client receives correct intrinsics/distortion for the connected camera
- Unknown cameras fall back to `ReachyMiniLiteCamSpecs` with a warning

**Platform fixes**
- Windows: MJPEG caps negotiation for `mfvideosrc`, named pipe path for IPC, pipe detection in `daemon_check`
- macOS: capsfilter after `avfvideosrc` for format negotiation
- RPi: `libcamerasrc` YUY2 frames now handled by client-side `v4l2convert`
- Simulation: UDP loopback unchanged (avoids `webrtcsink` async deadlock with `appsrc`)

**Docs**
- Updated architecture diagrams (Lite + Wireless unified)
- Rewrote media backend options in `python-sdk.md`
- New "Disabling Media / Direct Hardware Access" section in `media-architecture.md`
- Removed autodoc entries for deleted classes (`OpenCVCamera`, `SoundDeviceAudio`, `AudioBase`, `CameraBase`)
- Fixed all stale references to `webrtc_daemon.py` across platform docs and troubleshooting
- New example doc page: `examples/custom_media_manager.md`

**Tests**
- Video tests are self-contained (IPC source fixture, no daemon needed)
- Placo kinematics test skipped on Windows (missing dependency)
- `cv2.undistortPoints` replaced with numpy-based implementation + tests

### Breaking changes

- `media_backend="gstreamer"` and `media_backend="sounddevice_opencv"` still work but emit `FutureWarning` — use `"local"` or `"default"` instead
- `CameraBase`, `AudioBase`, `OpenCVCamera`, `SoundDeviceAudio` classes are removed
- GStreamer is now a required dependency (was optional)

## To test

- [ ] some examples (look at, sound play, sound record)
- [ ] conversation app (3 modes lite, wireless on device and wireless streaming)
- [ ] all four platforms (rpi, win, mac, osx)

:warning:  needs #948 to be merged.

### DIFF
--- a/src/reachy_mini/daemon/app/main.py
+++ b/src/reachy_mini/daemon/app/main.py
@@ -53,6 +53,8 @@ from reachy_mini.utils.wireless_version.startup_check import (
     check_and_update_wireless_launcher,
 )
 
+logger = logging.getLogger(__name__)
+
 
 @dataclass
 class Args:
@@ -112,9 +114,9 @@ def create_app(args: Args, health_check_event: asyncio.Event | None = None) -> F
             """Download datasets with logging."""
             try:
                 preload_default_datasets()
-                logging.info("Recorded move datasets pre-loaded successfully")
+                logger.info("Recorded move datasets pre-loaded successfully")
             except Exception as e:
-                logging.warning(f"Failed to pre-load some datasets: {e}")
+                logger.warning(f"Failed to pre-load some datasets: {e}")
 
         async def dataset_updater(interval_hours: float) -> None:
             """Background task that periodically checks for dataset updates."""
@@ -122,14 +124,14 @@ def create_app(args: Args, health_check_event: asyncio.Event | None = None) -> F
             while True:
                 try:
                     await asyncio.sleep(interval_seconds)
-                    logging.info("Checking for dataset updates...")
+                    logger.info("Checking for dataset updates...")
                     loop = asyncio.get_event_loop()
                     await loop.run_in_executor(None, preload_with_logging)
                 except asyncio.CancelledError:
-                    logging.info("Dataset updater task cancelled")
+                    logger.info("Dataset updater task cancelled")
                     break
                 except Exception as e:
-                    logging.warning(f"Error in dataset updater: {e}")
+                    logger.warning(f"Error in dataset updater: {e}")
 
         # Pre-download recorded move datasets in background to avoid delays on first play
         # This runs in asyncio's default ThreadPoolExecutor (fire and forget)
@@ -142,7 +144,7 @@ def create_app(args: Args, health_check_event: asyncio.Event | None = None) -> F
             dataset_updater_task = asyncio.create_task(
                 dataset_updater(args.dataset_update_interval_hours)
             )
-            logging.info(
+            logger.info(
                 f"Dataset updater started (interval: {args.dataset_update_interval_hours}h)"
             )
 
@@ -180,18 +182,18 @@ def create_app(args: Args, health_check_event: asyncio.Event | None = None) -> F
 
             # Ensure cleanup happens even if there's an exception
             try:
-                logging.info("Shutting down app manager...")
+                logger.info("Shutting down app manager...")
                 await app.state.app_manager.close()
             except Exception as e:
-                logging.exception(f"Error closing app manager: {e}")
+                logger.exception(f"Error closing app manager: {e}")
 
             try:
-                logging.info("Shutting down daemon...")
+                logger.info("Shutting down daemon...")
                 await app.state.daemon.stop(
                     goto_sleep_on_stop=args.goto_sleep_on_stop,
                 )
             except Exception as e:
-                logging.exception(f"Error stopping daemon: {e}")
+                logger.exception(f"Error stopping daemon: {e}")
 
     app = FastAPI(
         lifespan=lifespan,
@@ -290,7 +292,7 @@ def run_app(args: Args) -> None:
     handler = logging.StreamHandler(sys.stderr)
     handler.setLevel(args.log_level)
     handler.setFormatter(
-        logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+        logging.Formatter("%(name)s - %(levelname)s - %(message)s")
     )
     root_logger.addHandler(handler)
 
@@ -298,6 +300,19 @@ def run_app(args: Args) -> None:
     apps_logger = logging.getLogger("reachy_mini.apps.manager")
     apps_logger.setLevel(args.log_level)
     apps_logger.propagate = True  # Ensure it propagates to root logger
+
+    # Downgrade noisy polling routes to DEBUG in uvicorn access logs
+    class AccessLogFilter(logging.Filter):
+        _POLLING_PATHS = {"/health-check", "/api/hf-auth/relay-status"}
+
+        def filter(self, record: logging.LogRecord) -> bool:
+            msg = record.getMessage()
+            if any(path in msg for path in self._POLLING_PATHS):
+                record.levelno = logging.DEBUG
+                record.levelname = "DEBUG"
+            return True
+
+    logging.getLogger("uvicorn.access").addFilter(AccessLogFilter())
 
     # Install exception hook to catch uncaught exceptions
     def exception_hook(
@@ -360,11 +375,11 @@ def run_app(args: Args) -> None:
                     )
                     health_check_event.clear()
                 except asyncio.TimeoutError:
-                    logging.warning("Health check timeout reached, stopping app.")
+                    logger.warning("Health check timeout reached, stopping app.")
                     server.should_exit = True
                     break
                 except asyncio.CancelledError:
-                    logging.info("Health check task cancelled.")
+                    logger.info("Health check task cancelled.")
                     break
 
         try:
@@ -374,9 +389,9 @@ def run_app(args: Args) -> None:
                 )
             await server.serve()
         except KeyboardInterrupt:
-            logging.info("Received Ctrl-C, shutting down gracefully.")
+            logger.info("Received Ctrl-C, shutting down gracefully.")
         except Exception as e:
-            logging.exception(f"Error during server operation: {e}")
+            logger.exception(f"Error during server operation: {e}")
             raise
         finally:
             # Cancel health check task if it exists
@@ -390,9 +405,9 @@ def run_app(args: Args) -> None:
     try:
         asyncio.run(run_server())
     except KeyboardInterrupt:
-        logging.info("Shutdown complete.")
+        logger.info("Shutdown complete.")
     except Exception as e:
-        logging.exception(f"Error during shutdown: {e}")
+        logger.exception(f"Error during shutdown: {e}")
         sys.stderr.flush()
         raise
 
@@ -596,7 +611,7 @@ def main() -> None:
     if args.log_file:
         file_handler = logging.FileHandler(args.log_file, mode="a")
         file_handler.setFormatter(
-            logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+            logging.Formatter("%(name)s - %(levelname)s - %(message)s")
         )
         logging.getLogger().addHandler(file_handler)
         logging.getLogger().setLevel(args.log_level)
--- a/src/reachy_mini/daemon/app/services/bluetooth/bluetooth_service.py
+++ b/src/reachy_mini/daemon/app/services/bluetooth/bluetooth_service.py
@@ -5,6 +5,7 @@ Includes a fixed NoInputNoOutput agent for automatic Just Works pairing.
 """
 # mypy: ignore-errors
 
+import fcntl
 import logging
 import os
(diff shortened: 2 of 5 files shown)


cdussieux:
Tested on: 
wireless + desktopApp + Ubuntu22 + conversationApp : :white_check_mark: 
 lite + python daemon + Ubuntu22 + conversationApp : :white_check_mark:

CarolinePascal:
**Tests summary**

| Platform              | Audio tests            | Video tests | Play sound example     | Record sound example | Look at example | Remarks                                                                 |
|----------------------|------------------------|-------------|------------------------|----------------------|-----------------|-------------------------------------------------------------------------|
| macOS + Lite         | ✅                     | ✅          | ✅                     | ✅                   | ✅              | Only after the AVFoundation fix                                         |
| Windows + Lite       | ✅                     | ✅          | ✅                     | ✅                   | ✅              | Examples startup is quite slow, I get timeouts on the web socket connection |
| Windows + Wireless   | ✅ (except playback)   | ✅          | ✅ (except playback)   | ✅                   | ✅              | Examples startup is quite slow, I get timeouts on the web socket connection **half of the time** |

The timeout issues on Windows are **very** painful. I did not get the full backtrace on Windows, but we timeout when trying to connect to `ws://reachy-mini.local:8000/ws/sdk`.

src/reachy_mini/media/media_manager.py
```diff
@@ -3,187 +3,175 @@
...
-                "default": cls.GSTREAMER,
-                "default_no_video": cls.GSTREAMER_NO_VIDEO,
-            }
-            return aliases.get(value.lower())
-        return None
+
+# Mapping from deprecated enum members to their replacement
+_DEPRECATED_BACKENDS: dict[MediaBackend, MediaBackend] = {
+    MediaBackend.GSTREAMER: MediaBackend.LOCAL,
+    MediaBackend.GSTREAMER_NO_VIDEO: MediaBackend.LOCAL,
```
alozowski, line 69: Disclaimer: I can be wrong
`GSTREAMER_NO_VIDEO` and `SOUNDDEVICE_NO_VIDEO` now resolve to `LOCAL`, and `LOCAL` initialises both camera and audio. That means legacy "audio-only" backends now open camera too, which feels like a behavioral break rather than just a deprecation warning. Should these deprecated values preserve audio-only semantics until removal?

src/reachy_mini/daemon/daemon.py
alozowski: there is a new `effective_use_audio` compute above, but `_setup_backend()` still receives the raw `use_audio` value here. In the direct `Daemon(no_media=True).start(use_audio=True)` path this still instantiates backend-side audio/DoA and touches audio hardware. Should this pass `effective_use_audio` instead?

src/reachy_mini/reachy_mini.py
```diff
@@ -177,6 +184,68 @@ def media(self) -> MediaManager:
...
+            resp = requests.post(
+                f"{self._daemon_http_url}/api/media/acquire", timeout=10
+            )
+            resp.raise_for_status()
+        except requests.RequestException as e:
+            self.logger.error(f"Failed to re-acquire media on daemon: {e}")
+            return
+
+        self.media_manager.close()
+        self.media_manager = self._configure_mediamanager("default", self._log_level)
```
alozowski: `acquire_media()` always rebuilds the media manager with "default" rather than restoring the originally requested backend – this means an explicit "webrtc" or "local" choice is lost after a release/acquire cycle. Should we reuse `self._media_backend` here?

src/reachy_mini/media/camera_constants.py
```diff
@@ -298,3 +302,44 @@ class GenericWebcamSpecs(CameraSpecs):
...
+        return cast(CameraSpecs, ReachyMiniLiteCamSpecs)
+
+    cls = _SPECS_BY_NAME.get(name)
+    if cls is None:
+        _logger.warning(
+            "Unknown camera specs name %r — falling back to ReachyMiniLiteCamSpecs.",
+            name,
+        )
+        return cast(CameraSpecs, ReachyMiniLiteCamSpecs)
+    return cast(CameraSpecs, cls)
```
alozowski: I'm not sure, but it seems like `get_camera_specs_by_name()` returns the specs class, not an instance, even though the type/docstring say instance. That makes downstream checks like `isinstance(self.camera_specs, MujocoCameraSpecs)` effectively dead? Would it be better to instantiate here or switch those downstream checks to class-based logic?

src/reachy_mini/reachy_mini.py
```diff
@@ -177,6 +184,68 @@ def media(self) -> MediaManager:
...
+        After calling this, the camera and microphone are available for direct
+        access via OpenCV / sounddevice / etc.  The SDK's media_manager is
+        switched to NO_MEDIA.
+
+        Idempotent: safe to call multiple times.
+        """
+        if self._media_released:
+            return
+
+        try:
```
alozowski, line 209: This introduces raw REST calls directly inside ReachyMini, while the rest of daemon communication mostly goes through dedicated client abstractions. Could `release_media()` / `acquire_media()` live in the existing client layer instead, so transport details stay out of the sdk object?

src/reachy_mini/media/media_server.py
```diff
@@ -0,0 +1,1210 @@
...
+            self._logger.debug(f"Set provide-clock=False on {factory_name}")
+        else:
+            self._logger.debug(
+                f"{factory_name} — keeping default provide-clock behaviour."
+            )
+
+        pipeline.add(audiosrc)
+        audiosrc.link(webrtcsink)
+
+    def _build_audio_source(self) -> Optional[Gst.Element]:
```
alozowski, line 775: The audio-device detection/build logic here is now very similar to the copy in `audio_gstreamer.py`. Since both paths need to stay behaviorally aligned, could we extract a shared helper instead of maintaining two near-identical implementations?

src/reachy_mini/media/webrtc_client_gstreamer.py
```diff
@@ -152,38 +154,105 @@ def __init__(
...
+        return self.resized_K
+
+    @property
+    def D(self) -> Optional[npt.NDArray[np.float64]]:
+        """Distortion coefficients, or ``None``."""
+        if self.camera_specs is not None:
+            return self.camera_specs.D
+        return None
 
     def set_resolution(self, resolution: CameraResolution) -> None:
```
alozowski, line 191: Looks like the resolution/calibration logic here is nearly identical to `camera_gstreamer.py`, is it worth consolidating that logic so the `LOCAL` and `WEBRTC` readers do not drift over time?

src/reachy_mini/daemon/daemon.py
```diff
@@ -76,31 +80,74 @@ def __init__(
...
     def __del__(self) -> None:
         """Destructor to ensure proper cleanup."""
         self.logger.debug("Cleaning up Daemon resources...")
-        if self._webrtc is not None:
-            self._webrtc.stop()
-            self._webrtc.__del__()
-            self._webrtc = None
+        if self._media_server is not None:
+            self._media_server.stop()
+            self._media_server.__del__()
```
alozowski, line 106: `__del__()` feels a pretty fragile cleanup pattern, is it possible to make `GstMediaServer` expose an explicit `close()` / `shutdown()` method and have the daemon call that instead?

src/reachy_mini/reachy_mini.py
```diff
@@ -177,6 +184,68 @@ def media(self) -> MediaManager:
...
+    @property
+    def media_released(self) -> bool:
+        """Whether the daemon's media hardware has been released for direct access."""
+        return self._media_released
+
+    def release_media(self) -> None:
+        """Tell the daemon to release camera and audio hardware.
+
+        After calling this, the camera and microphone are available for direct
+        access via OpenCV / sounddevice / etc.  The SDK's media_manager is
```
alozowski, line 201: This docstring says `release_media()` switches the local `MediaManager` to `NO_MEDIA`, but the implementation only closes the current manager and leaves the old backend value in place. Could we either update the docs or actually replace it with a `NO_MEDIA` manager to keep runtime state self-consistent?

src/reachy_mini/media/webrtc_client_gstreamer.py
```diff
@@ -152,38 +154,105 @@ def __init__(
         self._audio_send_ready = False
         self._appsrc = None
         self._appsrc_pts = 0  # running PTS in nanoseconds for appsrc buffers
+        self._playbin: Optional[Gst.Element] = None  # for play_sound
```
alozowski, line 157: `_playbin` looks unused, should we remove it?

src/reachy_mini/media/media_server.py
```diff
@@ -0,0 +1,1210 @@
...
+                instead of a physical camera.
+
+        Raises:
+            RuntimeError: If no camera is detected (unless use_sim is True)
+                or camera specifications cannot be determined.
+
+        """
+        self._logger = logging.getLogger(__name__)
+        self._logger.setLevel(log_level)
+        self._log_level = log_level
```
alozowski, line 90: `_log_level` and `resized_K` look unused, should we remove them?

src/reachy_mini/media/webrtc_client_gstreamer.py
FabienDanieau: yeah I removed the abstract class but actually it is still useful. I'll make another PR

FabienDanieau: #981

src/reachy_mini/media/media_server.py
FabienDanieau: #981

src/reachy_mini/reachy_mini.py
FabienDanieau: right. could be addressed in #977

src/reachy_mini/media/webrtc_client_gstreamer.py
FabienDanieau: #981

src/reachy_mini/media/media_server.py
FabienDanieau: #971

src/reachy_mini/media/media_manager.py
FabienDanieau: I should not break anything but slightly less optimal since the video frames can be fetched and converted into BGR for nothing (for instance the radio app).
I'd keep things simple for now and leave it like this. If we observed performance issues we can port back the only audio feature.

PR #967, opened 2026-03-15
Remove regional access sections

alozowski:


haixuanTao:
I think this is still okay. But I would just avoid mentioning more than this.

alozowski:
@haixuanTao thank you for your review! let's close this PR then but keep in mind to not add more related info

PR #982, opened 2026-03-20
fix(symlink): fixing symlink issue with hf hub on windows

CarolinePascal:
This PR fixes an edge case where the unsupported symlink warning on Windows is not correctly caught by `snapshot_download`.


FabienDanieau (approved):
LGTM

PR #984, opened 2026-03-23
Refactor/media base classes

FabienDanieau:
Some remaining comments from https://github.com/pollen-robotics/reachy_mini/pull/964 are addressed here.

Basically this PR restore the audio and camera abstract classes since the local and webrtc implementation share some code.

### DIFF
--- a/tests/unit_tests/test_audio.py
+++ b/tests/unit_tests/test_audio.py
@@ -54,7 +54,11 @@ def test_stop_play_sound(backend: MediaBackend) -> None:
     time.sleep(1)
     media.stop_playing()
     print(f"Stopped playing sound with {backend.value} backend.")
-    assert media.audio._playbin is None, "Playbin should be None after stop_playing()"
+    if backend == MediaBackend.LOCAL:
+        # GstWebRTCClient plays sound on the daemon side — no local _playbin
+        assert media.audio._playbin is None, (
+            "Playbin should be None after stop_playing()"
+        )
     # Give a moment to confirm no crash after stopping
     time.sleep(0.5)
     media.close()


src/reachy_mini/io/ws_client.py
```diff
@@ -283,6 +284,40 @@ def wait_for_task_completion(self, task_uid: UUID, timeout: float = 5.0) -> None
...
+    def _media_request(self, path: str) -> bool:
+        """POST to a daemon media endpoint.
+
+        Returns:
+            True on success, False on failure.
+
+        """
+        url = f"http://{self.host}:{self.port}{path}"
+        try:
+            resp = requests.post(url, timeout=10)
```
FabienDanieau, line 314: @pierre-rouanet does it make sense to directly call the REST API from here, or should I prefer the websocket

pierre-rouanet: It should work. But I kind like the idea of allowing to also have the ability to do it via the web socket to be consistent.
Not really a strong argument, more of an aesthetic feeling.

src/reachy_mini/media/media_manager.py
```diff
@@ -23,11 +23,15 @@
...
 import warnings
 from enum import Enum
-from typing import Optional, Union
+from typing import Optional
 
 import numpy as np
 import numpy.typing as npt
 
+from reachy_mini.media.audio_base import AudioBase
+from reachy_mini.media.camera_base import CameraBase
```
alozowski: Here we import `AudioBase`/`CameraBase` at module import time, and `AudioBase` immediately imports `gstreamer_utils`, which imports `gi`. That means even `MediaBackend.NO_MEDIA` now eagerly loads the GStreamer stack, do we want to avoid loading GStreamer when we declare NO_MEDIA?

src/reachy_mini/media/webrtc_client_gstreamer.py
```diff
@@ -102,16 +92,14 @@ def __init__(
                 with a warning.
 
         """
-        self.logger = logging.getLogger(__name__)
-        self.logger.setLevel(log_level)
+        CameraBase.__init__(self, log_level=log_level)
+        AudioBase.__init__(self, log_level=log_level)
```
alozowski, line 100: Calling both base initialisers here also reassigns `self.logger` twice. In practice, `GstWebRTCClient` ends up logging as `reachy_mini.media.audio_base` instead of the concrete backend module, which makes debugging much harder. I'd keep logger ownership in the concrete class, or let the bases accept an existing logger instead of creating their own

src/reachy_mini/media/audio_base.py
```diff
@@ -0,0 +1,135 @@
...
+and cleanup logic so that ``GStreamerAudio`` and ``GstWebRTCClient``
+don't duplicate them.
+
+Subclasses must implement:
+- ``start_recording()``, ``stop_recording()``
+- ``start_playing()``, ``stop_playing()``
+- ``push_audio_sample()``
+- ``play_sound()``, ``upload_sound()``, ``list_sounds()``, ``delete_sound()``
+
+"""
```
alozowski, line 16: This docstring looks stale: it says subclasses must implement `upload_sound()`, `list_sounds()`, and `delete_sound()`, but I can't find them in the codebase, so let's correct the docstring

src/reachy_mini/io/ws_client.py
FabienDanieau: the REST API is kind of convenient here https://github.com/pollen-robotics/reachy-mini-os/pull/61
let's keep it that way since it's not meant to be used a lot

Issue #991, opened 2026-03-26
Expose authenticated private Hugging Face Spaces in the daemon catalog


alozowski:
The daemon's app catalog needs to include private Hugging Face Spaces that the logged-in user can access, while keeping the curated dashboard selection intact and avoiding duplicate entries for the same app. Today, private preview Spaces are not reliably returned by `/api/apps/list-available`, and naïve aggregation of the public catalog with the curated selection can produce duplicates or hide curated-only apps. I suggest a token-backed Hugging Face discovery, stable catalog deduplication by repo identity, preservation of curated-only entries, and clearer logging when authenticated discovery falls back to the public API.

PR #992, opened 2026-03-26
Expose authenticated private Hugging Face Spaces in the daemon catalog

alozowski:
Solves https://github.com/pollen-robotics/reachy_mini/issues/991. This updates the daemon catalog so `/api/apps/list-available` can return private Hugging Face Spaces accessible through the stored user token. It adds token-backed Space discovery, keeps the curated dashboard selection in the aggregate catalog, deduplicates catalog entries by stable repo identity, preserves curated-only apps, and logs when authenticated discovery falls back to the public API. This makes private preview Spaces discoverable from the desktop app without regressing the curated catalog behavior.

### DIFF
--- a/src/reachy_mini/apps/manager.py
+++ b/src/reachy_mini/apps/manager.py
@@ -49,16 +49,9 @@ class RunningApp:
 
 
 def _get_catalog_app_key(app: AppInfo) -> str:
-    """Build a stable key for deduplicating catalog entries."""
-    for key in ("id", "repo_id", "repoId", "space_id", "spaceId", "app_id", "appId"):
-        value = app.extra.get(key)
-        if isinstance(value, str) and value:
-            return value
-
-    if app.url:
-        return app.url
-
-    return app.name
+    """Return the Hugging Face space id used to deduplicate catalog entries."""
+    value = app.extra.get("id")
+    return value if isinstance(value, str) else ""
 
 
 class AppManager:
@@ -338,8 +331,11 @@ class AppManager:
         catalog_apps: list[AppInfo] = []
         seen_catalog_apps: set[str] = set()
 
-        for app in [*hf_space_apps, *dashboard_selection_apps]:
+        for app in [*dashboard_selection_apps, *hf_space_apps]:
             app_key = _get_catalog_app_key(app)
+            if not app_key:
+                catalog_apps.append(app)
+                continue
             if app_key in seen_catalog_apps:
                 continue
             seen_catalog_apps.add(app_key)
--- a/src/reachy_mini/apps/manager.py
+++ b/src/reachy_mini/apps/manager.py
@@ -48,6 +48,19 @@ class RunningApp:
     status: AppStatus
 
 
+def _get_catalog_app_key(app: AppInfo) -> str:
+    """Build a stable key for deduplicating catalog entries."""
+    for key in ("id", "repo_id", "repoId", "space_id", "spaceId", "app_id", "appId"):
+        value = app.extra.get(key)
+        if isinstance(value, str) and value:
+            return value
+
+    if app.url:
+        return app.url
+
+    return app.name
+
+
 class AppManager:
     """Manager for Reachy Mini apps."""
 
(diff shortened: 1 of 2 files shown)


src/reachy_mini/apps/sources/hf_space.py
```diff
@@ -2,30 +2,126 @@
...
+from . import hf_auth
 
 # Constants
 AUTHORIZED_APP_LIST_URL = "https://huggingface.co/datasets/pollen-robotics/reachy-mini-official-app-store/raw/main/app-list.json"
 HF_SPACES_API_URL = "https://huggingface.co/api/spaces"
 # TODO look for js apps too (reachy_mini_js_app)
-HF_SPACES_FILTER_URL = "https://huggingface.co/api/spaces?filter=reachy_mini_python_app&sort=likes&direction=-1&limit=500&full=true"
+HF_SPACES_FILTER = "reachy_mini_python_app"
+HF_SPACES_FILTER_URL = f"https://huggingface.co/api/spaces?filter={HF_SPACES_FILTER}&sort=likes&direction=-1&limit=500&full=true"
+HF_SPACES_LIMIT = 500
```
pierre-rouanet, line 18: not used in your url builder

src/reachy_mini/apps/sources/hf_space.py
```diff
@@ -2,30 +2,126 @@
...
+    return AppInfo(
+        name=space_id.split("/")[-1],
+        description=short_description,
+        url=f"https://huggingface.co/spaces/{space_id}",
+        source_kind=SourceKind.HF_SPACE,
+        extra=item,
+    )
+
+
+def _list_all_spaces_with_hf_api(token: str) -> list[SpaceData]:
```
pierre-rouanet: Using the HfApi here is nice, but should we avoid having duplicated ways of listing the hf spaces? We could provide None as token if we are not logged no?

src/reachy_mini/apps/sources/hf_space.py
```diff
@@ -2,30 +2,126 @@
...
+    return card_data if isinstance(card_data, dict) else {}
+
+
+def _normalize_space_data(space_data: SpaceData) -> SpaceData:
+    """Normalize HF API responses to the shape used by the app store."""
+    normalized = dict(space_data)
+
+    created_at = normalized.get("createdAt") or normalized.get("created_at")
+    if created_at is not None:
+        normalized["createdAt"] = created_at
```
pierre-rouanet, line 62: You could still have both 'created_at' and createdAt in your normalized dict, no? We should only keep one to normalize.

src/reachy_mini/apps/manager.py
```diff
@@ -48,6 +48,19 @@ class RunningApp:
     status: AppStatus
 
 
+def _get_catalog_app_key(app: AppInfo) -> str:
+    """Build a stable key for deduplicating catalog entries."""
+    for key in ("id", "repo_id", "repoId", "space_id", "spaceId", "app_id", "appId"):
```
pierre-rouanet: I feel this could be really dangerous and really hard to understand where the key is coming from. 
I think we should enforce a specific key. Why not user/space_name? 

We should also keep in mind that at the end, apps are python package and if two apps have different space catalog keys but the same package name, installing one will replace the other.

src/reachy_mini/apps/manager.py
```diff
@@ -284,11 +297,30 @@ async def current_app_status(self) -> Optional[AppStatus]:
...
+            self.list_available_apps(SourceKind.DASHBOARD_SELECTION),
+            self.list_available_apps(SourceKind.LOCAL),
+            self.list_available_apps(SourceKind.INSTALLED),
         )
-        return sum(results, [])
+
+        catalog_apps: list[AppInfo] = []
+        seen_catalog_apps: set[str] = set()
+
+        for app in [*hf_space_apps, *dashboard_selection_apps]:
```
pierre-rouanet: don't you want to do the opposite? dashboard first hf second?

pierre-rouanet (changes requested):

pierre-rouanet:
To be sure we are on the same page here. 
What is the problem you want to solve? Allow apps developer to install their private apps?

alozowski:
@pierre-rouanet the primary goal is to be able to install the conv app from a specific PR and verify it works in the desktop app before merging. We recently merged a PR that broke the conv app installation in the desktop app – this pipeline would have caught that. Letting devs install private apps is a nice secondary benefit, but the main thing is preventing that kind of breakage

pierre-rouanet:
Ok sounds good (and useful indeed)!

So my main concern is around the catalog_app_key that may allow an app to "erase"another app silently. Let me know if you want me to work on this as well!

src/reachy_mini/apps/manager.py
alozowski: yes, curated entries should win. I changed the merge order to dashboard first and HF second, so when the same space exists in both sources we keep the dashboard version

alozowski:
@pierre-rouanet thank you for the comments! I tried to address them: `list_all_apps()` now uses a single `HfApi` path, normalisation keeps only canonical HF keys, deduplication is now explicitly based on `extra["id"]`, and curated dashboard entries now take precedence over the broader HF catalog – I hope it's better

alozowski:
Separate note: private forks that keep the same python package name and `reachy_mini_apps` entry point as the original app will still conflict at install time, even if they have a different HF space id. I mean if, for example, I want to test `pollen-robotics/reachy_mini_conversation_app_PR273` I need to uninstall the original `pollen-robotics/reachy_mini_conversation_app`

pierre-rouanet (approved):
Looks good to me! Thanks @alozowski

PR #995, opened 2026-03-27
documentation #969: update gstreamer installation section

### LINKED ISSUE #969: Missing system dependency for uv sync --group dev on Linux
### Ticket Type

🧹 Maintenance / Documentation (Contributing)

### Operating System

Linux

### Description

The [contributing guide](https://github.com/pollen-robotics/reachy_mini?tab=contributing-ov-file#code-quality) instructs contributors to run `uv sync --all-extras --group dev` to set up the dev environment. 

On Linux, this fails to build pygobject unless libgirepository1.0-dev is installed first:
```
Package gobject-introspection-1.0 was not found in the pkg-config search path.
Command 'pkg-config --exists gobject-introspection-1.0 >= 1.64.0' returned non-zero exit status 1.
```

### Fix: add to contributing docs
`sudo apt install libgirepository1.0-dev`  # Linux only, required before uv sync

FabienDanieau:
Address this issue https://github.com/pollen-robotics/reachy_mini/issues/969

@brainwavecoder9 I've made more explicit that linux users must install gstreamer. It's not just about contributing. Wdyt?

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,5 +1,9 @@
 # Contributing
 
+## Requirements
+
+Follow the [instructions](https://huggingface.co/docs/reachy_mini/SDK/installation) to install the SDK. Please note that the Linux users have to [manually install gstreamer](https://huggingface.co/docs/reachy_mini/SDK/gstreamer-installation).
+
 ## Code quality
 
 The quality of code is insured by ruff and mypy. Please make sure to run them before pushing your code.
@@ -38,7 +42,7 @@ Please note that mpy results depend on the installed package. The [CI](../.githu
 uv sync --all-extras --group dev
 ```
 
-## Code testing
+## Code testing
 
 The code is tested with pytest. You can run the tests with the following command:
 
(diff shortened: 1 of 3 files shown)


brainwavecoder9:
I like the callouts for OS-specific steps for both GStreamer and USB permissions. It's nice to have clean breakouts for the different systems. I'm less certain about expectations around GStreamer, but it sure sounds like it's going to become a blanket requirement. _EDIT: It looks like GStreamer is moving toward a self-contained Python distribution (gstreamer-bundle), which removes the need for system-level installation on macOS and Windows for most users, but for the time being Linux still requires distro packages._

I don't think this will close #969 as libgirepository1.0-dev will be necessary to build pygobject. It would probably be good to keep that as a separate requirement in the [contributing guide](https://github.com/pollen-robotics/reachy_mini?tab=contributing-ov-file) (e.g., adding `sudo apt install libgirepository1.0-dev` as a linux-specific requirement before `uv sync --group dev`).

pierre-rouanet (approved):

PR #999, opened 2026-03-27
fix: prevent GStreamer segfault when launching app subprocesses

darkseed:
libgstpython.dylib is compiled against the system Python framework (/Library/Frameworks/Python.framework) but the desktop app uses a standalone CPython build from uv. When GStreamer's plugin scanner forks and loads this plugin, the ABI mismatch causes a segfault (exit code 255).

Additionally, the daemon's own gstreamer_bundle.pth sets GST_* env vars at startup which leak into app subprocesses, causing double-init conflicts with the app venv's own .pth file.

Fix by:
- Stripping inherited GST_* env vars before launching app subprocesses
- Setting GST_REGISTRY_FORK_DISABLE=yes to prevent the fatal fork

### DIFF
--- a/src/reachy_mini/apps/manager.py
+++ b/src/reachy_mini/apps/manager.py
@@ -107,6 +107,21 @@ class AppManager:
             app_name, self.wireless_version, self.desktop_app_daemon
         )
 
+        # Build clean environment for app subprocess.
+        # The daemon's own gstreamer_bundle.pth sets GST_* env vars at startup.
+        # Passing them to the subprocess causes double-init conflicts with the
+        # app venv's own .pth file.  Clearing them lets the subprocess set up
+        # GStreamer from scratch (if the app even needs it).
+        # GST_REGISTRY_FORK_DISABLE prevents the plugin scanner from forking:
+        # the forked scanner loads libgstpython.dylib which is compiled against
+        # the system Python framework, not the standalone CPython used here,
+        # causing a segfault.
+        env = os.environ.copy()
+        for key in list(env.keys()):
+            if key.startswith("GST_"):
+                del env[key]
+        env["GST_REGISTRY_FORK_DISABLE"] = "yes"
+
         # Launch app as subprocess with unbuffered output
         self.logger.getChild("runner").info(f"Starting app {app_name}")
         process = await asyncio.create_subprocess_exec(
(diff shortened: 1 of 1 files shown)


darkseed:
This might fix https://github.com/pollen-robotics/reachy-mini-desktop-app/issues/185 just didn't have time to test it yet.

CarolinePascal:
I confirm that setting `GST_REGISTRY_FORK_DISABLE=yes` prevents the `math` import issue in https://github.com/pollen-robotics/reachy-mini-desktop-app/issues/185 in the situation where the user has python 3.12 installed from python.org. The stripping of the GStreamer environment variables is a sensible addition as well.

Wdyt @pierre-rouanet and @cdussieux  ?

pierre-rouanet:
We will release really soon (this week) a new version of the desktop app where the whole uv venv is built at runtime to avoid those issues. I'll close the issue here as this is a desktop-app issue not Reachy-mini.

See https://github.com/pollen-robotics/reachy-mini-desktop-app/issues/185

CarolinePascal:
👋 

Replacing `env["GST_REGISTRY_FORK_DISABLE"] = "yes"` by `env.pop("PYTHONPATH", None)` makes this fix compatible with the latest runtime build introduced in `reachy_mini_desktop_app==0.9.21`!

darkseed:
Good catch, also getting rid of GST_REGISTRY_FORK_DISABLE also makes sure if the crash moves on to the main process that the main process also doesn't crash.

brainwavecoder9:
The underlying issue is fully described in https://github.com/pollen-robotics/reachy-mini-desktop-app/issues/185#issuecomment-4107190385

The `math` module error was introduced by setting uv's python-build-standalone distribution (cpython-3.12-macos-aarch64-none) as venv Python in PR 198's runtime bootstrap. The approach was a well-intended resolution for the above issue.

However, MacOS users now both simultaneously need python.org 3.12 and they cannot run with python.org 3.12. If you're not seeing an issue without python.org's 3.12 installed, then you don't have the configuration necessary to reproduce the actual problem, and any fixes may be chasing the wrong issue (or introducing new ones). With python.org's 3.12 a prewarm is not needed, and backing out 198 would resolve the math issue. 

It seems more straightforward to require an unmanaged copy of python.org's 3.12 as a temporary resolution (which some users may already have in place), rather than continuing down this path of compounding mitigations.

If libgstpython loaded cleanly, gstreamer_bundle.pth would execute without crashing, and the statically-linked Python would never be an issue.

PR #1001, opened 2026-03-30
Feature/head wobbler

### LINKED ISSUE #720: Transfer head wobbling on the core side
Currently this only works in the conv app. This should be a good feature on the core side directly. Reachy mini would be able to enable/disable it whenever, on any sound (even music or pings) and it would simplify the synchronization method (we could go back to a much more direct approach that works on anything going through the card sound?)

RemiFabre:


FabienDanieau:
**summary of my changes**

- **Merged `main`** into the branch and cleaned up the resolution (`8c6bdf2f`): the merge had dropped some main-side changes and duplicated a wobbler helper block in `audio_gstreamer.py`; fixed that plus a `protocol.py` syntax error in the `AnyCommand` union.
- **Refactored `HeadWobbler` to be PTS-driven, no thread** (`8ec95694`): got rid of the background thread + `audio_queue` + `MOVEMENT_LATENCY_S = 0.2` constant. Each incoming buffer now calls `SwayRollRT.feed(pcm)` and registers one `GLib.timeout_add` per hop on the existing pipeline GLib main loop. A generation counter invalidates pending timeouts cleanly on `stop()`/`reset()`. The `HOP_MS` spacing still comes from whichever speech-tapper module you loaded.
- **Simplified the speech tappers** (`6a4fffd2`): now that the wobbler appsink caps pin the stream to mono F32LE 16 kHz and the upstream `audioresample`/`audioconvert` do the work, dropped `_to_float32_mono` and `_resample_linear` from all 4 variants. `sample_rate` became a per-instance attribute threaded in from the audio backend (no more module-level `SR`). v2's `_bandpass_energy` takes `sample_rate` explicitly.
- **Wireless audio fixes**:
  - Per-branch `audioconvert + audioresample` after the tee so the wobbler's mono/16 kHz caps don't drag the alsasink branch into an IEC958 fallback on XMOS — that was blocking wake-up sound entirely on wireless.
  - `sync=True` on the wobbler appsink for playbin paths (`--wav` / `play_sound`): callback fires at the buffer's PTS on the pipeline clock = when the audiosink outputs it → A/V sync for free, no latency math.
  - `sync=False` kept on live push paths (push_audio_sample, WebRTC ingress) because `is-live` + PTS-sync deadlocks under back-pressure.
- **Cleanups**: `goto_sleep` now disables wobbling at entry so residual offsets don't fight the sleep pose (`c125be78`). Removed ~100 lines of dead `play_sound` scaffolding and a per-sample parent-walk replaced by a cached pipeline ref (`ce84d672`). Ruff + mypy clean across `motion/` (`e158da58`). `openapi.json` regenerated for the new `POST /api/media/wobbling/{enable,disable}` endpoints (`6e34964c`).
- **TTS demo** (`2c66d6af`): new `examples/sound_tts.py` that calls `Qwen/Qwen3-TTS` via `gradio-client` with a natural-language voice-style prompt, plays the result, wobbles the head in sync. Duration probed via `GstPbutils.Discoverer` so any decodable format works. Doc page + toctree entry added; `gradio-client` in the `examples` extra. Try with : `python examples/sound_tts.py --voice-description "Speak with joyfull voice." --wobbler-version v2`


Tests rewritten around the PTS-driven API (24 cases, `GLib.timeout_add` mocked). Validated end-to-end on desktop (pulsesink) and wireless (alsasink → XMOS), `--wav` + `--live` paths.

FabienDanieau:
Let's choose one of the 4 speech taper versions for now. We can improve / make it possible to configure in another issue

FabienDanieau (approved):
Looks good on my side. You can even try https://huggingface.co/spaces/FabienDanieau/tts-reachymini
One know bug, when using this feature with a SDK from a remote pc it's not working so well. But it seems to be an edge use case.

PR #1004, opened 2026-04-02
feat(api): add first-time setup completion flag endpoints

tfrere:
## Summary

- Adds `GET /api/config/first-time-setup` and `POST /api/config/first-time-setup` endpoints to track whether the robot has already completed its initial first-time setup
- The boolean flag is persisted to `~/.cache/reachy_mini/config.json` so it survives daemon restarts
- Follows existing router patterns (Pydantic models, `APIRouter` prefix, alphabetical registration in `main.py`)

### `GET /api/config/first-time-setup`

Returns:
```json
{ "is_completed": false }
```

### `POST /api/config/first-time-setup`

Body:
```json
{ "is_completed": true }
```

Returns:
```json
{ "is_completed": true }
```

## Test plan

- [ ] Start daemon and call `GET /api/config/first-time-setup` - should return `false` by default
- [ ] Call `POST /api/config/first-time-setup` with `{ "is_completed": true }` - should return `true`
- [ ] Call `GET /api/config/first-time-setup` again - should still return `true`
- [ ] Restart daemon and call `GET` again - should still return `true` (persisted to disk)
- [ ] Verify `~/.cache/reachy_mini/config.json` contains the expected JSON

### DIFF
--- a/src/reachy_mini/daemon/app/routers/first_wake_up.py
+++ b/src/reachy_mini/daemon/app/routers/first_wake_up.py
@@ -1,0 +1,29 @@
+"""First wake-up API routes."""
+
+import logging
+
+from fastapi import APIRouter
+from platformdirs import user_state_path
+
+router = APIRouter(prefix="/first-wake-up")
+logger = logging.getLogger(__name__)
+
+_FIRST_WAKE_UP_FLAG = user_state_path("reachy_mini") / ".first_wake_up_done"
+
+
+@router.get("/status")
+async def get_first_wake_up_status() -> dict[str, bool]:
+    """Check whether the robot has completed its first wake-up."""
+    return {"is_completed": _FIRST_WAKE_UP_FLAG.exists()}
+
+
+@router.post("/set")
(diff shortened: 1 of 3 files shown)


src/reachy_mini/daemon/app/routers/first_wake_up.py
```diff
@@ -0,0 +1,29 @@
...
+
+import logging
+from pathlib import Path
+
+from fastapi import APIRouter
+
+router = APIRouter(prefix="/first-wake-up")
+logger = logging.getLogger(__name__)
+
+_FIRST_WAKE_UP_FLAG = Path.home() / ".cache" / "reachy_mini" / ".first_wake_up_done"
```
pierre-rouanet: Maybe use something like? (I think ~/.cache would work on all os but not sure)

```python
from platformdirs import user_state_path                                                                                                                                                                           
                                                                                                                                                                                                                     
  _FIRST_WAKE_UP_FLAG = user_state_path("reachy_mini") / ".first_wake_up_done"       
```

PR #1006, opened 2026-04-02
1005 app assistant fall back to http upload when git push fails

### LINKED ISSUE #1005: App assistant: fall back to HTTP upload when git push fails
The publish command currently uses git push to deploy apps to HuggingFace Spaces. This requires git credentials, which are not available when the user has only logged in via the desktop app (OAuth token).

Add an automatic fallback: when git push fails, retry with HfApi.upload_folder(delete_patterns=["*"]), which works with the OAuth token stored by the desktop app. This preserves git history (new commit, not a force push) and produces the same end result. No changes needed for users who already have git credentials set up.

RemiFabre:
The publish command currently uses git push to deploy apps to HuggingFace Spaces. This requires git credentials, which are not available when the user has only logged in via the desktop app (OAuth token).

Add an automatic fallback: when git push fails, retry with HfApi.upload_folder(delete_patterns=["*"]), which works with the OAuth token stored by the desktop app. This preserves git history (new commit, not a force push) and produces the same end result. No changes needed for users who already have git credentials set up.

### DIFF
--- a/src/reachy_mini/apps/assistant.py
+++ b/src/reachy_mini/apps/assistant.py
@@ -650,6 +650,13 @@ def request_app_addition(new_app_repo_id: str) -> bool:
     return True
 
 
+def _git_env() -> dict[str, str]:
+    """Return env vars that prevent git from prompting for credentials."""
+    env = os.environ.copy()
+    env["GIT_TERMINAL_PROMPT"] = "0"
+    return env
+
+
 def try_to_push(console: Console, _app_path: Path) -> bool:
     """Try to push changes to the remote repository."""
     console.print("Pushing changes to the remote repository ...", style="bold blue")
@@ -658,6 +665,7 @@ def try_to_push(console: Console, _app_path: Path) -> bool:
         cwd=_app_path,
         capture_output=True,
         text=True,
+        env=_git_env(),
     )
     if push_result.returncode != 0:
         console.print(
@@ -782,6 +790,7 @@ def publish(
             cwd=app_path,
             capture_output=True,
             text=True,
+            env=_git_env(),
         )
         if pull_result.returncode != 0:
             console.print(
@@ -947,6 +956,7 @@ def publish(
             cwd=app_path,
             capture_output=True,
             text=True,
+            env=_git_env(),
         )
         if push_result.returncode != 0:
             console.print(f"[red]git push failed: {push_result.stderr}[/red]")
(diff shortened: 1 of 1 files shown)


RemiFabre:
Asked CC to review my own PR out of curiosity:

PR #1006: Fall back to HTTP upload when git push fails                                             
                                                                                                     
  What it does: When git push to a HuggingFace Space fails (e.g., user only has an OAuth token from  
  the desktop app, not git credentials), it falls back to HfApi.upload_folder(). Also adds           
  GIT_TERMINAL_PROMPT=0 to prevent git from hanging waiting for input.                               
                                                                                                     
  ---                                                                                             
  The real risk: silent data loss on the fallback path                                               
                                                                                                     
  The main concern is delete_patterns=["*"] in _upload_via_api. Combined with the relaxed pull       
  failure handling, here's a scenario that could destroy remote-only changes:                        
                                                                                                     
  1. Remote Space has changes not present locally                                                    
  2. git pull fails (no credentials) — now just a yellow warning, was previously a hard sys.exit(1)  
  3. git push fails (no credentials)                                                                 
  4. API fallback fires with delete_patterns=["*"] — wipes everything on the remote and replaces it  
  with local state                                                                                   
                                                                                                     
  The old code was safe here: pull failure = hard stop. The new code turns it into a soft warning and
   continues, which opens the door to the fallback overwriting diverged remote content without the   
  user realizing.                                                                                    
                                                                                                     
  Suggested fix: If git pull fails AND there's a .git dir (meaning git flow was expected), warn the  
  user that remote changes may be lost and require explicit confirmation before falling through to   
  the API upload path.                                                                               
                                                                                                     
  ---                                                                                             
  Other observations
                    
  - _git_env() — Good addition. Prevents git from blocking on credential prompts in automation.      
  - No-.git path — Clean. If there's no git repo locally, going straight to API upload makes sense.  
  No regression risk since that path didn't exist before.                                            
  - New-space path (bottom of diff) — The initial publish also gets the fallback, which is fine since
   there's no remote state to lose on a brand-new Space.                                             
  - exit() vs sys.exit(1) inconsistency — Pre-existing, not introduced by this PR, but the "push     
  anyway" block calls bare exit() even after sys.exit(1) on failure, which is a bit confusing.       
  - Lazy import huggingface_hub as hf inside _upload_via_api — Fine for a fallback path, though HfApi
   is already imported at module level.                                                              
                                                                                                     
  ---                                                                                                
  Verdict                                                                                            
                                                                                                  
  The _git_env() fix and the overall fallback idea are solid. The risk is specifically in the        
  existing-Space + git-repo path where a failed pull is now non-fatal but the fallback does a        
  destructive delete_patterns=["*"] upload. That combination can silently overwrite remote changes.  
  I'd flag that as the one thing to address before merging.

cdussieux (approved):

PR #1015, opened 2026-04-09
Prepare release 1.6.3

pierre-rouanet:
- macOS WebRTC audio fix (desktop app): Reverted the queue/audiorate workaround added in 1.6.2 for macOS audio and instead fixed the root cause — osxaudiosrc now keeps its default clock behaviour


FabienDanieau (approved):
LGTM

PR #1016, opened 2026-04-10
fix(simulation)

CarolinePascal:
This PR fixes a few simulation bugs on macOS  - it removes the resolution setting error, and adds support for gstreamer segfault fix.


CarolinePascal:
Ping @RemiFabre @alozowski !

alozowski (approved):
It fixed the sim run for the conv app 🔥

PR #1018, opened 2026-04-10
bug #1017: set the appsrc as not live to avoid drop of first samples

### LINKED ISSUE #1017: first words on conv app are not rendered
This happens mostly with the lite version and the conv app. When the robot says a sentence the first word could be truncated.

FabienDanieau:
This should fix the issue with the first words not being playback. It seems that it happened only with the lite.
I've double checked with oss conv app, the openai conv app and `python examples/sound_play.py --live`.


FabienDanieau:
solved by https://github.com/pollen-robotics/reachy_mini/pull/1023

PR #1022, opened 2026-04-13
tests(resolution): commenting set_resolution test with MuJoCo as it is not yet implemented

CarolinePascal:


FabienDanieau (approved):
LGTM

PR #1023, opened 2026-04-13
Reduce local GStreamer playback latency

andimarafioti:
## Summary
This reduces perceived latency on the local GStreamer playback path used by `GStreamerAudio`.

The patch is intentionally narrow:
- disable `appsrc` auto-timestamping for local playback
- assign explicit `pts` and `duration` to pushed playback buffers
- keep a playback timeline across consecutive buffers
- reset that timeline after `start_playing()`, `stop_playing()`, `clear_player()`, and after sufficiently large idle gaps
- lower sink buffering when the selected sink exposes `buffer-time` / `latency-time`
- add unit coverage for the playback timing helper

## Why
On the local playback path, sparse realtime TTS audio arrives in bursts rather than as a perfectly continuous stream.

In practice, two things were hurting responsiveness:
- the default sink buffering was large enough to report about `220 ms` minimum playback latency
- pushed buffers did not carry explicit playback timestamps, which made the first chunk after an idle gap or flush more likely to start late or get shaved

Together, that made local realtime speech feel less snappy than it should and could cut the start of a reply.

## Scope
This PR only affects the local GStreamer playback path in `GStreamerAudio`.

It does not change:
- recording
- `play_sound()`
- WebRTC audio

## Compatibility Notes
There is an explicit tradeoff here:
- lower buffering can increase underrun risk on slower systems or with burstier producers
- sink latency tuning only applies when the selected sink exposes `buffer-time` / `latency-time`
- sinks that do not expose those properties still benefit from explicit playback timestamps and reset-on-flush behavior

The playback timeline reset after a sufficiently large gap is intentional for sparse conversational audio. If we later use this same path for workloads that depend on preserving a single uninterrupted playback clock across long silences, this policy may need to be revisited.

## Validation
- `./.venv/bin/ruff check src/reachy_mini/media/audio_gstreamer.py tests/unit_tests/test_audio_gstreamer.py`
- `./.venv/bin/mypy src/reachy_mini/media/audio_gstreamer.py tests/unit_tests/test_audio_gstreamer.py`
- `./.venv/bin/pytest tests/unit_tests/test_audio_gstreamer.py -q`

### DIFF
--- a/src/reachy_mini/media/audio_gstreamer.py
+++ b/src/reachy_mini/media/audio_gstreamer.py
@@ -77,6 +77,10 @@ gi.require_version("GstApp", "1.0")
 
 from gi.repository import GLib, Gst  # noqa: E402
 
+_PLAYBACK_GAP_RESET_NS = 200 * Gst.MSECOND
+_PLAYBACK_SINK_BUFFER_TIME_US = 50_000
+_PLAYBACK_SINK_LATENCY_TIME_US = 5_000
+
 
 class GStreamerAudio(AudioBase):
     """Audio implementation using GStreamer.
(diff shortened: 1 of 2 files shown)


src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -176,6 +197,7 @@ def _init_pipeline_record(self, pipeline: Gst.Pipeline) -> None:
 
     def _init_pipeline_playback(self, pipeline: Gst.Pipeline) -> None:
         self._appsrc = Gst.ElementFactory.make("appsrc")
+        self._appsrc.set_property("do-timestamp", False)
```
FabienDanieau, line 202: false by default by we can leave it like this. At least it's explicit

src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -77,6 +77,10 @@
 
 from gi.repository import GLib, Gst  # noqa: E402
 
+_PLAYBACK_GAP_RESET_NS = 200 * Gst.MSECOND
```
FabienDanieau: I would prefer to have them within the class, see audio_base.py

FabienDanieau (approved):
Working well with Linux Lite + Wireless! 
Adding minor comments on the coding style.

src/reachy_mini/media/audio_gstreamer.py
andimarafioti: Awesome :)

src/reachy_mini/media/audio_gstreamer.py
andimarafioti: Moved the playback tuning constants into `GStreamerAudio` to match the class-constant style in `audio_base.py`, and updated the unit tests accordingly in `cbb9dcd7`. Re-ran `ruff`, `mypy`, and `pytest tests/unit_tests/test_audio_gstreamer.py -q`.

PR #1024, opened 2026-04-13
fix: authenticate HF API requests to include private spaces

tfrere:
## Summary
- Pass the stored HF token as `Authorization: Bearer` header when querying the HuggingFace Spaces API
- This allows users who have logged in (via OAuth or token) to see their private spaces tagged with `reachy_mini_python_app` in the app catalog
- Without this fix, the API only returns public spaces since the requests were unauthenticated
- The fix is backward-compatible: when no token is available, behavior is unchanged (public-only)

## Changes
- `hf_space.py`: add `_get_auth_headers()` helper that reads the HF token from `hf_auth`
- Pass auth headers to `list_all_apps()`, `list_available_apps()`, and `_fetch_space_data()`

## Test plan
- [ ] Without HF login: verify only public spaces are returned (unchanged behavior)
- [ ] With HF login (user has private spaces tagged `reachy_mini_python_app`): verify private spaces appear in the catalog with `private: true`
- [ ] Verify logout removes private spaces from the catalog

### DIFF
--- a/src/reachy_mini/apps/sources/hf_space.py
+++ b/src/reachy_mini/apps/sources/hf_space.py
@@ -2,11 +2,14 @@
 
 import asyncio
 import json
-from typing import Any, Dict
+import logging
+from typing import Any, Dict, Optional
 
 import aiohttp
 
 from .. import AppInfo, SourceKind
+
+logger = logging.getLogger(__name__)
 
 # Constants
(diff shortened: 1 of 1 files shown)


tfrere:
Closing in favor of the existing branch feat/991-private-hf-space-catalog which has a more complete implementation using huggingface_hub SDK.

PR #1027, opened 2026-04-14
[codex] Fix macOS media startup and local camera fallback

andimarafioti:
## What changed

This PR fixes two related macOS media issues:

1. `reachy-mini-daemon` could segfault during startup when GStreamer device probing ran in-process on macOS.
2. Local apps like the conversation app could still fail to get camera frames even after the daemon stayed up, because the daemon's macOS GStreamer video path was unavailable in this runtime.

The changes here:
- apply macOS GStreamer environment fixes early in daemon/media entrypoints
- avoid in-process `Gst.DeviceMonitor` on macOS by probing devices in a subprocess
- add a direct local OpenCV camera fallback for macOS local apps when no local IPC video path is available
- keep WebRTC audio active when that direct camera fallback is selected
- add regression tests for the startup path, media-manager lifecycle, fallback selection, and FFmpeg AVFoundation parsing

## Root cause

On this macOS setup, in-process GStreamer device probing was unstable and could crash the daemon. After isolating that path, the daemon no longer crashed, but it still reported `No camera found` because the runtime had no usable GStreamer macOS video source even though AVFoundation capture itself still worked locally.

That left local Python apps in a bad state:
- daemon startup survived
- WebRTC audio could connect
- camera reads stayed `None`

The new fallback avoids depending on the daemon's broken macOS GStreamer video source for local apps while preserving the existing audio path.

## Impact

- `reachy-mini-daemon` should now start reliably on affected macOS environments instead of segfaulting on startup.
- Local macOS apps can still get camera frames even when the daemon cannot expose video through its GStreamer pipeline.
- The conversation app path now works again with direct camera frames plus WebRTC audio.

## Validation

Ran locally:

- `ruff check src/reachy_mini/reachy_mini.py src/reachy_mini/media/device_detection.py src/reachy_mini/media/media_manager.py src/reachy_mini/media/camera_opencv.py`
- `pytest -q tests/unit_tests/test_reachy_mini_media_selection.py tests/unit_tests/test_media_manager.py tests/unit_tests/test_gstreamer_init.py tests/unit_tests/test_media_server.py tests/unit_tests/test_device_detection.py`

Also verified behavior manually:
- repeated `reachy-mini-daemon` startup/shutdown cycles on macOS no longer segfaulted
- local `ReachyMini(...).media.get_frame()` selected `OpenCVCamera` and returned frames
- `get_audio_sample()` returned audio again with the fallback active
- the conversation app `camera` tool produced an image payload instead of `No frame available`


andimarafioti:
After further analysis, this might have been an issue only on my machine :/

PR #1035, opened 2026-04-20
Fix/audio probe fallback

pierre-rouanet:
Fix for a corner case where the wireless mic is not working (like mine).

Stop the wireless robot's broken mic from killing the video stream as well. 
Also silenced DoA warning spam.


FabienDanieau (approved):
LGTM. conv app still working on regular setup

PR #1039, opened 2026-04-21
fix(wifi): always fall back to AP mode after forget_all and reboot

tfrere:
## Context

After using **Settings → Clear all networks** in the desktop app (which calls `POST /wifi/forget_all`), the robot stayed unreachable: no WiFi, no AP (`reachy-mini-ap`) broadcast. A reboot did **not** recover it either; only running `HOTSPOT` via Bluetooth (`bluetooth/commands/HOTSPOT.sh`) brought the AP back.

Root-cause analysis on `src/reachy_mini/daemon/app/routers/wifi_config.py` uncovered three stacked issues, all fixed here.

### 1. `forget_all` fallback never triggered (the main bug)

```python
def forget_all() -> None:
    with busy_lock:                          # lock is held here
        ...
        if get_current_wifi_mode() == WifiMode.DISCONNECTED:   # always False
            setup_wifi_connection(name="Hotspot", ..., is_hotspot=True)
```

`get_current_wifi_mode()` starts with:

```python
if busy_lock.locked():
    return WifiMode.BUSY
```

Since the worker thread runs inside `with busy_lock:`, the helper returns `BUSY`, never `DISCONNECTED`, so the hotspot is **never** brought back up after clearing all networks.

Fix: introduce `compute_wifi_mode()` (the pure detection logic, ignoring the busy flag) and capture `was_connected_to_wifi` **before** the deletion. The fallback now fires whenever we were on a non-Hotspot wifi OR the post-delete state is not HOTSPOT.

### 2. Stale `Hotspot` profile breaks `connection up` silently

`forget_all` intentionally keeps the `Hotspot` profile. On the next boot, `ensure_wifi_on_startup` calls `setup_wifi_connection`, which for an existing profile only runs `nmcli.connection.up("Hotspot")`. If the stored profile is stale (wrong `mode`, missing `ipv4.method=shared`, autoconnect disabled...) this fails silently and the retry loop just logs warnings.

Fix: new `ensure_hotspot_active()` helper that auto-heals by deleting and recreating the `Hotspot` profile via `nmcli.device.wifi_hotspot` when activation throws.

### 3. wlan0 hygiene missing in the Python paths

`bluetooth/commands/HOTSPOT.sh` does the right prep (`nmcli device disconnect wlan0` + `rfkill unblock wifi`) before the daemon restart. The Python fallbacks (`/setup_hotspot`, startup recovery, `forget_all` fallback) did not. When `wlan0` was in a transient state after a connection deletion, activation could fail for no good reason.

Fix: `_prepare_wlan0_for_hotspot()` mirrors those commands (best-effort, failures logged at debug level) and is run at the start of `ensure_hotspot_active()`.

## Summary of changes

- New `compute_wifi_mode()` pure helper; `get_current_wifi_mode()` now delegates to it.
- New `ensure_hotspot_active()` with `_prepare_wlan0_for_hotspot()` hygiene + profile auto-heal.
- `forget_all` uses `compute_wifi_mode` and a pre-deletion `was_connected_to_wifi` check.
- `/setup_hotspot`, `/connect` recovery, `/forget`, `/forget_all`, `ensure_wifi_on_startup` all route through `ensure_hotspot_active` so the recovery behavior is identical everywhere.
- Hard-coded `"Hotspot"` string replaced with `HOTSPOT_CONNECTION_NAME` constant.

No API change, no new dependency.

## Test plan

- [ ] From the desktop app, open Settings → **Clear all networks** while connected via WiFi. Verify `reachy-mini-ap` reappears within ~10 s without rebooting.
- [ ] Connect the robot to a WiFi, then **reboot**, then `Clear all networks` from the desktop app -> AP should come back.
- [ ] Manually corrupt the Hotspot profile (`nmcli connection modify Hotspot 802-11-wireless.mode infrastructure`) and reboot -> startup recovery should auto-heal the profile and bring the AP up.
- [ ] Existing flow: `POST /wifi/connect` with bad password -> daemon should still revert to hotspot.
- [ ] Bluetooth `CMD_HOTSPOT` keeps working (unchanged, but verifying no regression).

## Related

- Supersedes the earlier partial attempt on `fix/wifi-forget-all-hotspot-fallback` (based on main), which only fixed issue #1.

### DIFF
--- a/src/reachy_mini/daemon/app/routers/wifi_config.py
+++ b/src/reachy_mini/daemon/app/routers/wifi_config.py
@@ -1,6 +1,7 @@
 """WiFi Configuration Routers."""
 
 import logging
+import subprocess
 from enum import Enum
 from threading import Lock, Thread
 
@@ -10,6 +11,7 @@ from pydantic import BaseModel
 
 HOTSPOT_SSID = "reachy-mini-ap"
 HOTSPOT_PASSWORD = "reachy-mini"
+HOTSPOT_CONNECTION_NAME = "Hotspot"
 
 
 router = APIRouter(
@@ -38,18 +40,27 @@ class WifiStatus(BaseModel):
     connected_network: str | None
 
 
-def get_current_wifi_mode() -> WifiMode:
-    """Get the current WiFi mode."""
-    if busy_lock.locked():
-        return WifiMode.BUSY
+def compute_wifi_mode() -> WifiMode:
+    """Compute the real WiFi mode, ignoring busy state.
 
+    Unlike ``get_current_wifi_mode``, this never returns ``BUSY`` and can be
+    safely called from code that already holds ``busy_lock`` (e.g. worker
+    threads inside ``with busy_lock:``).
+    """
     conn = get_wifi_connections()
-    if check_if_connection_active("Hotspot"):
+    if check_if_connection_active(HOTSPOT_CONNECTION_NAME):
         return WifiMode.HOTSPOT
     elif any(c.device != "--" for c in conn):
         return WifiMode.WLAN
     else:
         return WifiMode.DISCONNECTED
+
+
+def get_current_wifi_mode() -> WifiMode:
+    """Get the current WiFi mode."""
+    if busy_lock.locked():
+        return WifiMode.BUSY
+    return compute_wifi_mode()
 
 
 @router.get("/status")
@@ -58,7 +69,7 @@ def get_wifi_status() -> WifiStatus:
     mode = get_current_wifi_mode()
 
     connections = get_wifi_connections()
-    known_networks = [c.name for c in connections if c.name != "Hotspot"]
+    known_networks = [c.name for c in connections if c.name != HOTSPOT_CONNECTION_NAME]
 
     connected_network = next((c.name for c in connections if c.device != "--"), None)
 
@@ -97,9 +108,17 @@ def setup_hotspot(
 
     def hotspot() -> None:
         with busy_lock:
-            setup_wifi_connection(
-                name="Hotspot", ssid=ssid, password=password, is_hotspot=True
-            )
+            # Use the default hotspot helper when the caller didn't override
+            # credentials so we benefit from the auto-heal behavior.
+            if ssid == HOTSPOT_SSID and password == HOTSPOT_PASSWORD:
+                ensure_hotspot_active()
+            else:
+                setup_wifi_connection(
+                    name=HOTSPOT_CONNECTION_NAME,
+                    ssid=ssid,
+                    password=password,
+                    is_hotspot=True,
+                )
 
     Thread(target=hotspot).start()
     # TODO: wait for it to be really started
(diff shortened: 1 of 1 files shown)


src/reachy_mini/daemon/app/routers/wifi_config.py
```diff
@@ -277,6 +293,54 @@ def remove_connection(name: str) -> None:
         nmcli.connection.delete(name)
 
 
+def _prepare_wlan0_for_hotspot() -> None:
```
pierre-rouanet, line 296: This should only be called if we are not already connected to the hotspot. Otherwise this will disconnect and reconnect. Or check if connected.

src/reachy_mini/daemon/app/routers/wifi_config.py
```diff
@@ -277,6 +293,54 @@ def remove_connection(name: str) -> None:
...
+    recovery paths (WiFi API + BLE command) behave identically. Any failure is
+    logged but never raised: we want to keep trying to bring the AP up even if
+    one of these preparation steps is unavailable on the host.
+    """
+    for cmd in (
+        ["nmcli", "device", "disconnect", "wlan0"],
+        ["rfkill", "unblock", "wifi"],
+    ):
+        try:
+            subprocess.run(cmd, capture_output=True, text=True, timeout=10)
```
pierre-rouanet: Why not using the nmcli lib directly? The 10s timeout seems risky, no? Is it needed?

pierre-rouanet (commented):
Bugfix looks good! 

The hygiene stuff is less clear to me. At least I think we should check in which cases we really run the prepare_wlan0_for_hostpot.

pierre-rouanet (approved):

tfrere:
Good point, you were right - the hygiene was running unconditionally, which
would kick associated clients on the AP for no reason.

Addressed in a35b1f88:
- `ensure_hotspot_active()` now early-returns when `compute_wifi_mode() ==
  WifiMode.HOTSPOT`, so neither `_prepare_wlan0_for_hotspot()` nor the
  `setup_wifi_connection()` call runs when the AP is already live.
- `_prepare_wlan0_for_hotspot()` is therefore only invoked when we are
  genuinely transitioning the interface into AP mode - i.e. coming from
  `WLAN` or `DISCONNECTED`. Its docstring now states this precondition
  explicitly so the next reader doesn't have to derive it from the call
  sites.
  
Concrete call-site behavior after this patch:

| Entry point                    | When hygiene runs                       |
|--------------------------------|-----------------------------------------|
| `POST /setup_hotspot`          | only if not already on `Hotspot`        |
| `POST /connect` recovery       | always (we just tried a WLAN connect)   |
| `POST /forget`                 | only if we were actually connected      |
| `POST /forget_all` fallback    | only if we were on non-Hotspot WiFi     |
| `ensure_wifi_on_startup`       | only if mode is `DISCONNECTED`          |

Also addressed your other comment in the same commit: `nmcli device
disconnect wlan0` now goes through the `nmcli` python binding, and the
`rfkill unblock wifi` timeout is down to 3s with an inline comment
explaining why that subprocess call is still worth keeping (combo WiFi/BT
chip soft-locks on the CM5, no equivalent in the nmcli lib).

PR #1042, opened 2026-04-21
Fix doc CI build

FabienDanieau:


FabienDanieau:
fixed by https://github.com/huggingface/doc-builder/pull/784

PR #1045, opened 2026-04-23
feat(bluetooth): add BLE commands for WiFi provisioning

tfrere:
## Summary

Expose four new BLE commands so a mobile client (e.g. the upcoming Reachy Mini mobile app) can provision Wi-Fi on a brand-new robot over Bluetooth, **without** ever having to switch the phone's Wi-Fi onto the robot hotspot.

Under the hood the Bluetooth service just proxies to the existing `/wifi/*` FastAPI routes over `http://127.0.0.1:8000`. No `nmcli` logic is duplicated: we reuse the daemon's `busy_lock`, threading and automatic hotspot fallback.

## New BLE commands

Client writes to `COMMAND_CHAR_UUID` (`…56789abcdef1`), reads/notifies on `RESPONSE_CHAR_UUID` (`…56789abcdef2`).

| Command | Auth | Response | Notes |
|---|---|---|---|
| `WIFI_STATUS` | none | compact JSON `{"mode","connected","known","error"}` | safe to poll, excluded from log filter |
| `WIFI_SCAN` | PIN | JSON array of up to 12 SSIDs, or `ERROR: ...` | blocking `nmcli rescan` (~3-10s) |
| `WIFI_CONNECT <json>` | PIN | `OK: Connecting to <ssid>` or `ERROR: ...` | payload `{\"ssid\":\"...\",\"psk\":\"...\"}`; fire-and-forget |
| `WIFI_FORGET <ssid>` | PIN | `OK: Forgotten <ssid>` or `ERROR: ...` | falls back to hotspot server-side if active |

Like today's \`CMD_*\` pattern, mutating commands require a prior `PIN_xxxxx` auth. **Unlike** `CMD_*`, `WIFI_*` commands do not reset `self.connected`, so a client can chain `PIN -> WIFI_SCAN -> WIFI_CONNECT -> WIFI_STATUS` in a single session.

`WIFI_STATUS` is unauthenticated on purpose: same rationale as the existing public `NETWORK_STATUS` characteristic, so the mobile UI can show current Wi-Fi state even before the user types the PIN.

## Why proxy to the daemon instead of calling nmcli directly?

The Bluetooth service runs as its own systemd unit under `/bluetooth/bluetooth_service.py`, with the **system** Python (not the daemon venv). Importing `wifi_config.py` would require adding `nmcli` + its deps to system Python and loading the full router module in a root-run script. Proxying over localhost:
- keeps the BT service's dependency footprint minimal (stdlib `urllib` only),
- reuses every safeguard already in `/wifi/*` (busy lock, hotspot fallback, error relay),
- stays consistent with how the desktop app talks to the same endpoints today.

## BLE MTU considerations

Responses are designed to fit in a single BLE MTU (default write MTU is ~185 bytes on iOS/Android after negotiation, up to 517):
- `WIFI_STATUS` uses short JSON keys and rarely exceeds ~200 bytes.
- `WIFI_SCAN` caps the SSID list at 12 entries.
- `WIFI_CONNECT` / `WIFI_FORGET` return short human-readable strings.

## Failure handling

All four helpers translate daemon errors into compact strings:
- 409 -> `ERROR: Busy` (an nmcli operation is in flight)
- 404 -> `ERROR: Unknown ssid` (for `WIFI_FORGET`)
- 400 -> `ERROR: Cannot forget hotspot`
- `URLError` -> `ERROR: Daemon unreachable`

For `WIFI_CONNECT`, we pre-call `/wifi/reset_error` so the client can reliably observe **this** attempt's outcome via `WIFI_STATUS`.

## Test plan

- [ ] PIN auth works as before (\`PIN_<pin>\` -> \`OK: Connected\`).
- [ ] \`WIFI_STATUS\` returns valid JSON on a freshly booted robot (hotspot mode).
- [ ] \`WIFI_SCAN\` returns a list after PIN auth; returns \`ERROR: Not connected...\` without PIN.
- [ ] \`WIFI_CONNECT {\"ssid\":\"...\",\"psk\":\"...\"}\` kicks off a real connection and \`WIFI_STATUS\` eventually reports \`mode=wlan\`, \`connected=<ssid>\`.
- [ ] Bad PSK: \`WIFI_STATUS.error\` surfaces the nmcli error and the robot falls back to hotspot (reuses existing fallback logic).
- [ ] \`WIFI_FORGET <active-ssid>\` falls back to hotspot.
- [ ] Existing commands (`PING`, `STATUS`, `JOURNAL_*`, `CMD_*`) still behave the same.
- [ ] `WIFI_STATUS` does not spam the bluetooth service logs when polled.

## Scope / non-goals

- Does **not** change the HTTP `/wifi/*` API.
- Does **not** touch the installer or `startup_check` (the change is contained in the single file copied to `/bluetooth/bluetooth_service.py`).
- Does **not** introduce any new Python dependency.


tfrere:
Closing - merged into mobile-app-integration-light at 591d6d58 (BLE WIFI commands shipped).

PR #1047, opened 2026-04-24
feat(hf-auth): POST /refresh-relay to recover zombie relay

tfrere:
## Summary

Adds `POST /api/hf-auth/refresh-relay` so clients can force the central signaling relay to drop its SSE channel and re-register with the currently stored HF token. Recovers from the "zombie relay" state where `/relay-status` claims `connected` but `/central-robot-status` returns `robots: []`, so the robot appears online to the daemon but is invisible to central.

## The bug this heals

We have observed (during mobile-app dev on a wall-powered robot kept up for days) a state where:

- `GET /api/hf-auth/relay-status` returns `{ "state": "connected", "is_connected": true }`
- `GET /api/hf-auth/central-robot-status` returns `{ "available": true, "robots": [] }`

Which means the relay is still holding an SSE channel open with the central signaling server, but central no longer lists this robot as a producer for the authenticated user. Any WebRTC client (conversation app, hand-tracker, mobile app) then stalls at "waiting for reachy..." because no `startSession` ever gets a matching producer.

Common causes observed:
- Token rotation on central-side that the relay did not pick up.
- A transient error during `setPeerStatus` that the relay swallowed without dropping its connection.
- The relay was started before any HF token was stored and never re-attempted registration after login.

Until now the only recovery was SSH + `sudo systemctl restart reachy-mini-daemon`.

## The fix

One small endpoint that piggy-backs on the existing `notify_token_change` path (the one the login flow already uses). That function cleanly tears down the SSE connection, refreshes the HF token via `huggingface_hub.get_token`, and reconnects. Works with both raw user tokens and OAuth access tokens because we go through the relay's normal reconnect logic rather than re-validating the token shape.

```python
@router.post("/refresh-relay")
async def refresh_relay() -> dict[str, Any]:
    token = hf_auth.get_hf_token()
    from reachy_mini.media.central_signaling_relay import notify_token_change
    await notify_token_change(token)
    return {"status": "requested", "token_available": bool(token)}
```

Response shape:
```json
{ "status": "requested", "token_available": true }
```

Returns `{"status": "skipped", "reason": "relay_unavailable"}` if the relay module cannot be imported (e.g. running the daemon in a stripped-down profile), so callers can still treat the endpoint as safe-to-call.

## Why not auto-heal inside the relay

The relay would need a watchdog that periodically cross-checks `setPeerStatus` against central's producer list, which means one more long-running task talking to an external service on a timer. We decided to keep the recovery explicit and client-initiated for now: the mobile app already knows how to call both `/relay-status` and `/central-robot-status`, so it can detect the desync and ask the daemon to heal. If the state turns out to happen more than rarely, we can promote the same logic into a periodic task with zero API changes.

## Test plan

- [ ] Log in, confirm `/relay-status` returns `connected` and `/central-robot-status` returns `robots: [...]` with at least one entry.
- [ ] Force the zombie state (e.g. reset the relay's registration server-side, or wait for it to happen naturally during dev).
- [ ] `POST /api/hf-auth/refresh-relay` returns `{"status": "requested", "token_available": true}`.
- [ ] `/central-robot-status` returns the robot again within ~5s.
- [ ] Logged-out case: `POST /refresh-relay` returns `{"status": "requested", "token_available": false}` (the relay transitions to `WAITING_FOR_TOKEN`, which is the correct behaviour).

Made with [Cursor](https://cursor.com).

### DIFF
--- a/src/reachy_mini/daemon/app/routers/hf_auth.py
+++ b/src/reachy_mini/daemon/app/routers/hf_auth.py
@@ -135,9 +135,15 @@ async def refresh_relay() -> dict[str, Any]:
     token = hf_auth.get_hf_token()
 
     try:
-        from reachy_mini.media.central_signaling_relay import notify_token_change
+        from reachy_mini.media.central_signaling_relay import notify_force_reconnect
 
-        await notify_token_change(token)
+        # Unconditionally drop & reconnect. We used to call
+        # `notify_token_change(token)` here, but that path is guarded
+        # by an `old_token == new_token` early-return in the relay,
+        # which turned this endpoint into a no-op whenever the token
+        # had not changed - exactly the case we ship this endpoint
+        # for. `notify_force_reconnect` skips that guard.
+        await notify_force_reconnect()
     except ImportError:
         return {
             "status": "skipped",
@@ -145,7 +151,7 @@ async def refresh_relay() -> dict[str, Any]:
             "reason": "relay_unavailable",
         }
     except Exception as e:
-        logger.warning("[refresh-relay] notify_token_change failed: %s", e)
+        logger.warning("[refresh-relay] notify_force_reconnect failed: %s", e)
         raise HTTPException(
             status_code=500, detail=f"Failed to refresh relay: {e}"
         ) from e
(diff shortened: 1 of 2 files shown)


tfrere:
Closing - merged into mobile-app-integration-light at 5174a80b (with follow-up fix 339ac5cb).

PR #1048, opened 2026-04-25
feat(webrtc): generic HTTP-proxy command for unified transport

tfrere:
## Summary

Adds an `HttpProxyCmd` to the WebRTC DataChannel command schema. A peer (typically a remote mobile/web client connected via the central signaling server) can now invoke any `/api/...` route on the daemon by sending:

```json
{
  "type": "http_proxy",
  "request_id": "<opaque>",
  "method": "GET" | "POST" | "PUT" | "DELETE" | "PATCH",
  "path": "/api/...",
  "body": <json>?,
  "headers": {<str: str>}?,
  "timeout_s": 30
}
```

The daemon forwards `(method, path, body, headers)` to its own `http://127.0.0.1:8000` server and replies on the DataChannel with:

```json
{
  "type": "http_proxy_response",
  "request_id": "<echoed>",
  "status": <int>,
  "body": <parsed json | raw text | null>,
  "headers": {<str: str>},
  "error": <str | null>
}
```

## Why

Today the only routes a remote peer can reach are the typed DataChannel commands (head pose, motors, audio, recording). Everything else - hf-auth, apps lifecycle, daemon status, recorded-move playback, etc. - lives only on the local HTTP API, which a remote mobile app cannot fetch (mixed-content from \`https://*.hf.space\`, NAT, no DNS for \`reachy-mini.local\` off-LAN).

This makes WebRTC a complete superset transport: LAN apps can keep using direct HTTP if they want low overhead, remote apps get the same surface area without each endpoint needing a typed DataChannel command. We keep the option to migrate hot paths to typed `*Cmd` schemas later for self-documentation and validation.

## Trust boundary

Authorization is enforced upstream by the WebRTC signaling layer (HF central producer auth in remote, GStreamer signaling on LAN). Whoever can establish a session with the daemon already has full robot control via the existing typed commands; the proxy does not enlarge that surface, it just unifies what the existing peer can do.

## Implementation notes

- Hop-by-hop / loopback-irrelevant headers (`host`, `content-length`, `connection`, `transfer-encoding`) are stripped from the forwarded request to avoid confusing aiohttp.
- `body` accepts either a plain string/bytes (forwarded raw) or a json-serialisable object (forwarded as `application/json`).
- Response body is auto-parsed when Content-Type contains `application/json`, otherwise returned as text.
- The dispatcher uses `asyncio.create_task` so a slow REST call cannot stall the DataChannel handler thread.
- `aiohttp` is imported lazily to avoid pulling it on platforms that disable the WebRTC stack.

## Test plan

- [x] Schema parsing validated on robot via `command_adapter.validate_json`
- [x] End-to-end round-trip: \`POST /api/hf-auth/refresh-relay\` over the dispatcher → 200 + correct body + headers + echoed \`request_id\`
- [ ] End-to-end via real WebRTC DC from the mobile app (covered by the umbrella PR work in progress on \`integration/mobile-app-daemon\`)

## Context

Part of the integration work tracked in the umbrella PR for the new mobile app, which needs transparent daemon API access whether the phone is on the LAN or remote.


tfrere:
Closing - reverted in mobile-app-integration-light at 047492de. Architecture moved to typed WebRTC DataChannel + BLE only, no generic HTTP-proxy.

PR #1056, opened 2026-04-27
Warn on SDK daemon version mismatch

andimarafioti:
## Summary

Adds an SDK-side version check using the daemon status already fetched during media manager setup. If the installed SDK package version differs from the daemon-reported package version, the SDK emits a visible `RuntimeWarning` explaining that mismatched versions can create issues.

## Validation

- `uv run pytest tests/unit_tests/test_daemon.py::test_sdk_warns_on_daemon_version_mismatch -q`
- `uv run pytest tests/unit_tests/test_daemon.py -q`
- `uv run ruff check src/reachy_mini/reachy_mini.py`
- `uv run mypy src/reachy_mini/reachy_mini.py`

### DIFF
--- a/src/reachy_mini/reachy_mini.py
+++ b/src/reachy_mini/reachy_mini.py
@@ -9,6 +9,8 @@ It also includes methods for multimedia interactions like playing sounds and loo
 import asyncio
 import logging
 import time
+import warnings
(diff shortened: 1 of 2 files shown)


FabienDanieau (approved):
All good!

PR #1070, opened 2026-05-04
draft: mobile app daemon integration (dev/mobile-app-integration)

tfrere:
## Summary

Draft integration branch for the Reachy Mini **mobile app** daemon work. This is the line we use day-to-day locally and on robots; it supersedes the older umbrella branch `integration/mobile-app-daemon` in breadth (extra signaling, health, BLE TLV, central URL default, tray-on-central, etc.).

**Not intended for merge as-is without review split** — this PR exists so Pollen can see the full diff, run CI, and decide how to land it (cherry-picks vs smaller PRs).

## Scope (high level)

- WebRTC DataChannel `http_proxy` / `ws_proxy` for unified transport
- HF auth: token bridging, refresh-relay, tray + lite registering on central
- BLE: Wi-Fi provisioning commands, `WIFI_PROBE`, TLV advert (`install_id`, `central_peer_id`, `network_mode`), pairing disabled on adapter
- Central signaling: heartbeat from welcome, producer meta / peer health, default `REACHY_CENTRAL_URL` alignment
- Stable `install_id`, `/api/daemon/identity`, macOS GStreamer DeviceMonitor guard
- Docs: `docs/known-issues/libnice-session-reuse-crash.md`

## Related

- Umbrella / tracking: #1049 (`integration/mobile-app-daemon`) — child PRs there are subsets of this line.
- Smaller upstream PRs still open: #1045–#1048, #1038, #1050–#1052, #1029 (likely superseded by TLV advert).

## Notes

- Branch: `dev/mobile-app-integration` (pushed on `origin`).
- Rebase onto `main` will be needed before any non-draft merge (branch is behind `main` by some commits).

Made with [Cursor](https://cursor.com)

PR #1075, opened 2026-05-04
fix(api): dual-mount /wifi/* under /api/wifi/* so mobile RobotClient can reach them

tfrere:
## Summary

Stacked on top of #1069 (`mobile-app-integration-light`). Carve-out from the umbrella draft #1070.

The mobile app's unified `RobotClient` prefixes all calls with `/api` (so it can swap LAN HTTP for the WebRTC HTTP-proxy data channel transparently). The `wifi_config` router was only mounted at `/wifi/*` (legacy first-boot tooling contract used by the Bluetooth provisioning service), which made the mobile app's `forgetCurrentNetwork` / status calls 404.

This PR dual-mounts the same router:

- `/wifi/*` stays as-is for the existing BLE provisioning flow and any first-boot tooling that already targets it.
- `/api/wifi/*` is the canonical path for everything that talks to the daemon through `RobotClient` (LAN HTTP and WebRTC `http_proxy`).

Both mounts point at the same router instance, so behavior is identical regardless of which prefix the client uses.

## Files

- `src/reachy_mini/daemon/app/main.py` (+5 / 0): adds `router.include_router(wifi_config.router)` next to the existing app-level mount, with comments explaining the legacy vs canonical paths.

## Test plan

- [ ] `curl http://reachy-mini.local:8000/wifi/status` returns 200 (legacy path still works).
- [ ] `curl http://reachy-mini.local:8000/api/wifi/status` returns 200 (new canonical path).
- [ ] BLE provisioning service still successfully forwards `WIFI_STATUS` / `WIFI_SCAN` / `WIFI_CONNECT` to the daemon (it uses the legacy `/wifi/*` path).
- [ ] Mobile app's `RobotClient.forgetCurrentNetwork()` no longer 404s.

## Notes

- Cherry-picked from `da70f5be` on `dev/mobile-app-integration`. No conflicts (auto-merge applied cleanly on top of the WiFi work already merged into #1069 via #1071).

### DIFF
--- a/src/reachy_mini/daemon/app/main.py
+++ b/src/reachy_mini/daemon/app/main.py
@@ -240,7 +240,12 @@ def create_app(args: Args, health_check_event: asyncio.Event | None = None) -> F
         app.include_router(cache.router)
         app.include_router(logs.router)
         app.include_router(update.router)
+        # Legacy mount at `/wifi/...` (used by the Bluetooth provisioning service
+        # and older first-boot tooling). Kept for backward compatibility.
         app.include_router(wifi_config.router)
+        # New mount under `/api/wifi/...` so the mobile app can reach it via
+        # the unified `RobotClient` (which prefixes everything with `/api`).
+        router.include_router(wifi_config.router)
 
     app.include_router(router)
     app.include_router(sdk_ws.router)


tfrere:
Closing - mobile app now does Wi-Fi setup via BLE (WIFI_SCAN/WIFI_CONNECT/WIFI_FORGET/WIFI_PROBE), no HTTP /api/wifi/* needed. Branch remains on remote.

PR #1076, opened 2026-05-04
feat(bluetooth): disable BLE pairing on the daemon adapter

tfrere:
## Summary

Stacked on top of #1069 (`mobile-app-integration-light`). Carve-out from the umbrella draft #1070.

Every GATT characteristic exposed by the daemon (commands, responses, install_id, network status) is intentionally **unencrypted** because the information it carries is also broadcast in the BLE advertisement manufacturer data. There is therefore no security benefit in having the adapter accept SMP pairing requests, and the only user-visible side-effect of being pairable was an iOS / Android "Pair this accessory?" prompt the first time the mobile app connected.

This PR drops the `NoInputNoOutput` Just Works agent registration and sets the adapter to `Pairable=False` so the mobile OS no longer surfaces a pairing dialog. The `NoInputAgent` class itself is kept in the module for reference in case a future encrypted-write characteristic warrants re-enabling bonding.

Also includes a new diagnostic note at `docs/known-issues/libnice-session-reuse-crash.md` capturing the libnice session-reuse abort observed on the central / Wi-Fi WebRTC path - independent of pairing, but written up at the same time so the investigation is not lost.

## Files

- `src/reachy_mini/daemon/app/services/bluetooth/bluetooth_service.py` (+22 / -10): drop agent registration, set `Pairable=False`, update module docstring.
- `docs/known-issues/libnice-session-reuse-crash.md` (NEW, +107): diagnostic write-up of an unrelated WebRTC libnice issue, documented for future debugging.

## Test plan

- [ ] First-time BLE connect from iOS: no "Pair this accessory?" system dialog appears.
- [ ] First-time BLE connect from Android: same, no pairing prompt.
- [ ] GATT read/write on existing characteristics (commands, responses, network status, install_id, robot_name) still works without pairing.
- [ ] `bluetoothctl show` on the daemon's adapter reports `Pairable: no`.
- [ ] No regression in BLE provisioning flow (`WIFI_STATUS`, `WIFI_SCAN`, `WIFI_CONNECT`).

## Notes

- Cherry-picked from `2958b079` on `dev/mobile-app-integration`. No conflicts.
- The libnice doc is bundled because it was authored in the same commit upstream and is read-only documentation; if reviewers prefer it as a separate PR, happy to split.

### DIFF
--- a/docs/known-issues/libnice-session-reuse-crash.md
+++ b/docs/known-issues/libnice-session-reuse-crash.md
@@ -1,0 +1,107 @@
+# libnice abort on session reuse (consumer peer ID stickiness)
+
+- **Status**: open, no fix shipped
+- **Severity**: high (kills the daemon process every few sessions on Wi-Fi/central path)
+- **Affects**: every WebRTC client going through HF central (mobile app, web SDK, conversation app on Spaces, any future client). USB/loopback path is immune.
+- **First observed**: 2026-04-30, on `reachy-mini.local` (192.168.1.19), `libnice10 0.1.22-1`, `gstreamer1.0-nice 0.1.22-1` (Debian 12 arm64).
+- **Triage author**: investigation captured during a chat session, see `agent-transcripts` for the live trace.
+
+## Symptom
+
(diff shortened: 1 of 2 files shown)


tfrere:
Closing for cleanup. Branch light/feat-ble-no-pairing remains on remote, can be reopened against mobile-app-integration-light if/when we want to ship BLE pairing hardening.

PR #1073, opened 2026-05-04
feat(daemon): stable robot identity + BLE TLV advert + WIFI_PROBE

tfrere:
## Summary

Stacked on top of #1069 (`mobile-app-integration-light`). Carve-out from the umbrella draft #1070 covering the **robot identity layer** the mobile app needs to dedupe BLE / mDNS / HF central / loopback HTTP sightings of the same physical robot into a single registry row, plus the **BLE-side consumer** of that identity (TLV advertisement) and a related **Wi-Fi diagnostic command** that lives in the same file.

Two commits, both squashed from cherry-picks:

1. `feat(daemon): stable robot identity (install_id + user-chosen name)` - the daemon-side identity layer
2. `feat(bluetooth): TLV-format BLE advert + WIFI_PROBE diagnostic` - BLE-side consumer + Wi-Fi probe

The intermediate `api_revision` bumps and the transitional "default name gate" (introduced then dropped within the same stack) have been collapsed: the squash represents the final, post-iteration state. **No new `/api/daemon/version` surface is added**; mobile clients can probe the identity capability by feature-detecting `GET /api/daemon/identity` (200 vs 404), and consult the existing `GET /update/install-source` for daemon versioning.

### HTTP API
- `GET /api/daemon/identity` -> `{ install_id, robot_name, central_peer_id }`
  - `install_id` is generated on first daemon boot and persisted to `~/.config/reachy_mini/daemon.json`. Stable across renames, token rotations, and tray reinstalls.
  - `central_peer_id` comes from the `welcome` SSE frame the relay receives from HF central (rotates on every relay reconnect).
- `GET /api/daemon/robot-name` -> `{ name, source: "default" | "persisted" | "cli" }`
- `POST /api/daemon/robot-name` -> `DaemonStatus` (live rename, propagates to mDNS + central without dropping any active WebRTC session)

### Persistence
- `daemon/app/config.py` (NEW): atomic JSON read/write under `$XDG_CONFIG_HOME/reachy_mini/daemon.json`, with `robot_name` and `install_id` fields.

### BLE GATT
- Two new characteristics: `install_id` (`12345678-1234-5678-1234-56789abcdef7`) and `robot_name` (`12345678-1234-5678-1234-56789abcdef8`), lazy-resolved through the daemon's loopback HTTP with a 5s read cache.

### Central signaling
- `meta.install_id` baked into every `setPeerStatus` producer registration.
- `notify_robot_name_change()` re-emits `setPeerStatus` without a reconnect on rename.
[...]

PR #1082, opened 2026-05-05
feat(bluetooth): WIFI_PROBE diagnostic command for first-time setup

tfrere:
## Summary

Adds a public read-only `WIFI_PROBE` BLE command that returns a JSON snapshot of the connectivity stack:

```json
{
  "wlan":     "ok | down | error",
  "gateway":  "ok | unreachable",
  "dns":      "ok | fail",
  "internet": "ok | fail",
  "daemon":   "ok | fail"
}
```

Mobile clients use this on a failed first-time Wi-Fi setup to surface a precise actionable message (\"Wi-Fi is up but DNS fails\") instead of the current generic \"the robot didn't join the network in time\" after the 60 s watchdog.

## Why

The existing `WIFI_STATUS` command only reports the high-level mode (`hotspot` / `wlan` / `disconnected` / `busy`). The user has no way to tell apart a wrong PSK, a flaky DNS, a missing default route, or a wedged FastAPI daemon. `WIFI_PROBE` is the diagnostic primitive that closes this gap.

## Design

- **Public, no auth**: same surface as `WIFI_STATUS`. A diagnostic that requires the PIN-flow auth would defeat its own purpose during PIN-flow troubleshooting.
- **Five parallel probes, strict total budget**: BlueZ treats >5 s as a write-failure on the response characteristic, so a stuck DNS resolver could otherwise propagate to the whole BLE call. Probes run on a `ThreadPoolExecutor` with `WIFI_PROBE_TOTAL_BUDGET_S = 2.5 s` on top of each probe's individual `timeout = 1.5 s`. Probes that don't complete within the budget are reported as `timeout`.
- **Probe selection**: each layer of the stack is checked independently so the result identifies *which* hop failed:
  | Probe | How |
  |---|---|
  | `wlan` | `/sys/class/net/wlan0/operstate` + `ip -4 addr` (link-up + has IPv4) |
  | `gateway` | parse `ip route show default` then `ping -c 1 -W 1 <gw>` |
  | `dns` | `socket.getaddrinfo(\"huggingface.co\")` (IPv6-aware) with timeout |
  | `internet` | HEAD `https://huggingface.co/` (4xx counts as ok - the request reached the upstream) |
  | `daemon` | GET the local FastAPI root on loopback (the BLE service is a separate systemd unit, so this is non-trivial information) |
- **No mutating side-effects**: every probe is read-only. Safe to call repeatedly from the UI.

## Why pick Hugging Face as the DNS / Internet canary?

The fleet's mobile and desktop clients depend on HF anyway (auth, central signaling). A failure here is genuinely actionable: a robot that can't reach HF can't be remote-controlled, so this is the right canary.

## Carve-out from #1073

This is the *minimal isolated slice* derived from the `WIFI_PROBE` portion of #1073 (\"stable robot identity + BLE TLV advert + WIFI_PROBE\"). Everything else from that draft (TLV-format advert, `install_id`, `robot_name` propagation, `/api/daemon/identity`) is intentionally **out of scope** here so the diagnostic can land without coupling to the larger identity refactor.

## Files

- `daemon/app/services/bluetooth/bluetooth_service.py`: 5 probe helpers (`_probe_wlan`, `_probe_gateway`, `_probe_dns`, `_probe_internet`, `_probe_daemon`) + `_wifi_probe()` + `WIFI_PROBE` dispatch case in `_handle_command`. Imports of `concurrent.futures` and `re` added at the top.

Single file, +210 / -5.

## Test plan

On a Wi-Fi-connected robot, healthy stack:

```bash
# Send WIFI_PROBE over BLE and read the response characteristic.
{"wlan":"ok","gateway":"ok","dns":"ok","internet":"ok","daemon":"ok"}
```

Total wall-clock under 1 s on a healthy stack.

Failure injection (each isolated, run in sequence):

- [ ] Forget current Wi-Fi: expect `wlan: down`, downstream probes `unreachable` / `fail`
- [ ] `echo \"\" | sudo tee /etc/resolv.conf`: expect `dns: fail`, `internet: fail`, others `ok`
- [ ] `sudo iptables -A OUTPUT -d 1.1.1.1 -j DROP` then a route-blocking equivalent: expect `internet: fail` without affecting `dns` (which uses local resolver) or `gateway`
- [ ] `sudo systemctl stop reachy-mini-daemon`: expect `daemon: fail` while the BLE service stays up and answers
- [ ] `sudo tc qdisc add dev wlan0 root netem delay 5000ms`: expect the probe call to return within `WIFI_PROBE_TOTAL_BUDGET_S` with `timeout` markers on the affected probes

## Out of scope

- **Mobile-app consumption**: the mobile app hooks the new command into its WiFi setup screen in a separate change; the daemon side is fully usable independently for any client (CLI, scripts, ...).
- **Versioned BLE TLV advertisement** (`install_id`, `central_peer_id`, `network_mode`): drafted in a related branch but kept out of this PR to keep scope minimal.
- **Auth around `WIFI_PROBE`**: read-only by design; no PIN required (consistent with `WIFI_STATUS`).

### DIFF
--- a/src/reachy_mini/daemon/app/services/bluetooth/bluetooth_service.py
+++ b/src/reachy_mini/daemon/app/services/bluetooth/bluetooth_service.py
@@ -5,10 +5,12 @@ Includes a fixed NoInputNoOutput agent for automatic Just Works pairing.
 """
 # mypy: ignore-errors
 
+import concurrent.futures
 import fcntl
 import json
 import logging
 import os
+import re
 import socket
 import subprocess
 import urllib.error
@@ -765,12 +767,15 @@ class BluetoothCommandService:
             else:
                 return "ERROR: Incorrect PIN"
 
-        # WiFi provisioning commands. WIFI_STATUS is public (read-only snapshot);
-        # mutating commands require prior PIN authentication. Unlike CMD_*, we do
-        # NOT reset `self.connected` afterwards so a client can chain
-        # scan -> connect -> poll status in a single provisioning session.
+        # WiFi provisioning commands. WIFI_STATUS and WIFI_PROBE are public
+        # (read-only snapshots); mutating commands require prior PIN
+        # authentication. Unlike CMD_*, we do NOT reset `self.connected`
+        # afterwards so a client can chain scan -> connect -> poll status
+        # in a single provisioning session.
         elif upper == "WIFI_STATUS":
             return _wifi_status()
+        elif upper == "WIFI_PROBE":
+            return _wifi_probe()
         elif upper == "WIFI_SCAN":
             if not self.connected:
                 return "ERROR: Not connected. Please authenticate first."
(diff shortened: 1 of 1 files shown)


tfrere:

PR #1083, opened 2026-05-05
feat(central-relay): advertise meta.transport for USB/Wi-Fi tagging

tfrere:
**Stacked on #1081** (heartbeat). Single-field addition to the producer ``meta`` so listeners can badge robots as USB or Wi-Fi from the central listing alone.

## Why

The mobile picker has three sections - **Local USB**, **BLE**, **Distant (Central)** - and the same physical robot can appear in two of them at once. A tray-spawned daemon also registers with HF central, so it shows up in *Distant* alongside *Local USB*. Today there is no way to tell those rows apart from central alone, and the picker either renders duplicates or has to cross-reference ``robotName`` (which collides on the default ``reachy_mini``).

``meta.transport`` is the smallest authoritative signal that closes the gap. The daemon already knows whether it was launched by the tray (via the ``--desktop-app-daemon`` CLI flag stored on ``self.desktop_app_daemon``); this PR plumbs that bit through to the producer ``meta`` so listeners see it without needing a side channel.

## Design

- **One field, free-form string.** ``meta.transport`` is ``\"usb\"`` for a tray-spawned daemon and ``\"wifi\"`` otherwise. Free-form rather than a typed enum so future fronts (``\"ethernet\"``, ``\"sim\"``, ``\"mockup\"``) reach listeners without a relay change. Listeners that don't recognise a value fall back to *Wi-Fi* by convention.
- **Single source of truth.** New ``_build_producer_meta()`` returns the dict; ``_producer_status_payload`` embeds it verbatim. Both the post-welcome registration round-trip and the periodic heartbeat re-emission go through this path, so a future field (``install_id``, ``capabilities``, ``health``, ``version``) cannot drift between the two.
- **Discriminator on the daemon side.** ``daemon.py`` maps ``self.desktop_app_daemon -> \"usb\" else \"wifi\"``. The relay itself stays transport-agnostic and forwards whatever the daemon decided.
- **Default ``\"wifi\"``.** A direct ``CentralSignalingRelay(...)`` call matches the autonomous Pi-side case (Wireless variant or Lite hosted on its own Pi). The desktop tray opts in to ``\"usb\"`` because it is the special case.

## Wire shape

Before:

\`\`\`json
{ \"type\": \"setPeerStatus\", \"roles\": [\"producer\"], \"meta\": {\"name\": \"reachymini\"} }
\`\`\`

After:

\`\`\`json
{
  \"type\": \"setPeerStatus\",
  \"roles\": [\"producer\"],
  \"meta\": {\"name\": \"reachymini\", \"transport\": \"usb\"}
}
\`\`\`

Central forwards ``meta`` verbatim already, so **no central change is needed**. Pre-feature daemons keep emitting ``{name}`` only and listeners must default ``transport=undefined -> \"wifi\"``.

## Files

| File | What |
|---|---|
| ``media/central_signaling_relay.py`` | New ``transport`` ctor arg, new ``_build_producer_meta()`` method, ``_producer_status_payload()`` calls it, ``start_central_relay`` factory forwards the param. |
| ``daemon/daemon.py`` | Compute ``\"usb\" if self.desktop_app_daemon else \"wifi\"`` and pass it to ``start_central_relay``. |
| ``tests/unit_tests/test_central_signaling_relay.py`` | 6 new tests guarding the wire shape. |

Total: **+122 / -1**.

## Tests

20 total (14 heartbeat + 6 new):

\`\`\`
test_meta_carries_robot_name                       PASSED
test_meta_carries_transport_wifi_by_default        PASSED
test_meta_carries_transport_usb_when_set           PASSED
test_meta_forwards_unknown_transport_value_verbatim PASSED
test_meta_shape_is_minimal                         PASSED
test_meta_used_by_producer_status_payload          PASSED
\`\`\`

The shape-locking test (``test_meta_shape_is_minimal``) is the contract guard: any future field addition forces a new test rather than silently editing this one.

- [ ] Tray daemon: launch with ``--desktop-app-daemon``, observe ``meta.transport == \"usb\"`` in central's ``/api/robot-status``.
- [ ] Pi-side daemon (Wi-Fi): no flag, observe ``meta.transport == \"wifi\"``.
- [ ] Mobile picker (separate change): badge cards from ``meta.transport``.

## Out of scope

This is the smallest sensible slice of the broader producer-meta contract drafted on ``dev/mobile-app-integration``. The remaining pieces each warrant their own PR with their own test plan and will land separately on top of this one:

- **Withdraw()** distinct from ``stop()`` for graceful exit (kills 30 s phantom on shutdown)
- **meta.health + error_code** with the ``peer_health.py`` truth table for visible degraded states
- **\"Tray with no hardware\"** 30 s watchdog → auto-withdraw on USB unplug
- **debounced ``update_producer_meta()``** for live state transitions

The new ``_build_producer_meta()`` is the extension point for those follow-ups: each future field is one line in that helper plus its own unit test.

### DIFF
--- a/src/reachy_mini/daemon/daemon.py
+++ b/src/reachy_mini/daemon/daemon.py
@@ -178,10 +178,21 @@ class Daemon:
         try:
             from reachy_mini.media.central_signaling_relay import start_central_relay
 
+            # ``transport`` reflects how a remote client physically
+            # reaches this daemon: a tray-spawned daemon means the robot
+            # is plugged into the user's desktop over USB; an autonomous
+            # Pi-side daemon (Wireless variant or Lite hosted on its own
+            # Pi) is reached over Wi-Fi (or LAN). We use ``desktop_app_daemon``
+            # as the discriminator since it is the single authoritative
+            # signal we already have on the daemon and it cannot lie:
+            # only the tray launches the daemon with that flag.
+            transport = "usb" if self.desktop_app_daemon else "wifi"
+
             self.logger.info("Starting central signaling relay...")
             await start_central_relay(
                 hf_token=hf_token,
                 robot_name=self.robot_name,
+                transport=transport,
                 robot_app_lock=self.robot_app_lock,
             )
             self.logger.info("Central signaling relay started")
--- a/src/reachy_mini/media/central_signaling_relay.py
+++ b/src/reachy_mini/media/central_signaling_relay.py
@@ -100,6 +100,7 @@ class CentralSignalingRelay:
         local_uri: str = LOCAL_GSTREAMER_SIGNALING,
         hf_token: Optional[str] = None,
         robot_name: str = "reachymini",
+        transport: str = "wifi",
         on_state_change: Optional[Callable[["RelayState", Optional[str]], None]] = None,
         robot_app_lock: Optional[RobotAppLock] = None,
     ):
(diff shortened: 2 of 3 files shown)


tfrere:

PR #1085, opened 2026-05-05
feat(bluetooth): TLV-format BLE advertisement with hardware_id

tfrere:
**Stacked on #1084** (``meta.hardware_id``). Replaces the IPv4-list ManufacturerData with a TLV stream so mobile clients can dedupe BLE sightings against the same robot's Local USB / Distant rows **at scan time**, without GATT-connecting (a connect breaks the scan and needs pairing on iOS).

## Why

The mobile picker's three sections (Local USB, BLE, Distant) now all expose ``hardware_id`` *except* BLE pre-connect: the legacy advert encoded IPv4 addresses, which mobile clients never consumed (they read them from the GATT ``NETWORK_STATUS`` characteristic anyway). Without a stable id in the advert, the picker either renders duplicate cards or has to GATT-connect to every advertised peripheral just to learn its identity.

This PR repurposes the 16 manufacturer-data bytes the legacy IPv4 code already owned and ships ``hardware_id_prefix`` instead. A mobile parser consumes it from ``manufacturerData`` directly in the scan callback and assigns ``device.hardwareId`` synchronously - a single unified card per robot becomes possible from the moment a beacon arrives.

## Wire layout

\`\`\`
byte 0      : format version (0x02)
bytes 1..N  : TLV blocks  -  [tag][len][value]

  tag 0x01 (8 bytes)  hardware_id_prefix
                      first 8 bytes of the SHA-256 prefix that
                      get_hardware_id() exposes elsewhere
                      (16 hex chars = 8 binary bytes); omitted
                      when get_hardware_id() returns None.
\`\`\`

When no Reachy is attached, the manufacturer-data section is omitted entirely - the advert still registers with just Flags + LocalName.

## Budget

Legacy advert = 31 bytes total. Per AD struct: 2 bytes overhead (length + type) + payload.

| AD | Bytes |
|---|---:|
| Flags | 3 |
| Complete Local Name \"ReachyMini\" | 12 |
| ManufacturerData (mfg id 2 + version 1 + TLV 10) | 15 |
| **Total** | **30** |

Margin: **1 byte**. A future TLV tag MUST keep the AD under 31 bytes total or BlueZ silently drops the advert.

## Why drop the IPv4 list

- Every existing mobile / desktop consumer reads IPs from the GATT ``NETWORK_STATUS`` characteristic, not the advert.
- IP info is not stable per robot (changes on every Wi-Fi rejoin), so it adds churn to the advert-refresh loop without any scan-time consumer benefit.
- Reclaiming those 16 bytes is what makes room for a stable id.

## Files

| File | What |
|---|---|
| ``daemon/app/services/bluetooth/bluetooth_service.py`` | New ``encode_advert_payload(hardware_id_hex)``, two TLV constants, two ``adv.manufacturer_data`` call sites updated, legacy ``encode_network_ips()`` + ``_MAX_ADVERTISED_IPS`` removed. |
| ``tests/unit_tests/test_bluetooth_advert.py`` (NEW) | 8 tests locking the wire shape. |

Total: **+260 / -44** across 1 file modified + 1 file added.

## Tests

8 new pytest cases all green:

\`\`\`
test_encode_emits_version_byte_then_hwid_tlv          PASSED
test_encode_returns_empty_when_hwid_is_none           PASSED
test_encode_returns_empty_on_short_hwid               PASSED
test_encode_returns_empty_on_non_hex                  PASSED
test_encode_uses_only_first_8_bytes_of_hwid           PASSED
test_encode_payload_total_size_under_budget           PASSED
test_format_version_is_v0x02                          PASSED
test_hwid_tlv_tag_and_len_are_stable                  PASSED
\`\`\`

The shape-locking constants (``_BLE_ADVERT_FORMAT_VERSION``, ``_TLV_TAG_HARDWARE_ID_PREFIX``, ``_TLV_HARDWARE_ID_PREFIX_LEN``) are asserted explicitly so a future refactor cannot silently break every parser in the fleet.

- [ ] Pi-side daemon with Reachy attached: ``hcitool lescan -d`` shows manufacturer data ``02 01 08 <8-byte hwid>``.
- [ ] Hex-decode the 8 bytes: must match the first 16 chars of ``GET /api/daemon/hardware-id``.
- [ ] Disconnect the audio device at runtime: next 10 s refresh tick drops the manufacturer data entry from the advert.
- [ ] Mobile picker (separate change): a single BLE card per robot with ``id:<5 hex chars>`` matching the same robot's Distant / Local USB cards.

## Backwards compatibility

- Daemon shipping this code + mobile client that does not parse TLV yet -> BLE list shows no hwid (same as pre-PR state).
- Pre-PR daemon + TLV-aware mobile client -> ``device.hardwareId = null``, picker falls back to whatever it had before.

No coordinated rollout needed.

## Out of scope

- **Mobile-app TLV parser**: lives in the consuming app, not in this PR.
- **Additional TLV tags** (``network_mode``, ``capabilities``, ...): the format is designed for them, but adding any pushes past the 1-byte budget margin. Tracked separately.

### DIFF
--- a/src/reachy_mini/daemon/app/services/bluetooth/bluetooth_service.py
+++ b/src/reachy_mini/daemon/app/services/bluetooth/bluetooth_service.py
@@ -852,7 +852,7 @@ class BluetoothCommandService:
         ad_manager = dbus.Interface(adapter, LE_ADVERTISING_MANAGER_IFACE)
         self.adv = Advertisement(self.bus, 0, "peripheral", self.device_name)
         self.adv.service_uuids = [REACHY_STATUS_SERVICE_UUID]
-        self.adv.manufacturer_data = encode_network_ips()
+        self.adv.manufacturer_data = encode_advert_payload(get_hardware_id())
         self._ad_manager = ad_manager
         ad_manager.RegisterAdvertisement(
             self.adv.get_path(),
@@ -863,7 +863,11 @@ class BluetoothCommandService:
             ),
         )
 
-        # Refresh both GATT network characteristic and advertisement payload
+        # Refresh both GATT network characteristic and advertisement payload.
+        # The advert payload only changes if the audio device is hot-(un)plugged,
+        # which is rare; the periodic re-emit is mostly for the GATT
+        # network char update + a re-register safety net in case BlueZ
+        # ever loses our advertisement.
         GLib.timeout_add_seconds(10, self._refresh_network_info)
 
         logger.info(f"✓ Bluetooth service started as '{self.device_name}'")
@@ -872,7 +876,7 @@ class BluetoothCommandService:
         """Periodic callback: update GATT characteristic and advertisement."""
         self.app.reachy_status.update_network_status()
 
-        new_data = encode_network_ips()
+        new_data = encode_advert_payload(get_hardware_id())
         if new_data != self.adv.manufacturer_data:
             self.adv.manufacturer_data = new_data
             try:
@@ -881,7 +885,7 @@ class BluetoothCommandService:
                     self.adv.get_path(),
                     {},
                     reply_handler=lambda: logger.info(
-                        "Advertisement re-registered with updated IPs"
+                        "Advertisement re-registered with updated payload"
                     ),
                     error_handler=lambda e: logger.error(
                         f"Failed to re-register advertisement: {e}"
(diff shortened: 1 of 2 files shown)


tfrere:

PR #1089, opened 2026-05-05
Added missing reshape for webRTC goto command

### LINKED ISSUE #1088: Missing reshape() in goto command when sent through webRTC

RemiFabre:
(the bug was not visible from the Python SDK because the pose was not serialized)

### DIFF
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -922,7 +922,7 @@ class Backend:
             send_response({"status": "ok", "command": "set_full_target"})
 
         elif isinstance(cmd, GotoTargetCmd):
-            head = np.array(cmd.head) if cmd.head else None
+            head = np.array(cmd.head).reshape(4, 4) if cmd.head else None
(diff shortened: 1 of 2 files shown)


RemiFabre:
Added 3 unit tests:

1. test_goto_target_reshapes_flat_head_to_4x4 — primary regression test. Sends a GotoTargetCmd with a flat 16-element head (built from a non-identity 4x4 to catch any axis-mixing) and asserts _async_goto receives a (4, 4) array equal to the original pose.
2. test_goto_target_passes_none_when_head_omitted — antennas-only goto, asserts head stays None (the if cmd.head branch). Guards against a naive future "always reshape" change that would crash on None.
3. test_goto_target_via_webrtc_json_message — full WebRTC entry path. Feeds a JSON string through _handle_webrtc_message, which is the exact code path the bug lived on. Asserts the same (4, 4) outcome end-to-end through command_adapter parsing.

I confirmed test 1 and 3 fail on the unfixed code (they report shape == (16,)), and all three pass on the fixed code. Test 2 passes on both (it guards a different thing).

pierre-rouanet (approved):
Thanks, LGTM!

PR #1094, opened 2026-05-07
feat(sdk): expose robot.preselectedRobotId from URL hint

tfrere:
## Summary

Pairs with the credential hand-off shipped in #1093. When an iframe host (typically the Reachy Mini mobile shell) embeds an SDK consumer, it can now ALSO tell the consumer which robot to target so the user does not get bounced through a redundant robot picker after we silently logged them in via the URL fragment.

## What changes

\`js/reachy-mini.js\`

- New private helper \`readPreselectedRobotIdFromUrl()\` (next to \`consumeFragmentCredentials\`) which tries
  1. the URL fragment (\`#robot_peer_id=…\`),
  2. the URL query (\`?robot_peer_id=…\`),
  3. otherwise returns \`null\`.

- Constructor stores the result on \`this._preselectedRobotId\` once, at instantiation. Subsequent URL mutations (\`consumeFragmentCredentials()\` strips the auth keys, history navigation, in-app routing) do not move the target out from under the consumer.

- Public read-only getter \`robot.preselectedRobotId\` returning \`string | null\`.

- Top-of-file READ-ONLY PROPERTIES JSDoc updated.

## Usage

Apps opt in with two lines after \`authenticate()\` succeeds:

\`\`\`js
if (await robot.authenticate()) await robot.connect();
await robot.startSession(robot.preselectedRobotId ?? pickedId);
\`\`\`

That is enough to skip the robot picker entirely when a host has already chosen a target.

## Why query AND fragment

The peer id is NOT a secret — the central signaling server publishes it on its public robot listing — so a query param is fine. The Reachy Mini mobile shell already emits \`?robot_peer_id=…\` today via [\`buildEmbedUrl.ts\`](https://github.com/pollen-robotics/reachy_mini_mobile_app/blob/main/src/apps/buildEmbedUrl.ts) and remains unchanged. Hosts that prefer the fragment for symmetry with #1093's credentials hand-off get identical behaviour. Both reads are tolerant of malformed input — we fall through to \`null\` rather than throwing.

## Pairs with

- #1093 (credential URL-fragment hand-off, already merged).
- A follow-up on \`reachy_mini_minimal_conversation\`'s \`src/main.ts\`: read \`robot.preselectedRobotId\` after \`authenticate()\`, set \`selectedRobotId\` directly, skip the picker. (Will land separately so SDK and app cycles are independent.)

## Test plan

- [ ] Standalone Space load (no \`robot_peer_id\` anywhere in URL) → \`robot.preselectedRobotId === null\`, picker behaviour unchanged.
- [ ] Mobile-shell embed with \`?robot_peer_id=abc123\` → \`robot.preselectedRobotId === 'abc123'\`.
- [ ] Hosts using \`#robot_peer_id=abc123\` (fragment-only) → same.
- [ ] Both query and fragment present → fragment wins (consistent with the credentials hand-off precedence).
- [ ] Malformed query / fragment → no throw, \`preselectedRobotId\` is \`null\`.

### DIFF
--- a/js/reachy-mini.js
+++ b/js/reachy-mini.js
@@ -94,6 +94,12 @@
  *   .micSupported     boolean           — true if robot offers bidirectional audio
  *   .micMuted         boolean           — your microphone mute state
  *   .audioMuted       boolean           — robot speaker mute state (local)
+ *   .preselectedRobotId string | null   — peer id from `?robot_peer_id=` /
+ *                                          `#robot_peer_id=`; null if absent.
+ *                                          Use it to skip your robot picker
+ *                                          when a host iframe (e.g. the
+ *                                          Reachy Mini mobile shell) embeds
+ *                                          this app.
  *
  *
  * EVENTS  (EventTarget — use addEventListener)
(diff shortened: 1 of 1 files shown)


tfrere:

PR #1097, opened 2026-05-07
fix(sdk): buffer ICE candidates that arrive before setRemoteDescription

tfrere:
## Summary

- In cross-origin iframes (mobile shell, vibe-coder preview) and in Safari / iOS WKWebView, the SSE-based signaling transport delivers `peer.ice` messages out-of-order with the `peer.sdp` offer. Calling `addIceCandidate` before `setRemoteDescription` throws `InvalidStateError: The remote description was null`, the error is swallowed by the catch block in `_handlePeerMessage`, and the candidate is silently dropped. On adverse networks this can wedge ICE and `startSession()` hangs until the surrounding timeout fires.
- Fix: queue ICE candidates received while `pc.remoteDescription` is `null` into `this._pendingRemoteIce` and replay them right after `setRemoteDescription` resolves. The buffer is reset at the top of every `startSession()` so stale candidates from a previous attempt cannot leak into a fresh `RTCPeerConnection`.

This is a no-op for callers running in well-behaved Chromium contexts (same-origin, desktop) where messages were already in order; only the slow / out-of-order path changes behaviour.

## Why this matters

Reproduced from the mobile-shell iframe pointing at the central relay:

```
[Log] [doStart] calling robot.startSession( ... )...
[Error] WebRTC error: – InvalidStateError: The remote description was null
    _handlePeerMessage (reachy-mini.js:1396)
```

After the fix the candidates land normally and `startSession()` resolves with the expected datachannel.

## Test plan

- [ ] Manual: load a Space hosting the SDK inside the Reachy Mini mobile shell iframe, observe `startSession()` succeeding without `InvalidStateError`.
- [ ] Manual: same flow on Chrome desktop (same-origin) - no behavioural change expected.
- [ ] Manual: confirm a stop / start cycle does not replay candidates from the previous session (buffer reset on `startSession`).

### DIFF
--- a/js/reachy-mini.js
+++ b/js/reachy-mini.js
@@ -539,6 +539,11 @@ export class ReachyMini extends EventTarget {
         this._iceConnected = false;
         this._dcOpen = false;
         this._micSupported = false;
+        // Buffer for ICE candidates that arrive before the SDP
+        // exchange completes (see _handlePeerMessage). Reset on every
+        // fresh session so stale candidates from a previous attempt
+        // don't get applied to a new RTCPeerConnection.
+        this._pendingRemoteIce = [];
 
         // Acquire mic eagerly so the browser permission prompt appears now,
         // but tracks stay disabled (muted) until the user explicitly unmutes.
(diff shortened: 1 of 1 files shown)


tfrere:

PR #1101, opened 2026-05-07
Tighten Linux camera detection fallback

andimarafioti:
## Summary
- Keep `find_video_device()` deterministic over provided devices.
- Accept `device.path` when `api.v4l2.path` is unavailable on Linux V4L2 devices.
- Cover the fallback with a unit test.

## Validation
- `uv run --no-sync python -m pytest tests/unit_tests/test_device_detection.py -q`
- `uv run --no-sync python -m ruff check src/reachy_mini/media/device_detection.py`
- `uv run --no-sync python -m ruff format --check src/reachy_mini/media/device_detection.py`
- `uv run --no-sync python -m mypy src/reachy_mini/media/device_detection.py`

### DIFF
--- a/src/reachy_mini/media/device_detection.py
+++ b/src/reachy_mini/media/device_detection.py
@@ -251,8 +251,9 @@ def gst_devices_to_device_infos(gst_devices: Any) -> list[DeviceInfo]:
 def gst_monitor_devices(filter_class: str) -> list[DeviceInfo]:
     """Start a ``Gst.DeviceMonitor``, query devices, and return :class:`DeviceInfo` list.
 
-    This is meant to be called by the media server / audio classes as a
-    thin wrapper around the monitor lifecycle.
+    This is the only function in this module that actually imports and uses
+    GStreamer.  It is meant to be called by the media server / audio classes
+    as a thin wrapper around the monitor lifecycle.
 
     Args:
         filter_class: GStreamer device class filter, e.g.
@@ -272,25 +273,6 @@ def gst_monitor_devices(filter_class: str) -> list[DeviceInfo]:
         return gst_devices_to_device_infos(monitor.get_devices())
     finally:
         monitor.stop()
-
-
-def gst_v4l2_devices() -> list[DeviceInfo]:
-    """Query Linux V4L2 devices directly through GStreamer's provider.
-
-    ``Gst.DeviceMonitor`` can sometimes expose a matching camera without
-    the concrete ``/dev/videoN`` path needed by ``v4l2src``.  The V4L2
-    provider has a narrower scope and tends to expose ``device.path`` for
-    those cameras.
-    """
-    factory = Gst.DeviceProviderFactory.find("v4l2deviceprovider")
-    if factory is None:
-        return []
-
-    provider = factory.get()
-    if provider is None:
-        return []
-
-    return gst_devices_to_device_infos(provider.get_devices())
 
 
 def find_audio_device(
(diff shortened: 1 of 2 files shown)


algal:
I am also encountering this issue and so I would benefit from this PR making it into main.

Specifically:

- I see it on my Reachy Mini Lite, which is connected to Debian stable.
- OS/OpenCV sees the camera as 38fb:1002 SunplusIT Inc Reachy Mini Camera.
- HF stable daemon reachy-mini==1.7.3 logs "No camera found".
- HF testbench direct OpenCV camera check succeeds at 1920x1080@60.
- A local one-line fallback to device.path fixes the daemon detection path.

I'm running a patched version locally to work around this issue, which changes the lookup code on this line:

 https://github.com/pollen-robotics/reachy_mini/blob/790f5b628f6fcfef135a56659bc908a0433886c8/src/reachy_mini/media/device_detection.py#L446

The root cause appears to be a Linux/GStreamer device-property mismatch: on my Debian stable system the camera path is exposed as `device.path` rather than `api.v4l2.path`, so the daemon fails to detect a valid V4L2 camera.
 
My camera image is also very blurry,  but I'm guessing that is unrelated.

I'd be happy to help update this PR to track main, or to submit a PR with my fix, if either would be helpful.

src/reachy_mini/media/device_detection.py
```diff
@@ -386,6 +404,58 @@ def _make_camera_specs(cam_name: str) -> CameraSpecs:
     return ReachyMiniLiteCamSpecs()
 
 
+def _linux_video_path(props: dict[str, str]) -> str:
+    """Return a usable V4L2 path from GStreamer device properties."""
+    return props.get("api.v4l2.path") or props.get("device.path") or ""
+
+
+def _find_video_device(
```
FabienDanieau: Do we still need `_find_video_device` and `find_video_device` ? It seems that the later is only used by the unit tests.

src/reachy_mini/media/device_detection.py
```diff
@@ -386,6 +404,58 @@ def _make_camera_specs(cam_name: str) -> CameraSpecs:
...
+    for cam_name in cam_names:
+        for device in devices:
+            if cam_name not in device.display_name:
+                continue
+
+            props = device.properties
+
+            match current_platform:
+                case "Linux":
+                    device_path = _linux_video_path(props)
```
FabienDanieau: Why not applying the fallback here? Would it be possible to reduce the number of changes (see @algal comment)

src/reachy_mini/media/device_detection.py
```diff
@@ -486,9 +529,31 @@ def get_video_device() -> Tuple[str, Optional[CameraSpecs]]:
...
+
+    device_path, camera_specs = _find_video_device(
+        devices, current_platform, DEFAULT_CAM_NAMES
+    )
+    if camera_specs is not None:
+        return device_path, camera_specs
+
+    if current_platform == "Linux":
+        try:
+            device_path, camera_specs = _find_video_device(
```
FabienDanieau: Do we have to call again _find_video_device? see previous comment

FabienDanieau (changes requested):
Working on debian as well but can't reproduce the issue. Anyway the fix is great but I think the code can be reduced a bit.

src/reachy_mini/media/device_detection.py
andimarafioti: Done. I applied the fallback directly here instead of adding the extra provider path: Linux now uses `api.v4l2.path` first and falls back to `device.path`, with a focused regression test for that case.

src/reachy_mini/media/device_detection.py
andimarafioti: Good point. I simplified this in `0e02709c`: `_find_video_device` is gone, and `find_video_device()` is kept as the single pure matching function. `get_video_device()` now delegates to it again, so it is no longer only exercised by unit tests.

src/reachy_mini/media/device_detection.py
andimarafioti: Removed this extra pass as suggested. `get_video_device()` is back to the simple monitor + `find_video_device()` flow, and the PR now only adds the direct `device.path` fallback.

FabienDanieau (approved):
Still works with Lite (debian) and Wireless. Thanks!

FabienDanieau:
@algal does that fix your issue?

PR #1117, opened 2026-05-12
feat(sdk): debounce ICE blips + emit iceStateChange / networkOnline / networkOffline

tfrere:
## Why

Routine mobile/laptop network noise (Wi-Fi blip, AP roam, brief screen-off, BT route change on iOS) was tearing down WebRTC sessions even when the underlying ICE link recovered a moment later. The previous handler reacted on the FIRST transition into `disconnected` / `failed`, firing `error` (and rejecting the session promise) before the spec-defined recovery window had a chance to play out.

Concrete consumer-side symptoms today:
- Mobile shell: full-screen *"session ended unexpectedly"* on a 2 s Wi-Fi flicker.
- Embedded Spaces / dev consoles: red banner during a normal rapid app-switch on iOS.
- The mobile app's own OpenAI bridge already debounces ICE blips on its peer (5 s grace). The robot peer didn't, so the AI side survived while Reachy dropped.

This PR rewrites the SDK side using the same pattern the OpenAI bridge uses, and adds two public events that let every consumer (mobile shell, iframed Spaces, standalone Spaces) react to transport changes without re-implementing the plumbing.

### 1. `oniceconnectionstatechange` is debounced

| State | Behaviour |
|---|---|
| `connected` / `completed` | Clears any pending grace, marks ICE up, proceeds with session-ready check (unchanged). |
| `disconnected` | Schedules a 3 s grace. If the tab is hidden, defers the grace to the next ` visibilitychange` (background timers are throttled, so a foreground timer would fire unpredictably). On grace expiry: if still `disconnected`, emit `error` with the same payload shape as before. |
| `failed` | Schedules a 1 s grace before rejecting the session promise + emitting `error`. We've observed real `failed → connected` transitions on rapid AP roams and BT route changes; 1 s is enough to absorb those without delaying a real failure noticeably. |

Existing consumers listening for `error` keep working.

### 2. New event \`iceStateChange\`

Fires on every \`iceConnectionState\` transition with \`{ state: RTCIceConnectionState }\`. Lets consumers render finer-grained UX (e.g. a transient *"Reconnecting…"* badge during \`disconnected\`) without attaching their own handler to \`_pc\`.

### 3. New events \`networkOnline\` / \`networkOffline\`

Forwarded from \`window.online\` / \`window.offline\` and (on browsers that ship it: Chrome, Android WebView) \`navigator.connection.change\`. Listeners are scoped to the live session - installed in \`startSession()\`, removed in \`stopSession()\` / \`disconnect()\`. Lets the mobile shell / iframed Spaces freeze motion + show a badge on offline and probe the data channel on online, without each consumer wiring its own browser listeners.

### 4. Symmetric cleanup

\`stopSession()\` and \`disconnect()\` both call \`_clearIceGrace()\` and \`_uninstallNetworkListeners()\` BEFORE closing \`_pc\`, so a queued grace callback can't dereference a dead peer connection a couple of ticks later.

## Non-goals (deferred)

This SDK is the SDP **answerer** (the daemon's GStreamer \`webrtcsink\` is the offerer). A true end-to-end ICE *restart* therefore requires the daemon to emit a new offer with \`iceRestart=true\`, which means a new signaling-protocol message + GStreamer-side wiring on \`reachy_mini_central\` and the daemon. That's a separate follow-up; **this PR is the client-side debounce + observability layer it will eventually plug into**.

Also out of scope: TURN servers (no STUN-only configuration changed here), full PC recreation as a final fallback.

## Replaces

Supersedes the abandoned WIP commit [\`5e35bec2\`](https://github.com/pollen-robotics/reachy_mini/commit/5e35bec2) on the \`test---sdh-improvement\` branch, which only stripped the \`disconnected\` error event without adding the grace window or the visibility-aware deferral.

## Test plan

- [ ] Healthy session: connect, talk, stop. \`iceStateChange\` visible in DevTools (\`checking\` → \`connected\`). No regression.
- [ ] Wi-Fi flicker (toggle AP off ~2 s): session survives, \`iceStateChange\` shows \`disconnected → connected\`, no \`error\`.
- [ ] Airplane mode 5 s then off: \`networkOffline\` fires, then \`networkOnline\`, ICE recovers, session survives.
- [ ] Hard network kill (10 s+): grace expires, \`error\` is emitted with the spec-compliant payload. Session can be restarted cleanly.
- [ ] iOS Safari WKWebView: lock screen 30 s during a session, unlock. Visibility-aware deferral fires correctly on return.
- [ ] Mobile-app vendored SDK swap: re-vendor \`vendor/reachy-mini.js\` once merged.

### DIFF
--- a/js/reachy-mini.js
+++ b/js/reachy-mini.js
@@ -221,6 +221,22 @@ const ICE_DISCONNECT_GRACE_MS = 3000;
  */
 const ICE_FAILED_GRACE_MS = 1000;
 
+/**
+ * Ceiling on how long we'll keep `_armIceGraceOnVisibility` waiting
+ * for the tab to come back. The daemon's `webrtcsink` runs a STUN
+ * consent-freshness check (RFC 7675, ~30 s default) and unilaterally
+ * tears its side of the session down past that window, releasing the
+ * producer slot on central. If the user backgrounded the tab for
+ * longer than this, running another 3 s foreground grace is a lie —
+ * the underlying transport is gone, nothing can recover. Give up
+ * straight away so the host shows the real "session expired" UX
+ * instead of a fake "Reconnecting…" badge that's never going to
+ * heal. 60 s gives a 2× margin over the daemon-side timeout — long
+ * enough to absorb a "phone in pocket for 45 s" case, short enough
+ * to be honest with the user.
+ */
+const MAX_VISIBILITY_DEFER_MS = 60_000;
+
 // ─── Internal helpers ────────────────────────────────────────────────────────
 
 /** Clamp a volume to [0, 100] and round to integer — mirrors the server-side
@@ -1366,6 +1382,7 @@ export class ReachyMini extends EventTarget {
      */
     _armIceGraceOnVisibility() {
         if (this._pendingVisibilityHandler) return;
+        const deferredAt = Date.now();
         const handler = () => {
             if (typeof document !== 'undefined' && document.hidden) return;
             document.removeEventListener('visibilitychange', handler);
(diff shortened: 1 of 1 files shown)


FabienDanieau (approved):
Ok! please reduce comments as much as possible :)

cdussieux:
solved by https://github.com/pollen-robotics/reachy_mini/pull/1183

PR #1133, opened 2026-05-20
Feature/daemon side move upload

RemiFabre:
(Working version but still WIP)


RemiFabre:
## What this adds

Ten new DataChannel/WS commands so apps can ship a recorded move (and
optionally an audio track) to the daemon and play them server-side,
instead of streaming `set_target` one frame at a time over the
network. All commands are fire-and-forget at the transport level (the
long-standing daemon convention); the playback task surfaces its
progress through broadcast messages keyed by upload_id.

| Command | Purpose |
|---|---|
| `upload_move_start`   | open a move slot, declare chunk count |
| `upload_move_chunk`   | send a fragment of the move JSON |
| `upload_move_finish`  | assemble + parse on the daemon |
| `upload_audio_start`  | open an audio slot keyed by the same upload_id |
| `upload_audio_chunk`  | send a fragment of base64-encoded WAV bytes |
| `upload_audio_finish` | decode + write to `/tmp/marionette-audio/{id}.wav` |
| `play_uploaded_move`  | kick off `Backend.play_move` on the slot (plays motion + audio in lockstep when both are present) |
| `cancel_move`         | flip `_move_cancelled` so the play loop exits |
| `play_uploaded_audio` | play the uploaded audio standalone (used by the app's recording flow) |
| `cancel_audio`        | stop_sound() on the media server |

## Why

**Motion**: streamed `set_target` is jittery on wireless robots because every frame is a WebRTC round-trip. Uploading the move once and letting `Backend.play_move` tick at 100 Hz locally on the CM4 removes the network from the playback inner loop.

**Audio**: streaming audio from the browser over a WebRTC audio sender works, but has variable pipeline latency (encoder warmup, jitter buffer adaptation, codec resample) that differs across platforms and even across recordings on the same platform. Uploading the audio and letting the daemon play both motion AND audio gives single-clock sync — no cross-network drift between browser AudioContext and robot speaker.

The standalone `play_uploaded_audio` exists so apps that record motion against a playing track can use the SAME audio pipeline at record time and play time. The pipeline-latency constant is then identical in both, so a single per-robot offset is enough to fine-tune sync.


## Wire shape

Client to server (all fire-and-forget):
```
upload_move_start  {upload_id, total_chunks, encoding?, description?, estimated_duration_s?}
upload_move_chunk  {upload_id, chunk_index, chunk}    times N
upload_move_finish {upload_id}

upload_audio_start  {upload_id, total_chunks, encoding?, description?}
upload_audio_chunk  {upload_id, chunk_index, chunk}   times N
upload_audio_finish {upload_id}

play_uploaded_move  {upload_id, play_frequency, initial_goto_duration, audio_lead_ms}
cancel_move         {}

play_uploaded_audio {upload_id}
cancel_audio        {}
```

`encoding` on `upload_move_start` is `"json"` (default, backwards compatible) or `"gzip+base64"`. The latter compresses recorded-move JSON by roughly 3-15x on the wire (lots of repeated float patterns), at the cost of a small base64 / gzip step on both sides. Recommended for any move over a few seconds long.

`encoding` on `upload_audio_start` is `"wav-base64"` (the only option today): raw WAV bytes sliced into chunks and base64-encoded. PCM doesn't compress well so we skip gzip; the daemon decodes back to bytes and writes the WAV to a temp file.

`audio_lead_ms` on `play_uploaded_move` (-2000..+2000):
- `> 0`: audio plays N ms BEFORE motion (compensates for slow motor pickup).
- `< 0`: audio plays N ms AFTER motion (compensates for slow GStreamer playbin warmup; this is the regime on the wireless robot, default `-100` in the Marionette app).
- `= 0`: both fire simultaneously.

Server to client (broadcast on every transport, filter by `upload_id`):
```
{type: "play_uploaded_move", upload_id, started: true, duration_s, has_audio}
{type: "play_uploaded_move", upload_id, finished: true,  has_audio}
{type: "play_uploaded_move", upload_id, cancelled: true, has_audio}
{type: "play_uploaded_move", upload_id, error: "...",    has_audio}

{type: "play_uploaded_audio", upload_id, started: true}
{type: "play_uploaded_audio", upload_id, error: "..."}
```

If a slot failed to assemble (chunk out of order, parse error), the first `play_uploaded_move` / `play_uploaded_audio` after the failure broadcasts `{error: "no such uploaded move/audio ..."}`.

`play_uploaded_audio` does NOT emit a finished event; callers know the duration from the WAV header and send `cancel_audio` when they're done (e.g. recording stopped).

## Out-of-scope (deliberately)

- No pause/resume of motion. Stop only; restart from frame 0.
- No streaming playback during upload. Full upload, then start.
- No on-the-fly audio decode for non-WAV formats. Audio must be canonical 16 kHz mono PCM WAV; clients normalize before upload.
- No persistence. Uploaded moves live in RAM; audio temp files live in `/tmp` and get evicted at play-end or TTL.

## Tested

A bunch of manual tests using the marionette-experimental repo both on wireless and lite.

E2E against a wireless robot over raw WS (`tests/scripts/wireless_e2e.py` in the marionette-experimental repo):

| Test | Wire bytes | Upload time |
|---|---|---|
| intro 3 s, json                  |  85 KB         |  3 ms |
| 30 s sweep, json                 | 403 KB         |  4 ms |
| 30 s sweep, gzip+base64          |  27 KB (7%)    |  2 ms |
| 3-min sweep, gzip+base64         | 169 KB (7%)    |  7 ms |
| intro + Seven Nation Army WAV (4.8 MB) | 6.4 MB audio b64 | 413 ms |

| Other timings | |
|---|---|
| Time from `play_uploaded_move` sent to `started` event | 30-300 ms (depends on parse cost) |
| Cancel mid-play, time to final event | ~50 ms |

Upload is bandwidth-bound now (previously RTT-bound, 65 ms / 238 ms for the un-pipelined chunk path). Over WebRTC through a central signaling server the win is much larger because per-chunk RTT used to dominate.

## Backwards compatibility

All additions; nothing removed or changed semantically in existing commands. Older clients that don't know the new commands are unaffected. Older clients that receive a stray `play_uploaded_move` / `play_uploaded_audio` broadcast safely ignore it (validate against `AnyServerMsg` which doesn't list the new types).

`Backend.play_move` gains an `audio_lead_s` parameter with default 0; existing callers (REST `/play/recorded-move-dataset/...`, Python SDK) keep the previous behavior bit-for-bit.

RemiFabre:
## JS SDK consumer API

To keep apps from re-deriving chunking / encoding / backpressure / broadcast routing, the same branch adds five typed methods to `js/reachy-mini.js`:

| Method | Returns | Use |
|---|---|---|
| `playMove(motion, opts)` | `Promise<{finished?, cancelled?, error?, has_audio?}>` | Upload motion (+ optional WAV) and play; resolves when playback ends |
| `cancelMove()` | `boolean` | Stop an in-flight `playMove` |
| `uploadAudio(blob, opts)` | `Promise<string>` | Standalone audio upload, returns `uploadId` |
| `playUploadedAudio(uploadId, opts)` | `Promise<{started: true, ...}>` | Trigger daemon-side audio playback; resolves on `started` broadcast (record-time sync anchor) |
| `cancelAudio()` | `boolean` | Stop an in-flight `playUploadedAudio` |

The SDK internally handles 12 KB chunking, `gzip+base64` for motion JSON, `wav-base64` for audio, backpressure pacing on `_dc.bufferedAmount`, and broadcast routing keyed by `upload_id`. Default `audioLeadMs` is `-100` — the empirical system-wide constant covering combined motor + GStreamer playbin warmup.

Docs: new "Daemon-side recorded-move playback" section in `docs/source/SDK/javascript-sdk.md`; pointer paragraph in `AGENTS.md`.

Validation: marionette-experimental was refactored on top of this API (`lib/move-upload.js` deleted, −725 lines) and tested end-to-end on the wireless robot — full record / preview / save / play / cancel flow works against the daemon at this branch tip.

cdussieux (approved):
added cancel and removed marionnete name

lgtm, solid ground for future cool apps

PR #1138, opened 2026-05-21
1136 enable motors should never generate a brutal motion

### LINKED ISSUE #1136: enable_motors() should never generate a brutal motion
If we disable_motors(), move the robot, then enable_motors(), the motors will try to go back to their old goal position. Instead we should set the goal position to the current present position (this scheme is already used when we change modes)

RemiFabre:
If we disable_motors(), move the robot, then enable_motors(), the motors will try to go back to their old goal position. Instead we set the goal position to the current present position (this scheme is already used when we change modes, see `set_head_operation_mode` / `set_antennas_operation_mode`).

While testing this I saw a subtle thing: if we only update `target_head_joint_positions`, the next IK tick will overwrite it with joints derived from the (outdated) `target_head_pose`, and the head will still brutally jump there. So I also update `target_head_pose` using the latest `current_head_pose` available (it's already FK'd at every control-loop tick, so no extra computation).
=> Btw I believe the same problem could happen when changing modes, it's just that it's not visible because in gravity compensation the head is already being held at `target_head_pose`, so when IK runs after the mode switch it produces joints equal to present and the snap is hidden.

This was tested in 2 ways. First in python using this:
```
Without the fix, step 4 snaps it back toward the previous goal.
"""

import time

from reachy_mini.utils import create_head_pose

with ReachyMini(media_backend="no_media") as mini:
    print("Going to base pose…")
    mini.goto_target(create_head_pose(), antennas=[0.0, 0.0], duration=1.0)
    time.sleep(1.2)

    print("Disabling torque — move the head by hand (2 s)…")
    mini.disable_motors()
    time.sleep(2.0)

    print("Re-enabling torque (must not move!)…")
    mini.enable_motors()
    time.sleep(1.0)

    print("Going to base pose…")
    mini.goto_target(create_head_pose(), antennas=[0.0, 0.0], duration=1.0)
    time.sleep(1.2)

    print("Done.")
```

Then in JS, using this:
https://huggingface.co/spaces/RemiFabre/reachy-mini-js-practices

In both cases, without this fix the head should jump very rapidly when you torque_off -> torque_on ("brutal torque ON" in the JS app). With the fix, this should never happen.

Also added unit tests that mock the interfaces, these fail without the fix and pass with it.


pierre-rouanet (approved):
LGTM, but maybe add a comment in the setter to explain that the target can be removed by enabling the torque. Before you could first set the target and then enable torque. Probably not used anywhere but just in case.
Could also be worth adding a line somewhere in agent files. In my experience, it does not really understand how compliancy and mode works.

RemiFabre:
> Before you could first set the target and then enable torque.

What if we added an automatic hook that calls the police when someone does this? We might have found a clever way to identify CRIMINALS early

RemiFabre:
> LGTM, but maybe add a comment in the setter to explain that the target can be removed by enabling the torque. Before you could first set the target and then enable torque. Probably not used anywhere but just in case. Could also be worth adding a line somewhere in agent files. In my experience, it does not really understand how compliancy and mode works.

done. Merging as soon as the CI passes

PR #1142, opened 2026-05-22
feat(media): detect and report stuck WebRTC sessions via watchdog

tfrere:
## Summary

Daemon-side fix for the historical "stuck at session" UX bug, where libnice (used by `webrtcbin` internally) freezes mid-`CHECKING` and the JS client spins forever waiting on an offer that will never complete. The daemon used to be silent in that scenario; this PR makes it observe its own `webrtcbin` state and translate stuck negotiations into a typed `endSession` so the client can render a real error view instead of a blank screen.

## Watchdog flow

1. `GstMediaServer._consumer_added` subscribes to webrtcbin's `notify::ice-connection-state`, `notify::connection-state`, and `notify::signaling-state` for each peer, and starts a `GLib.timeout_add_seconds` deadline (12s).
2. `connection-state == failed` short-circuits the deadline and reports `peer_connection_failed` immediately so hard network failures get a fast rejection.
3. The deadline reports `ice_negotiation_timeout` if the peer is still negotiating after 12s (the canonical libnice-frozen symptom).
4. `CentralSignalingRelay.notify_peer_session_failed` is the thread-safe entry point (uses `asyncio.run_coroutine_threadsafe` because the GStreamer signal callbacks fire on internal threads while the relay lives on its own asyncio loop). It translates the failure into an `endSession` message routed at the central `sessionId`, then clears the per-session bookkeeping.
5. The mapping `media_server peer_id` -> `central_session_id` reuses the existing `_local_to_central_session` dict; no new wire shape on the central side.

## gst-plugins-rs bump (0.14.1 -> 0.14.5)

The upstream `gst-plugins-rs` [CHANGELOG](https://github.com/GStreamer/gst-plugins-rs/blob/main/CHANGELOG.md) explicitly lists between 0.14.1 and 0.14.5:

- `webrtcsink: Fix deadlock when recalculating latency.`
- `webrtcsink: Fix deadlock between remote description and ICE handling.` <- **this is the root cause this watchdog is designed to surface**
- `webrtc: Various minor fixes and improvements.`

I bumped the install guide accordingly. The watchdog remains the safety net (libnice itself can still misbehave at the OS-package level on some Debian builds), but pinning to 0.14.5 means it should rarely fire in practice.

## Files changed

| File | Why |
|------|-----|
| `src/reachy_mini/media/media_server.py` | Watchdog state, signal handlers, deadline timer, `set_session_failed_handler` setter |
| `src/reachy_mini/media/central_signaling_relay.py` | Public thread-safe `notify_peer_session_failed` + inner coroutine that emits `endSession` to central |
| `src/reachy_mini/daemon/daemon.py` | Wires `media_server.set_session_failed_handler(relay.notify_peer_session_failed)` after the relay starts |
| `docs/source/SDK/gstreamer-installation.md` | Bumps the documented `gst-plugins-rs` pin from `0.14.1` to `0.14.5` |
| `tests/unit_tests/test_media_server_watchdog.py` | 19 new unit tests covering state tracking, deadline behaviour, eager fail-fast, idempotent teardown, no-double-fire |
| `tests/unit_tests/test_central_signaling_relay.py` | 7 new unit tests covering `notify_peer_session_failed` wire shape, bookkeeping cleanup, no-op on unmapped peers, cross-thread scheduling |

## Test plan

- [x] `pytest tests/unit_tests/test_media_server_watchdog.py tests/unit_tests/test_central_signaling_relay.py` -> 49 passed locally
- [x] `ruff check` and `ruff format --check` clean on touched files
- [ ] CI green (pytest + lint workflows)
- [ ] Manual smoke test on a real Reachy Mini Wireless: trigger a WebRTC connect over a flaky network, observe `[watchdog]` log lines + `endSession` reaching the JS client within 12s
- [ ] Manual smoke test on Reachy Mini Lite: same as above; confirms the wiring also fires under the desktop tray daemon

## Out of scope (follow-ups)

- JS-side handling of the new `reason` strings (`ice_negotiation_timeout`, `peer_connection_failed`) so the user-facing message can be more specific than "session failed". The wire shape is already there; the JS SDK just needs to switch on it.
- Bumping libnice itself: depends on the OS image (Bookworm ships 0.1.21, latest upstream 0.1.22). The 0.14.5 plugin bump is the actionable lever from this repo.
- Integration test that exercises a real `webrtcbin` reaching `connection-state == failed`. Needs a mock WebRTC peer; tracked separately along with the same gap on `central_signaling_relay`.

### DIFF
--- a/docs/source/SDK/gstreamer-installation.md
+++ b/docs/source/SDK/gstreamer-installation.md
@@ -75,7 +75,15 @@ To build and install the WebRTC plugin, run the following commands:
 # Clone the GStreamer Rust plugins repository
 git clone https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs.git
 cd gst-plugins-rs
-git checkout 0.14.1
+# 0.14.5 ships a critical webrtcsink fix for a deadlock between
+# remote description and ICE handling that manifested on the
+# daemon as the historical "stuck at session" UX (the JS client
+# spinning forever because libnice never reached `connected`).
+# Older 0.14.x tags (0.14.1, 0.14.2, 0.14.3) do NOT carry this
+# fix; see the upstream CHANGELOG. The daemon-side negotiation
+# watchdog (`reachy_mini.media.media_server`) is the safety net,
+# but pinning to a fixed plugin means that net should rarely fire.
+git checkout 0.14.5
 
 # Install the cargo-c build tool
 cargo install cargo-c
--- a/src/reachy_mini/daemon/daemon.py
+++ b/src/reachy_mini/daemon/daemon.py
@@ -187,13 +187,24 @@ class Daemon:
             transport = "wifi" if self.wireless_version else "usb"
 
             self.logger.info("Starting central signaling relay...")
-            await start_central_relay(
+            relay = await start_central_relay(
                 hf_token=hf_token,
                 robot_name=self.robot_name,
                 transport=transport,
                 robot_app_lock=self.robot_app_lock,
             )
             self.logger.info("Central signaling relay started")
+
+            # Wire the negotiation watchdog: when GstMediaServer
+            # detects a stuck WebRTC negotiation (typically libnice
+            # frozen mid-CHECKING), it forwards the failure to central
+            # via the relay so the JS client gets a typed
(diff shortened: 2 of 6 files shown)


cdussieux (approved):
lgtm

PR #1148, opened 2026-05-25
js: stop calling getUserMedia in the SDK

### LINKED ISSUE #1147: ease microphone access from the js lib
Remove the need to ask for the permission of the browser mic.

FabienDanieau:
The SDK eagerly called navigator.mediaDevices.getUserMedia({audio: true}) inside startSession, popping the browser's microphone permission prompt on every connection. The track was then added to the WebRTC peer connection muted, used solely to satisfy sendrecv negotiation so apps could later replaceTrack() their own audio onto the sender.

This is the wrong default. The SDK's job is transport; audio source policy belongs to the app. Apps that want to send actual audio to the robot speaker — TTS, prerecorded files, or the user's mic for teleop — already do so explicitly via sender.replaceTrack() on the audio sender from _pc.getSenders(). Forcing every connection to ask for mic permission, even when no user mic is involved, was friction for no gain.

Always set up the silent oscillator → MediaStreamDestination placeholder (previously only reached as the getUserMedia-denied fallback). WebRTC still negotiates a sendrecv audio sender, replaceTrack still works, the permission prompt is gone.

enableMicrophone is marked @deprecated in the d.ts and kept parsed in the constructor for backward compatibility, but no longer has any effect.

Existing teleop apps that relied on the SDK acquiring the user's mic must now do so themselves (getUserMedia + replaceTrack onto the audio sender) — a five-line addition mirroring what TTS apps already do.

Assisted-by: Claude:claude-opus-4-7

Can be tested with : https://huggingface.co/spaces/pollen-robotics/bedtime-story-generator
You'll need to use reachy-mini-sdk.js locally (the one of this PR).
Then `npm run dev`, and `http://localhost:5173/#hf_token=hf_xxx&hf_username=xxx&hf_token_expires=2099-01-01T00:00:00Z`

### DIFF
--- a/js/reachy-mini-sdk.d.ts
+++ b/js/reachy-mini-sdk.d.ts
@@ -60,6 +60,13 @@ export interface RobotState {
 /** SDK constructor options. */
 export interface ReachyMiniOptions {
   signalingUrl?: string;
+  /**
+   * @deprecated The SDK no longer acquires the user's microphone. A silent
+   * placeholder audio sender is always set up so apps can `replaceTrack`
+   * their own audio (TTS, files, or — for teleop — the user's mic) onto
+   * it. This option is parsed for backward compatibility but has no
+   * effect.
+   */
   enableMicrophone?: boolean;
   clientId?: string;
   appName?: string;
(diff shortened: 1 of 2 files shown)


cdussieux (approved):
lgtm

PR #1162, opened 2026-05-28
feat(audio): software AEC fallback when no XVF3800 board is present

### LINKED ISSUE #1161: Daemon audio: no echo cancellation when XVF3800 is absent (sim, no-board, dev setup)
## Symptom

Any client that uses the daemon's WebRTC audio duplex (mobile app, sdk-js demos, conversation app's WebRTC variant) ends up in a feedback loop when the daemon runs without an XVF3800 audio board: the speaker plays the assistant's voice, the mic recaptures it raw, and the loop is sent back as user input. The model interrupts itself and the conversation collapses.

Reproducible from the mobile app talking to a daemon launched with `--sim` or `--mockup-sim`, or with a real Reachy Mini Lite where the USB audio dongle isn't plugged.

## Root cause

`media/audio_gstreamer.py` only routes through hardware AEC when the wireless `.asoundrc` is present:

```python
if has_reachymini_asoundrc():
    # Wireless CM4: use the preconfigured .asoundrc ALSA devices
    # which route through the XMOS AEC loopback properly.
    audiosrc = Gst.ElementFactory.make("alsasrc")
    audiosrc.set_property("device", "reachymini_audio_src")
```

When this branch is not taken, both source and sink fall through to the platform default (`osxaudiosrc`/`pulsesrc`/`wasapi2src`/`autoaudiosrc` paired with the matching default sinks). The mic and speaker are then completely independent OS endpoints with no AEC reference between them.

There is no software AEC anywhere downstream:
- aiortc receives raw F32LE PCM from the appsink and forwards as-is.
- Browser-side WebRTC AEC doesn't apply because the daemon-published track is a remote track from the browser's POV (not a `getUserMedia` capture).
- `applyAudioConfig` (XVF3800 PP_AGC, PP_MIN_NS, PP_GAMMA_E, ...) targets the XMOS chip on the audio board and silently no-ops when the board isn't present.

So all production echo cancellation lives on the XVF3800 hardware. The "no board" path has none.

## Proposed fix

Insert a software AEC stage in the GStreamer pipelines when no XVF3800/`.asoundrc` is detected. GStreamer ships `webrtcdsp` + `webrtcechoprobe` (from `gst-plugins-bad`, based on WebRTC AEC3 and already an indirect dep of the project): the playback branch installs the `webrtcechoprobe` as the AEC reference, the recording branch routes through `webrtcdsp` referencing the same probe.

Sketch:

```
record:   audiosrc → queue → audioconvert → audioresample → webrtcdsp(probe=p) → appsink
playback: appsrc → audioconvert → audioresample → tee ─┬→ audiosink
                                                       └→ webrtcechoprobe name=p (fakesink)
```

Both pipelines must share a single `Gst.Pipeline` (or use a shared bin) for `webrtcdsp` to find the probe by name. `webrtcdsp` also handles noise suppression and AGC, which mirrors what the XVF3800 does, so the no-board path lands close to the on-board behaviour.

Detection logic (e.g. in `_init_pipeline_record`/`_init_pipeline_playback`): if `has_reachymini_asoundrc()` is false AND `get_audio_device(\"Source\")` returns no ReSpeaker card → enable the software AEC stage. Real Lite boards keep the current zero-AEC path because their on-board AEC is still active over USB.

## Alternatives considered

- UX-only banner in the mobile app suggesting headphones in sim → cheap mitigation, doesn't fix sdk-js or other clients.
- Mobile-side: reroute OpenAI playback to the phone speaker only when the daemon reports no audio board → changes the semantic of "the robot speaks" in sim and still leaves Lite-without-dongle broken.

## Related

- Same root cause as `pollen-robotics/reachy_mini_conversation_app#221` (which scopes its fix to the Gradio/JS UI split, not the AEC itself).
- Mobile app symptom tracker: `pollen-robotics/reachy_mini_mobile_app#46` (closed in favour of this one).

tfrere:
## Why

Closes #1161.

Without a Reachy Mini Wireless (XMOS via `.asoundrc`) or a ReSpeaker USB dongle (XVF3800), the GStreamer audio backend routes the platform default mic and speaker straight through with no echo cancellation. Any conversation client (mobile app, JS SDK, `reachy_mini_conversation_app`) that pushes the realtime audio unprocessed then hears its own voice back through the laptop mic and the loop collapses on feedback within a couple of turns.

This is the daemon-side companion fix referenced from `reachy_mini_conversation_app#46` (closed → redirected to this repo): putting the AEC in the audio backend means every client benefits without having to roll its own software AEC.

## What changed

A `webrtcdsp` + `webrtcechoprobe` pair (from `gst-plugins-bad`) is spliced on top of the existing pipelines whenever no hardware AEC is detected in the audio path.

**Record pipeline** (added between `audioresample` and the appsink):

```
... → audioresample → audioconvert → capsfilter(S16LE) → audiobuffersplit(10 ms)
                    → webrtcdsp(probe=reachymini_aec_probe,
                                delay-agnostic=true, extended-filter=true)
                    → audioconvert → appsink(F32LE)
```

**Playback pipeline** (parallel tee branch alongside the existing speaker + wobbler branches, only on the appsrc-based conversation pipeline):

```
                                                                   ┌→ queue → ac → ar → audiosink
                                                                   │
appsrc(F32LE) → tee ──────────────────────────────────────────────┼→ queue → ac → ar → appsink_wobbler
                                                                   │
                                                                   └→ queue → audioconvert → capsfilter(S16LE)
                                                                            → audiobuffersplit(10 ms)
                                                                            → webrtcechoprobe(name=reachymini_aec_probe)
                                                                            → fakesink(sync=false, async=false)
```

**Detection** is conservative — the stage is added only when there is **no** hardware AEC and `webrtcdsp` / `webrtcechoprobe` are actually packaged:

| Condition | SW AEC |
|---|---|
| `has_reachymini_asoundrc()` (Wireless XMOS loopback) | OFF |
| `get_audio_device(\"Source\") is not None` (ReSpeaker USB dongle) | OFF |
| `webrtcdsp` / `webrtcechoprobe` missing from the build | OFF (loud warning) |
| Otherwise | **ON** |

**Env-var overrides** (for ops + A/B testing):

* `REACHY_MINI_DISABLE_SW_AEC=1` force-disable.
* `REACHY_MINI_FORCE_SW_AEC=1` force-enable on hardware setups (compare SW vs HW AEC).

## Why this is small

* No `_build_audiosink_tee_bin` change → `play_sound` paths are untouched. Beeps are short and would otherwise collide on the probe's process-wide name.
* No new clock-domain plumbing → `delay-agnostic=true` + `extended-filter=true` handle the residual drift between the record and playback pipelines (they each pick up the platform mic / speaker clock).
* Backward compatible by default on hardware-equipped robots: `.asoundrc` and XVF3800 paths short-circuit the heuristic so the patch is a no-op there.

## Test plan

- [x] Module imports + lint clean (`ruff check`, `ruff format`).
- [x] Smoke test on macOS dev box: pipelines reach `PAUSED` in **disabled**, **forced**, and **auto** modes (auto correctly resolves to ON without hardware).
- [x] 14/14 unit tests in `tests/unit_tests/test_audio_gstreamer.py` pass:
  - 3 pre-existing `_compute_pts` cases.
  - 11 new env-var / auto-detection cases covering `DISABLE`, `FORCE`, asoundrc, ReSpeaker, missing-plugins, and the happy paths.
- [x] 6/7 `MediaBackend.LOCAL` tests in `test_audio.py` pass with SW AEC active (the 7th — `test_DoA` — needs a physical ReSpeaker and is pre-existing).
- [ ] **Needs hardware validation**: end-to-end conversation loop on
  - a Lite + USB ReSpeaker (should stay on HW AEC, no behavioural change),
  - a Wireless robot (should stay on HW AEC, no behavioural change),
  - a Lite **without** ReSpeaker / `--sim` / dev workstation (the case this PR fixes — verify the OpenAI Realtime / mobile-app conversation no longer collapses on feedback within the first turn).

## Caveat documented inline

`audiobuffersplit.output-buffer-duration` is a `GstFraction`. PyGObject's `set_property(\"output-buffer-duration\", Gst.Fraction(1, 100))` segfaults on the GStreamer 1.28 builds we ship; the fix routes through `Gst.util_set_object_arg(bsplit, \"output-buffer-duration\", \"1/100\")` which constructs the GValue C-side. Both call sites carry a comment explaining it so the next person doesn't \"clean it up\".

### DIFF
--- a/src/reachy_mini/media/audio_gstreamer.py
+++ b/src/reachy_mini/media/audio_gstreamer.py
@@ -25,6 +25,34 @@ Platform audio sources / sinks are discovered at runtime:
 The "Reachy Mini Audio" card is located by name via ``Gst.DeviceMonitor``.
 If no matching card is found the platform default is used instead.
 
+Software acoustic echo cancellation (no-board fallback)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When no XVF3800 audio board is in the audio path (no Wireless
+``~/.asoundrc`` AND no ReSpeaker USB dongle detected), the backend
+falls back to the platform default mic + speaker. Without the
+hardware AEC of the XMOS chip the OpenAI Realtime / conversation
+loop hears its own voice through the laptop speakers and collapses
+on feedback.
+
+To keep the simulation / dev workflow usable, this backend inserts
+a ``webrtcdsp`` + ``webrtcechoprobe`` pair (from
+``gst-plugins-bad``) on top of the record / playback pipelines in
+that case. Detection is automatic:
+
+* If ``has_reachymini_asoundrc()`` is true → no extra processing
+  (Wireless XMOS handles AEC).
+* If ``get_audio_device("Source")`` finds a ReSpeaker card → no
+  extra processing (USB XVF3800 handles AEC).
+* Otherwise → enable the software AEC stage.
+
+Two environment variables can be set to override the heuristic, for
+example to benchmark or troubleshoot:
+
+* ``REACHY_MINI_DISABLE_SW_AEC=1`` force-disables the fallback.
+* ``REACHY_MINI_FORCE_SW_AEC=1`` force-enables it (e.g. to compare
+  output against the hardware XVF3800 AEC on a board-equipped robot).
+
 Note:
     This class is typically used internally by ``MediaManager`` when the
     ``LOCAL`` backend is selected.  Direct usage is possible but usually
@@ -96,6 +124,12 @@ class GStreamerAudio(AudioBase):
     PLAYBACK_SINK_BUFFER_TIME_US = 50_000
     PLAYBACK_SINK_LATENCY_TIME_US = 5_000
 
+    # Shared name that ties the recording-side `webrtcdsp` to the
+    # playback-side `webrtcechoprobe`. Both elements communicate via
+    # a process-wide registry indexed by this string, so the value
+    # just needs to be stable and unique within the process.
+    _AEC_PROBE_NAME = "reachymini_aec_probe"
+
     def __init__(self, log_level: str = "INFO") -> None:
         """Initialize recording and playback pipelines.
 
@@ -114,6 +148,8 @@ class GStreamerAudio(AudioBase):
         self._thread_bus_calls = Thread(target=lambda: self._loop.run(), daemon=True)
         self._thread_bus_calls.start()
 
+        self._sw_aec_enabled = self._resolve_sw_aec_enabled()
+
         self._pipeline_record = Gst.Pipeline.new("audio_recorder")
         self._init_pipeline_record(self._pipeline_record)
         self._bus_record = self._pipeline_record.get_bus()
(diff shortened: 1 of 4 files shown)


src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -25,6 +25,34 @@
...
+* If ``has_reachymini_asoundrc()`` is true → no extra processing
+  (Wireless XMOS handles AEC).
+* If ``get_audio_device("Source")`` finds a ReSpeaker card → no
+  extra processing (USB XVF3800 handles AEC).
+* Otherwise → enable the software AEC stage.
+
+Two environment variables can be set to override the heuristic, for
+example to benchmark or troubleshoot:
+
+* ``REACHY_MINI_DISABLE_SW_AEC=1`` force-disables the fallback.
```
FabienDanieau: I would detect the use of the simulation based on https://github.com/pollen-robotics/reachy_mini/blob/c9549bb1fc7828aefb90d0703492d39443b4317b/src/reachy_mini/daemon/app/main.py#L483 rather than using an env variable

FabienDanieau: `--sim `as well

src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -129,6 +165,197 @@ def __init__(self, log_level: str = "INFO") -> None:
...
+        anything else falls back to the auto-detection heuristic.
+        """
+        truthy = {"1", "true", "yes", "on"}
+        if os.environ.get("REACHY_MINI_DISABLE_SW_AEC", "").lower() in truthy:
+            return False
+        if os.environ.get("REACHY_MINI_FORCE_SW_AEC", "").lower() in truthy:
+            return True
+        return None
+
+    def _resolve_sw_aec_enabled(self) -> bool:
```
FabienDanieau: why not detecting? here https://github.com/pollen-robotics/reachy_mini/blob/eb3a508b2ffcfe8e38be0c2fe3418df70967c453/src/reachy_mini/media/audio_gstreamer.py#L379 we already parse the audio devices

src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -180,7 +407,15 @@ def _init_pipeline_record(self, pipeline: Gst.Pipeline) -> None:
         audiosrc.link(queue)
```
FabienDanieau, line 211: It's recommended to attach webrtcdsp directly after the source. This is a mic there is no need for audiobuffersplit I think

src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -363,6 +598,20 @@ def _init_pipeline_playback(self, pipeline: Gst.Pipeline) -> None:
...
+        # tee bin used by `play_sound`. Two reasons:
+        #   1. Sound-file beeps are short and not interpretable by an
+        #      LLM as speech even if they leak into the mic, so the
+        #      cost/benefit doesn't justify the extra wiring.
+        #   2. `webrtcechoprobe` registers a process-wide name, so
+        #      keeping it bound to a single, long-lived element is the
+        #      simplest way to avoid duplicate-name collisions when a
+        #      playbin tee bin spins up mid-conversation.
+        if self._sw_aec_enabled:
+            self._attach_aec_probe(pipeline, tee)
```
FabienDanieau: Same you want to attach the probe close to the source. You may need audiobuffersplit but try without first. No tee should be involved the probe is a passthrough

FabienDanieau (changes requested):

src/reachy_mini/media/audio_gstreamer.py
tfrere: Done in 1f2c459 — dropped both env vars (`REACHY_MINI_DISABLE_SW_AEC` / `REACHY_MINI_FORCE_SW_AEC`) entirely. Detection now relies purely on `has_reachymini_asoundrc()` + `get_audio_device("Source")`, which already covers the `--sim` / `--mockup-sim` / dev-workstation case (no XMOS `.asoundrc`, no ReSpeaker dongle → SW AEC ON). No CLI plumbing needed.

src/reachy_mini/media/audio_gstreamer.py
tfrere: Yes, done in 1f2c459. The new `_resolve_sw_aec_enabled` only consults the existing device-detection helpers (`has_reachymini_asoundrc`, `get_audio_device`); the env-var path is gone. Net diff is -159 LOC and the logic now reads top-to-bottom.

src/reachy_mini/media/audio_gstreamer.py
tfrere: Updated in 1f2c459. `audiobuffersplit` is gone — confirmed via the upstream source that `webrtcdsp` rechunks to 10 ms frames internally via `GstAdapter`, so the splitter was redundant. New record chain: `audiosrc → queue → audioconvert → audioresample → capsfilter(S16LE/16k/2ch) → webrtcdsp → audioconvert → appsink`. The `capsfilter` stays because `webrtcdsp` only accepts `{8,16,32,48}` kHz, so we pin the resampler's output (and match the probe-side caps so both ends of the AEC see the same rate/channels). Verified the pipeline reaches PAUSED on macOS dev box.

src/reachy_mini/media/audio_gstreamer.py
tfrere: Done in 1f2c459. The probe is now inline between `appsrc` and the existing tee — no extra branch, no `fakesink`. New playback head: `appsrc → audioconvert → capsfilter(S16LE/16k/2ch) → webrtcechoprobe → tee → [speaker + wobbler branches]`. `audiobuffersplit` dropped here too. The 6/7 LOCAL `test_audio.py` tests that exercise this pipeline still pass.

tfrere:
Pushed `1f2c459` to address @FabienDanieau's review. Summary of the changes:

**1. Env vars gone.** `REACHY_MINI_DISABLE_SW_AEC` and `REACHY_MINI_FORCE_SW_AEC` are no longer read anywhere. `_resolve_sw_aec_enabled` now only consults `has_reachymini_asoundrc()` + `get_audio_device("Source")` + a plugin-availability probe. The `--sim` / `--mockup-sim` case falls out of the heuristic naturally (no `.asoundrc`, no ReSpeaker → SW AEC ON) so no CLI plumbing is required.

**2. Probe is inline now.** No more parallel tee branch and no `fakesink`. The probe sits between `appsrc` and the existing tee, in passthrough mode:

```
appsrc → audioconvert → capsfilter(S16LE/16k/2ch) → webrtcechoprobe → tee → [speaker + wobbler branches]
```

**3. `audiobuffersplit` dropped on both sides.** Verified from upstream that `webrtcdsp` chunks input into 10 ms frames internally via `GstAdapter`, so the splitter was redundant. As a side benefit we no longer trip the PyGObject `GstFraction` setter segfault, so the `Gst.util_set_object_arg` workaround also goes away.

**4. Both elements attached close to their respective sources** (per your suggestion + the canonical `pulsesrc ! webrtcdsp ! ...` GStreamer example).

The `capsfilter` between `audioresample` and `webrtcdsp` (and the symmetric one on the playback side) is kept because:

- `webrtcdsp` only accepts `{8, 16, 32, 48}` kHz, so the resampler's output rate needs pinning,
- the dsp and the probe MUST agree on rate and channel count, and pinning both to the same canonical caps (`S16LE/16k/2ch`) is the simplest way to guarantee that without spreading the assumption across both pipelines.

### Stats

- Diff is now **+112 / −271** (vs +416 / −1 in v1). One commit on top of the existing one.
- Tests: 7/7 unit tests green (3 pre-existing `_compute_pts` + 4 auto-detection cases), the env-var-driven cases were removed.
- LOCAL backend smoke: 6/7 `test_audio.py` cases pass with SW AEC ON (the 7th is `test_DoA` which needs a physical ReSpeaker, pre-existing failure on `main`).

@FabienDanieau ready for another pass when you have a moment.

tfrere:
**Scope extension: AEC now also patched on the WebRTC media server pipeline.**

Smoke-testing this branch against the desktop tray on macOS revealed that the AEC stage never fired in the mobile-app conversation loop. The reason: that flow runs through `GstMediaServer` (`media/media_server.py`), not the `GStreamerAudio` path the first commit covered.

The daemon actually carries audio through two independent GStreamer pipelines:

| Backend | Module | When it is used |
|---|---|---|
| `GStreamerAudio` | `media/audio_gstreamer.py` | LOCAL Python SDK audio (`MediaManager(backend=LOCAL)`) |
| `GstMediaServer` | `media/media_server.py` | **WebRTC** capture + per-peer playback (mobile app, browser SDK) |

The previous commit only inserted the `webrtcdsp` / `webrtcechoprobe` pair on the LOCAL pipeline, so the mobile conversation path (which is what the original issue is about) still had no echo cancellation in sim mode. This new commit extends the fix to `GstMediaServer`.

To avoid duplicating the helpers, the AEC plumbing was extracted to a new module `media/audio_aec.py`:
- `resolve_sw_aec_enabled(logger)` - the same auto-detection (`.asoundrc` -> ReSpeaker USB -> plugin availability).
- `make_aec_capsfilter()`, `make_aec_probe()`, `build_aec_dsp_chain()`, `build_aec_probe_chain()` - element factories shared by both backends.
- Constants `AEC_PROBE_NAME`, `AEC_SAMPLE_RATE`, `AEC_CHANNELS`.

Wired in `GstMediaServer`:
- `_configure_audio` now inserts `audioconvert -> audioresample -> capsfilter(S16LE/16k/2ch) -> webrtcdsp -> audioconvert` between the mic queue and `webrtcsink`.
- `_on_consumer_pad_added` inserts `audioconvert -> capsfilter -> webrtcechoprobe` between `opusdec` and the speaker/wobbler `tee`.
- The probe element is **created eagerly in `__init__`** and held alive on `self._aec_probe`. `webrtcechoprobe` registers itself in a process-wide registry at `set_property('name', ...)` time, so the paired `webrtcdsp` finds it during sender pipeline startup, even before any peer has connected. Without this, the sender pipeline failed at start with "No echo probe with name `reachymini_aec_probe` found." The same element is reused across peer reconnects (with an unparent step in between) to keep the registry entry stable.

Local verification (macOS, no ReSpeaker dongue):
- `pytest tests/unit_tests/test_audio_gstreamer.py` -> 7 passed (PTS + AEC detection).
- `GstMediaServer(sim_mode=MOCKUP).start()` reaches PLAYING, both `Inserted software AEC (...)` log lines fire, no GStreamer errors on the bus.

FabienDanieau:
Closing, the PR #1189 solves the issue.

PR #1168, opened 2026-05-29
feat/bluetooth/wifi scan and connect

cdussieux:
working well enough in my test runs

can be tested with https://github.com/pollen-robotics/reachy_mini_mobile_app/tree/test/bluetooth-wifi-connect


src/reachy_mini/daemon/app/services/bluetooth/BLE_WIFI_PROVISIONING.md
```diff
@@ -0,0 +1,152 @@
...
+| Command | Auth | Result (via notification) |
+|---|---|---|
+| `WIFI_STATUS` | public | `{"mode","connected","error"}`, plus `"known":[…]` **only when authed** |
+| `WIFI_KEYEX` | public | `{"kid","pk","alg"}` — robot ephemeral pubkey (see below) |
+| `WIFI_SCAN` | session | JSON array of SSIDs (byte-bounded to one MTU) or `ERROR: …` |
+| `WIFI_CONNECT_ENC <json>` | session | `OK: Connecting to <ssid>` or `ERROR: …`; poll `WIFI_STATUS` for outcome |
+| `WIFI_FORGET <ssid>` | session | `OK: Forgotten <ssid>` or `ERROR: …` |
+
+`WIFI_STATUS` withholds the saved-network list from unauthenticated peers (it's
+an owner-location fingerprint). `WIFI_KEYEX` is public because a bare public
```
pierre-rouanet, line 53: Still not sure why it's public? Does it need to?

pierre-rouanet (commented):
## First issue - an attacker nearby can still steal the WiFi password

>   What the PR promises: The whole point of this feature is that the WiFi password is encrypted before it leaves the phone, so nobody listening to the Bluetooth traffic can read it. The PR claims this holds even
  against an attacker who actively interferes (not just one quietly listening).
>   Why that promise is broken: The encryption uses a shared secret that the phone and the robot each compute from a quick key exchange, plus the 5-character PIN printed on the robot. The security argument is: "even
  if you record the traffic, you can't guess the PIN, because you'd also need that shared secret."
>   That argument only works against a passive attacker — someone who just records the radio traffic. It falls apart against an active attacker who sits in the middle:
>   1. When the phone asks the robot for its key, the attacker intercepts and answers with the attacker's own key instead. The phone can't tell the difference — there's nothing proving the key really came from the
  robot.
>   2. Now the phone and the attacker share a secret (not the phone and the robot). The attacker knows this secret completely.
>   3. The phone encrypts the password using that secret plus the PIN, and sends it.
>   4. The attacker already has the secret, so the only thing left protecting the password is the PIN. But the PIN is only 5 characters. The attacker just tries every possible PIN on their own laptop — no need to
  touch the robot again — until the password decrypts. That takes a fraction of a second.
  > So an attacker within Bluetooth range during the few seconds of setup can recover the WiFi password — which is exactly what this feature was built to prevent. The PR's documentation states the opposite ("an
  active attacker can't derive the key without the PIN"), and that statement is simply wrong.
>   The root cause: the phone never verifies it's actually talking to the real robot, and mixing a short PIN into ordinary encryption is not enough to stop this. Properly defending against it needs either:
>  - a method specifically designed so that a short PIN can't be guessed offline (these exist — the secret-exchange has to be built around the PIN, not just flavored with it), or
>   - giving each robot a permanent identity key at the factory that the phone can check, so a fake key gets rejected.

Proposed solution: 
> Recommendation: replace the key exchange with a PIN-based mutual handshake (a PAKE).

>   A PAKE — Password-Authenticated Key Exchange — is a key-agreement protocol built around a shared low-entropy secret (here, the PIN). Instead of "exchange keys, then flavor the result with the PIN," both sides feed the PIN into the exchange itself. The properties you get:
>   - An impostor or man-in-the-middle who doesn't know the PIN ends up with a useless key — the handshake simply fails. They can't get the phone to seal the password with a key they know.
>   - Critically, they can't test PIN guesses offline. Each guess requires a fresh, live attempt against the real robot. That's the whole game: it turns the attack from "100,000 silent guesses on a laptop" into
  "100,000 attempts the robot can see and block."
>   This fits your situation well because it needs no new manufacturing or backend infrastructure — the PIN already exists and is already printed on the robot. You'd keep everything else (AES-GCM sealing the password, the SSID as authenticated context); you only swap out the key-agreement half.

## Second issue - PIN brute forcing

I did not see any code preventing from brute forcing the PIN. Maybe we could
-   Count failed PIN attempts, with exponential backoff and a hard lockout after a small number (e.g. slow down after 3, lock out after ~10), reset on success.
-   Track the counter on the robot itself, not per Bluetooth connection. Otherwise an attacker just disconnects and reconnects to reset it. Tie it to the robot's state so reconnecting doesn't wipe the penalty.

pierre-rouanet:
Second issue should be fixed by https://github.com/pollen-robotics/reachy_mini/pull/1180

FabienDanieau (approved):
Was able to set up and connect to RM Wireless entirely via ble. All good :)

PR #1173, opened 2026-05-29
feat(daemon): portable log streaming without journalctl

tfrere:
## Summary

Daemon log streaming (the `/logs/ws/daemon` WebSocket and the `subscribe_logs` DataChannel command) shelled out to `journalctl -u reachy-mini-daemon`. On hosts without systemd - developer machines, the **desktop / mockup daemon**, any non-systemd Linux - `journalctl` is absent, so both paths failed with `journalctl not found` and clients (e.g. the mobile/desktop app live session) could read no logs at all.

This adds a portable fallback so log streaming works off-robot too:

- New `daemon/log_stream.py`: an in-process ring-buffer `logging.Handler` (bounded, thread-safe emit, async fan-out to live subscribers). Lines are formatted like `journalctl --output short-iso` (`<iso-ts> <message>`) so existing client parsers work unchanged across both sources.
- Installed once at daemon startup in `run_app` (`daemon/app/main.py`).
- Both transports fall back to this buffer on `FileNotFoundError`:
  - WebSocket route `daemon/app/routers/logs.py`
  - DataChannel handler `daemon/backend/abstract.py` (`subscribe_logs`)

`journalctl` is still preferred when present, so **on-robot behaviour is unchanged**. If the buffer was never installed, the previous `journalctl not found` error is preserved.

## Commits

- `style:` pure `ruff format` pass on `daemon/backend/abstract.py` (no logic), isolated so the feature diff stays minimal.
- `feat:` the log-streaming fallback.

## Test plan

- [ ] On a desktop/mockup daemon (macOS, no systemd): open a live session and subscribe to logs - lines now stream instead of `journalctl not found`.
- [ ] On a real robot (systemd present): logs still come from `journalctl` (no behavioural change).
- [ ] WebSocket `/logs/ws/daemon` and the `subscribe_logs` DataChannel path both stream and stop cleanly on disconnect.

### DIFF
--- a/src/reachy_mini/daemon/app/main.py
+++ b/src/reachy_mini/daemon/app/main.py
@@ -24,6 +24,7 @@ from fastapi.staticfiles import StaticFiles
 from fastapi.templating import Jinja2Templates
 
 from reachy_mini.apps.manager import AppManager
+from reachy_mini.daemon import log_stream
 from reachy_mini.daemon.app.routers import (
     apps,
     audio_config,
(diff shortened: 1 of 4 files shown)


tfrere:
Stale (118 commits behind main, no activity since May). The log streaming approach will be redone from scratch on current main if the need comes back.

PR #1177, opened 2026-06-01
Add vision subpackage with MediaPipe HeadTracker

### LINKED ISSUE #1175: Move MediaPipe head tracker from reachy_mini_toolbox to reachy_mini
Add a vision subpackage to expose the MediaPipe-based HeadTracker currently living in [reachy_mini_toolbox](https://github.com/pollen-robotics/reachy_mini_toolbox). The subpackage should live under `src/reachy_mini/vision/` and export `HeadTracker`, plus `mediapipe==0.10.14` should be added as a new optional dependency group named vision, also included in the all group. The existing `exit()` call on missing mediapipe should be replaced with a proper `ImportError`, since calling `exit()` at import time would kill the robot process

alozowski:
Adds daemon-side head tracking, that can run out-of-process so it never stalls the 50 Hz control loop, and to coexist with the conversation app on the CPU-bound CM4 / Pi 4 (supersedes the original in-process MediaPipe migration this branch started as)

**What changed**
- new vision/subpackage: YuNet detector (`cv2.FaceDetectorYN`) + a single-target tracker, run in a spawned `multiprocessing.Process` (own GIL/core). Replaces the in-process MediaPipe tracker, which serialised on the GIL and starved the loop.
- daemon eases the tracking aim per 50 Hz tick -> motion stays smooth regardless of detection rate, so the detector can run slow (10 fps) and cheap
- dedicated valve-gated, downscaled + rate-limited camera branch feeds the worker over its own unixfdsink socket – converts frames only while tracking, nothing when idle.
- full-frame detection at 320 px / 10 fps (no ROI crop), YuNet model fetched from the [HF hub](https://huggingface.co/pollen-robotics/face_detection_yunet_2023mar) 
- multi-person: gaze follows the closest / most-prominent face, with hysteresis so it doesn't ping-pong
- `enable_head_tracking(weight=0)` pauses (frees the head + detector CPU) without tearing down the process -> cheap per-turn on/off for the conv-app speak/listen handoff
- decouples the local IPC framerate from capture (helps https://github.com/pollen-robotics/reachy_mini/issues/1241): serves the IPC feed at a capped fps so on-device vision apps don't pay the full capture rate
- drops the `vision` / `mediapipe==0.10.14` extra, vision installs via the existing opencv extra. Missing OpenCV raises `ImportError` instead of `exit()`, so it can't kill the robot process

**Validated on-robot (Wireless/CM4):**
- loop holds ~50 Hz while tracking; coexists with the conversation app (~1.5 of 4 cores free).
- collision-safe with speaking + emotions/dances; crash isolation – a worker crash doesn't take the daemon down and self-heals on the next turn

DoA-driven search isn't yet in this PR, could be a next PR, as @FabienDanieau suggested.

Closes https://github.com/pollen-robotics/reachy_mini/issues/1175
Conv app PR: https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/381


FabienDanieau:
Is it an issue to have reachy_mini_toolbox as a dep in the conv app?

pierre-rouanet:
LGTM, indeed as @FabienDanieau points out. Maybe we can use toolbox as a dependency. Not strong opinion on this. Just if we do, we have to make sure that everything included runs on a RM wireless.

Maybe next step is ton add start/stop/getface API entry points to trigger the head tracking in background?

FabienDanieau:
I made some tests. [rf-detr-nano](https://huggingface.co/pollen-robotics/rf-detr-nano-face/tree/main) is too heavy for the rpi. Mediapipe seems to be the most suitable solution for edge devices. There is a lighter model that we can use instead of the facemesh of the tool box: see [face_track.py](https://github.com/user-attachments/files/28961441/face_track.py). 
We need to mind the [numpy 1.x dependency](https://github.com/google-ai-edge/mediapipe/issues/5612) though.

> Maybe next step is ton add start/stop/getface API entry points to trigger the head tracking in background?

I agree with @pierre-rouanet, the right call would be to enable head tracking on the daemon side (in a way similar we did for the wobbling see https://github.com/pollen-robotics/reachy_mini/pull/1001). Basically we need to add a appsink to the video pipeline to feed the head tracker. 

Also the tracking could be extended with the DoA. But let's keep that for another issue :)

alozowski:
let's say that mediapipe works here though it's not ideal:
https://github.com/pollen-robotics/reachy_mini/pull/1177/changes/6c7a9111029ce5be2c16f824d195cc2dfb8b00c0

I'm continuing my experiment with completely another approach in this branch

alozowski:
next checkpoint, commit 9a5b52fa926f4e93c33bd6a2d6c03dec2e585829
- opencv works 
- start the daemon from this branch
- enable headtracking with `curl -s -XPOST localhost:8000/api/media/tracking/enable -H Content-Type:application/json -d '{}'`
- disable it with `curl -s -XPOST localhost:8000/api/media/tracking/disable -H Conte{}` (truncated command, I know, but it works)

alozowski:
next checkpoint, commit c3063c2be7cd35aee809f086769e8d65392df743

src/reachy_mini/assets/face_detection_yunet_2023mar.onnx
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:8f2383e4dd3cfbb4553ea8718107fc0423210dc964f9f4280604804ed2552fa4
+size 232589
```
FabienDanieau: hugging face would be a better place for a model :)

alozowski: indeed!

alozowski: uploaded the model to the hub and updated the model upload settings accordingly:
https://huggingface.co/pollen-robotics/yunet-face-detection-2023mar

alozowski:
next checkpoint, almost the final: https://github.com/pollen-robotics/reachy_mini/pull/1177/changes/5b421106b20785a3009455e5c54bc87c7442395f
- we now resample instead of crop: detect on the whole frame downscaled to 320 px (gstreamer videoscale in the daemon tracking branch) every tick, without ROI window, so full FOV and robust to fast motion
- detection rate is 10 fps, but the head stays smooth because the daemon eases the aim at 50 Hz independent of the detection rate
- YuNet score threshold is 0.6 so it holds the face through head turns and at ~1 m
- YuNet model fetched from HF hub instead of the vendored ONNX (first enable needs network once, then it's cached)
- weight 0 now pauses tracking without tearing down the detector; weight > 0 resumes instantly
- local IPC camera feed capped at 15 fps (also touches the client `camera_gstreamer.py`) so on-device apps don't pay the full 30 fps convert

alozowski:
another checkpoint, https://github.com/pollen-robotics/reachy_mini/pull/1177/changes/f710992cad0f6b437f4a1b0d2070aa55d1559f6b

we don't yet have DoA tracking but I added a simple multi-person one:
- gaze follows the most prominent face (largest ≈ closest, weighted toward center) instead of sticking to the first one acquired
- switches only after a challenger leads by a margin for a dwell window, so no ping-pong between comparable faces as it used to be

alozowski:
and another big checkpoint (https://github.com/pollen-robotics/reachy_mini/pull/1177/changes/5db721e1aa6284837678ea17f3f7ff0092aff705) that dropped opencv from the tracking runtime, YuNet now runs on onnxruntime with a numpy decode:
- same detector and same weights: the model is the official dynamic-input YuNet export (`face_detection_yunet_2026may`, byte-identical weights to 2023mar), hosted on our HF and pinned by revision
- the decode reproduces `cv2.FaceDetectorYN` bit-for-bit, there's a parity test running both on the same model and frame
- frames are padded up to /32 (320x180 -> 320x192) because YuNet's feature pyramid needs stride-aligned dims, opencv did this silently, raw ORT crashes without it
- ORT session is pinned to 1 thread so the detector can't spread across cores and fight the control loop
- the robot installs [vision] (onnxruntime, ~16 MB) instead of [opencv] (~50 MB wheel / ~119 MB unpacked), so opencv is now dev/calibration-only
- `onnxruntime` bumped to `1.27.0` for both `vision` and `nn_kinematics` (verified FK/IK bit-identical), which needs python >=3.11

src/reachy_mini/vision/face_tracking.py
```diff
@@ -0,0 +1,285 @@
...
+        from reachy_mini.vision.face_detector import FaceDetector
+
+        Gst.init([])
+        if platform.system() == "Windows":
+            source = f"win32ipcvideosrc pipe-name={CAMERA_TRACKING_PIPE_NAME}"
+        else:
+            source = f"unixfdsrc socket-path={CAMERA_TRACKING_SOCKET_PATH}"
+        pipeline = Gst.parse_launch(
+            f"{source} ! queue leaky=2 max-size-buffers=1 ! "
+            "appsink name=sink drop=true max-buffers=1 sync=false"
```
FabienDanieau: make a pipeline with separate elements. See other comments regarding IPC

src/reachy_mini/media/media_server.py
```diff
@@ -969,19 +990,99 @@ def _build_ipc_branch(
...
             for elem in [identity, videoconvert_ipc, capsfilter_ipc, ipc_sink]:
                 pipeline.add(elem)
 
-            queue_ipc.link(identity)
+            videorate_ipc.link(identity)
             identity.link(videoconvert_ipc)
             videoconvert_ipc.link(capsfilter_ipc)
             capsfilter_ipc.link(ipc_sink)
 
+    def _build_tracking_branch(self, tee: Gst.Element, pipeline: Gst.Pipeline) -> None:
```
FabienDanieau: Why do we need a new branch? Can't the facetracking pipeline be a client of the current ipc branch?

src/reachy_mini/media/media_server.py
```diff
@@ -969,19 +990,99 @@ def _build_ipc_branch(
...
+        """
+        src_w, src_h = self.resolution
+        width = min(TRACKING_WIDTH, src_w)
+        height = max(2, round(width * src_h / src_w / 2) * 2)
+
+        queue = Gst.ElementFactory.make("queue", "tracking_queue")
+        valve = Gst.ElementFactory.make("valve", "tracking_valve")
+        videorate = Gst.ElementFactory.make("videorate", "tracking_videorate")
+        videoscale = Gst.ElementFactory.make("videoscale", "tracking_videoscale")
+        videoconvert = Gst.ElementFactory.make("videoconvert", "tracking_videoconvert")
```
FabienDanieau: you may want to use v4l2convert for the rpi

src/reachy_mini/media/media_server.py
```diff
@@ -82,6 +84,13 @@
 SESSION_FAILED_REASON_ICE_TIMEOUT = "ice_negotiation_timeout"
 SESSION_FAILED_REASON_PC_FAILED = "peer_connection_failed"
 
+# Downscaled, rate-limited feed for the out-of-process tracker; width & fps trade quality vs CPU.
+TRACKING_WIDTH = 320
+TRACKING_FPS = 10
+
+# Cap the local IPC feed below the capture rate; vision readers don't need the full frame rate.
+IPC_FPS = 15
```
FabienDanieau: why two fps limiters then?

src/reachy_mini/vision/face_detector.py
```diff
@@ -0,0 +1,141 @@
...
+        self, score_threshold: float = 0.6, nms_threshold: float = 0.3
+    ) -> None:
+        """Create the YuNet detector, confined to one CPU thread."""
+        model_path = hf_hub_download(_MODEL_REPO, _MODEL_FILE, revision=_MODEL_REVISION)
+        options = ort.SessionOptions()
+        # Confine the detector to one thread so it can't spread across cores and starve the loop.
+        options.intra_op_num_threads = 1
+        options.inter_op_num_threads = 1
+        self._session = ort.InferenceSession(
+            model_path, options, providers=["CPUExecutionProvider"]
```
FabienDanieau, line 73: could it be gpu on lite?

src/reachy_mini/vision/face_tracking.py
```diff
@@ -0,0 +1,285 @@
...
+        tracker = Tracker()
+    except Exception as e:
+        _logger.warning("Face tracker failed to start: %s", e)
+        return
+
+    crop_scale = camera_specs.default_resolution.value[3]
+    camera_matrix: NDArray[np.float64] | None = None
+    try:
+        while not stop.is_set():
+            sample = appsink.emit("try-pull-sample", 200_000_000)
```
FabienDanieau: you should be able to call pull sample directly (not using emit)

FabienDanieau (approved):

src/reachy_mini/vision/face_tracking.py
alozowski: Done! The worker now builds source -> queue -> videorate -> convert -> capsfilter -> appsink from explicit `ElementFactory.make` calls with properties

src/reachy_mini/media/media_server.py
alozowski: You're right, done! The dedicated branch is deleted and the tracker is now a client of the main IPC socket, same as `camera_gstreamer.py`

src/reachy_mini/media/media_server.py
alozowski: Done! The worker prefers v4l2convert (ISP does scale + convert in hardware, and it's the client side now), falling back to videoscale + videoconvert where it's unavailable

src/reachy_mini/media/media_server.py
alozowski: True, it was my oversight, I kept only one now: `IPC_FPS=15` is the daemon-side cap for all local consumers, the detector's 10 fps is a client-side videorate inside the worker, so the daemon no longer knows about detection rates

src/reachy_mini/vision/face_detector.py
alozowski: Currently it's deliberately CPU: the onnxruntime package ships only the CPU EP (GPU would mean `onnxruntime-gpu` + a CUDA stack), and the model runs in ~1–2 ms at 10 fps (~1% of one core), so transfer overhead would exceed the compute. Pinning `CPUExecutionProvider` keeps wireless and lite identical

src/reachy_mini/vision/face_tracking.py
alozowski: Done: `appsink.try_pull_sample(200 * Gst.MSECOND)`
The `emit()` workaround only existed because I didn't import GstApp

alozowski:
another checkpoint, https://github.com/pollen-robotics/reachy_mini/pull/1177/changes/a5018b04b1df818ff30c18ea1b771868ed802ce2: the tracker is now just a camera client
- dedicated tracking branch/socket/valve deleted: the tracker reads the same camera IPC as any local app, pause = the worker just disconnects
- one videorate total: `IPC_FPS=10` on the daemon paces everyone, detector included (measured with a fakesink client: 123 frames / 12 s)
- aim at the nose now (eye midpoint made the robot look slightly above), single-face only
- explicit gst elements, direct `try_pull_sample()`, `v4l2convert` on RPi, no more ctrl+c traceback
- example + SDK docs added

next to try: drop the worker process and use a plain thread, and it would save ~200 MB of duplicated python/ORT/GStreamer stack

FabienDanieau (approved):
Yeah thanks!

PR #1182, opened 2026-06-02
media: Python aiortc central consumer + Cloudflare TURN relay + creds-only SDK mode

FabienDanieau:
## Summary

Adds a hardware-free **Python WebRTC consumer** of the HF central signaling relay, plus the supporting pieces that let a cloud backend (HF Space, Cloud Run, …) read a Reachy Mini's camera/mic and drive it using a visitor's HF token — without instantiating the full `ReachyMini` hardware stack.

### What's in the branch
- **`ReachyCentralConsumer`** (`src/reachy_mini/media/central_consumer.py`) — an aiortc client that subscribes to central over SSE, negotiates WebRTC with one robot, and exposes decoded video frames (numpy RGB), bidirectional audio (`on_pcm` + outbound track), and a thread-safe `send_command` data channel. Includes a DTLS-cipher shim so aiortc completes DTLS with the robot's GStreamer `webrtcsink`.
- **Cloudflare TURN relay** (`media_server.py`) — the robot now offers short-lived TURN `relay` candidates (fetched from HF's hosted proxy with the daemon's own HF token, cached at half-TTL) so NAT-restricted remote consumers can connect. Credentials are never logged.
- **`connectToHostCredsOnly`** (TS host SDK) — a credentials-only embed boot for backend-owned WebRTC apps: forwards `{hfToken, robotPeerId, signalingUrl}` to the app's backend instead of opening a session in the iframe, avoiding a "robot busy" race.
- **Docs** — `docs/source/SDK/cloud-backend-consumer.md` and media-architecture updates.

### Robustness & hardening (latest commit)
- Recover from a **stale/lost producer**: the robot can reconnect to central under a new peerId; session establishment now runs solely through the discovery loop (pinned targets included), and a "Producer not found" drops the stale target so the robot is re-picked by name rather than retried forever on the SSE-read path.
- **Reconnect on a failed PeerConnection** (exceptions in aiortc event handlers don't reach the run loop, so we break the SSE loop instead of raising).
- Clear the last decoded frame on teardown so `latest_frame()` doesn't return a stale image after a disconnect.

### Hygiene
- `central_consumer.py` and the TURN helpers pass `ruff` and `mypy --strict`.

## Token safety
HF tokens travel only in `Authorization: Bearer` headers (never URLs/logs); TURN credentials are stripped from log lines (host only). A diff-wide scan found no hard-coded secrets.

## Test plan
Exercise against these backend-owned WebRTC Spaces (each builds a `ReachyCentralConsumer` per visitor):
- [ ] [pollen-robotics/hibiki-zero](https://huggingface.co/spaces/pollen-robotics/hibiki-zero) — bidirectional audio (mic in / speaker out)
- [ ] [pollen-robotics/reachy-stories](https://huggingface.co/spaces/pollen-robotics/reachy-stories) — audio + `goto_target` head commands on the data channel
- [ ] [pollen-robotics/rf-detr-realtime-webcam](https://huggingface.co/spaces/pollen-robotics/rf-detr-realtime-webcam) — video frame pull + realtime inference

For each: confirm connect succeeds (incl. via the offered TURN relay from a NAT-restricted Space), then restart the robot mid-session and confirm the consumer re-discovers the new producer and reconnects.


src/reachy_mini/media/central_consumer.py
```diff
@@ -0,0 +1,1053 @@
...
+* a thread-safe ``send_command`` for the ``data`` channel the daemon offers,
+  plus an ``on_command_ready`` hook for one-shot setup once it opens.
+
+All audio/command extras are optional — with none of them set this behaves
+exactly as the original video-only consumer.
+
+Use this from any cloud backend (HF Space, Cloud Run, etc.) that needs to
+read a Reachy Mini's camera and/or drive the robot via the visitor's HF
+token, without instantiating :class:`reachy_mini.ReachyMini` (which
+requires daemon access and the full hardware stack).
```
pierre-rouanet, line 22: As it is a submodule of reachy_mini it still requires to install reachy_mini and all dependencies. We could imagine having a reachy_mini_consumer as a lighter package. But right now, I'm pretty convinced it's not worth it.

src/reachy_mini/media/central_consumer.py
```diff
@@ -0,0 +1,1053 @@
...
+        on_pcm: Optional[Callable[[npt.NDArray[np.float32]], None]] = None,
+        out_track: Optional[MediaStreamTrack] = None,
+        on_command_ready: Optional[Callable[[], None]] = None,
+        audio_sample_rate: int = DEFAULT_AUDIO_SAMPLE_RATE,
+    ):
+        """Build a single-robot consumer.
+
+        Audio/command arguments are all optional; with none set this behaves
+        exactly as the original video-only consumer.
+
```
pierre-rouanet, line 190: Maybe also add comment for (at least) ice_servers_provider as it not obvious what can be  provided here.
I've seen after that it is actually explained in the md. Maybe just point there.

src/reachy_mini/media/central_consumer.py
```diff
@@ -0,0 +1,1053 @@
...
+        t = msg.get("type")
+        if t == "welcome":
+            self._self_peer_id = msg.get("peerId")
+            lease = msg.get("lease_seconds")
+            rec = msg.get("recommended_heartbeat_interval_seconds")
+            if isinstance(rec, (int, float)) and rec > 0:
+                self._heartbeat_interval = max(2.0, float(rec))
+            elif isinstance(lease, (int, float)) and lease > 0:
+                self._heartbeat_interval = max(2.0, float(lease) / 3.0)
+            print(f"[reachy] welcome peerId={self._self_peer_id} "
```
pierre-rouanet, line 462: replace all prints by logs

src/reachy_mini/media/media_server.py
```diff
@@ -299,6 +395,33 @@ def _consumer_added(
         # forever). See `ICE_NEGOTIATION_DEADLINE_S`.
         self._install_negotiation_watchdog(peer_id, webrtcbin)
 
+    def _apply_turn_servers(self, webrtcbin: Gst.Element) -> None:
```
pierre-rouanet, line 398: Maybe we should still add a parameter to make this optional, no? Even if it's best effort I'm wondering if this could not slow down the procedure in case where the turn server is slow to answer and we actually don't need it.
Especially as current timeout in requesting the turn server are pretty large and synchronous!
What do you think?

src/reachy_mini/media/central_consumer.py
```diff
@@ -0,0 +1,1053 @@
...
+
+    def is_connected(self) -> bool:
+        """Return True while the peer connection is in the "connected" state."""
+        return self._pc_connected
+
+    def is_command_ready(self) -> bool:
+        """Return True once the robot's ``data`` channel is open for commands."""
+        return self._cmd_channel_open and self._cmd_channel is not None
+
+    def send_command(self, envelope: dict[str, Any]) -> bool:
```
pierre-rouanet, line 276: Should the envelope relies on the existing pydantic model instead? As they are in the same package, we can have some static checks.

src/reachy_mini/media/central_consumer.py
```diff
@@ -0,0 +1,1053 @@
...
+                kinds.append(scheme + (" (auth)" if s.username else ""))
+        print(f"[reachy] building pc with ice servers: {kinds}")
+        config = RTCConfiguration(iceServers=ice_servers)
+        pc = RTCPeerConnection(configuration=config)
+
+        @pc.on("track")  # type: ignore[misc]  # pyee .on() is untyped
+        def _on_track(track: MediaStreamTrack) -> None:
+            print(f"[reachy] received remote track kind={track.kind}")
+            if track.kind == "video":
+                asyncio.create_task(self._consume_video(track))
```
pierre-rouanet, line 753: You should keep a reference to the created task, otherwise the GC will remove it sometimes 😓

pierre-rouanet (changes requested):
Code seems fine to me! I've requested a few easy/small changes in comments.

I'll do some real testing later.

src/reachy_mini/media/central_consumer.py
FabienDanieau: Agreed, not worth it. And #5 pulls the other way: reusing the `io.protocol` models only works in-package.

Your premise was broken though. `aiortc` and `av` were declared nowhere in `pyproject.toml`, so a clean install could not import this module at all. Added a `central-consumer` extra (`1b961c7`) rather than core deps, since a robot never runs the consumer.

src/reachy_mini/media/central_consumer.py
FabienDanieau: Done in `ee6a75d`. Real `Args:` block covering all ten params, with `ice_servers_provider` pointing at the md section. Ruff D417 only checks completeness once an `Args:` header exists, which is why the gap went unnoticed.

Also fixed a dead relative link in that doc, plus a local worktree name I had leaked into it.

src/reachy_mini/media/central_consumer.py
FabienDanieau: Done in `4df0b02`, all 30 of them.

Level picked by what recovers from the condition: `exception` where nothing retries (the video/audio consumer tasks, whose death silently freezes `latest_frame`), `warning` where a loop retries or the item is skipped, `debug` for `MediaStreamError`, which is how aiortc signals a normal end-of-track.

Also dropped the `[reachy]` prefix. `getLogger(__name__)` already gives the module, and 93% of logger calls in `src/` carry no prefix.

src/reachy_mini/media/central_consumer.py
FabienDanieau: Done in `c1f46ba`. `send_command` takes `AnyCommand | dict[str, Any]` and serializes models with `model_dump_json(exclude_none=True)`, so a malformed command is a type error here instead of an `"Invalid command"` reply from the far end.

Kept the dict path. Since you wrote this, main added JSON-RPC app-control on the same channel, and those frames are not `AnyCommand`.

`exclude_none` is safe: no nullable field in the 51 command models has a non-None default, so absent parses identically to explicit null.

src/reachy_mini/media/central_consumer.py
FabienDanieau: Good catch, fixed in `712a569`. Both go through `_spawn_track_task`, which holds the task and discards it on completion so the set does not grow across reconnects. The other three `create_task` calls already stored theirs.

Also cancelled on teardown, so one blocked elsewhere cannot outlive its session and keep feeding `on_pcm`.

src/reachy_mini/media/media_server.py
FabienDanieau: You were right, and it was worse than slow: `requests` applies `timeout` per socket operation and not to DNS at all, so the worst case was unbounded. It also recurred, since the cache expired at half-TTL and the next consumer refilled it.

Fixed in `9009a2e` by removing the blocking rather than gating it. `TurnCredentials` (now in `webrtc_utils`, per your other point, none of it touched GStreamer) refreshes on its own thread, and `_apply_turn_servers` is a cache read. Negotiation does no I/O, so the timeout question and the double-checked lock both go away.

The opt-out is `enable_turn`, read from a `turn_enabled` key in the daemon config rather than an env var, next to `startup_app` and `speaker_eq_gains`. Defaults on, so cloud apps work unconfigured.

One thing to expect when testing: libnice now logs `We still have alive TURN refreshes` on disconnect. It is webrtcsink freeing the ICE agent without `nice_agent_close_async()`, and can only appear now that TURN is configured. Costs one stranded allocation until it expires, relaying nothing. Working around it needs `gir1.2-nice-0.1` on the image plus racing webrtcsink teardown, so I left it.

`ice_servers_to_turn_uris` is public and tested now, including that a credential containing `:@/` stays percent-encoded.

FabienDanieau:
Pushed, all six comments addressed, one commit each on top (no rewrite, since review had started). Merged `main` in first: the branch was 129 commits behind and `media_server.py` was one of the most-changed files in that gap.

Three things found on the way, worth flagging:

- **`aiortc`/`av` were undeclared.** Module-level imports, declared nowhere in `pyproject.toml`, so this module could not be imported from a clean install. Now a `central-consumer` extra.
- **That broke `mypy`.** With the packages installed, `ignore_missing_imports` stopped resolving them to `Any` and 13 real errors surfaced. The earlier "passes `mypy --strict`" claim held only because the imports were unresolved. Fixed in the same commit, green over 145 files.
- **A log flood I introduced and caught.** The refresher treated "no HF token" as a transient failure, so a robot that is not logged in would have warned every 30s forever. Now only genuine network failures get the short backoff.

Tested on hardware, connects and streams. Expect a new libnice `alive TURN refreshes` warning on disconnect, explained in the TURN thread.

pierre-rouanet (approved):

PR #1186, opened 2026-06-02
Clear queued playback audio on WebRTC backend (barge-in)

### LINKED ISSUE #1108: GstWebRTCClient does not expose a way to clear queued playback audio
## Summary

When using Reachy Mini Wireless through the WebRTC media backend, realtime conversation interruptions do not stop already-buffered robot speech. The consuming app receives the interruption signal and attempts to clear playback, but the WebRTC audio backend does not expose a clear method equivalent to `GStreamerAudio.clear_player()`.

## Context

In `reachy_mini_conversation_app`, `input_audio_buffer.speech_started` is treated as user barge-in. The app calls its `clear_audio_queue()` hook, which logs:

```text
User intervention: flushing player queue
```

For local GStreamer audio, that path should call `audio.clear_player()` and `GStreamerAudio.clear_player()` then logs:

```text
Cleared player queue
```

On a wireless robot, however, the media backend appears to be `MediaBackend.WEBRTC`, where `robot.media.audio` is `GstWebRTCClient`. In the installed `reachy-mini==1.7.1` code, `GstWebRTCClient` has an audio `appsrc` and `push_audio_sample()`, but does not expose `clear_player()` or `clear_output_buffer()`.

As a result, the app enters the clear path, but no backend clear method is called, and already queued assistant audio keeps playing.

## Evidence

During a test on 2026-05-07, the realtime backend did cancel generation correctly:

```text
23:57:28.583 speech started
23:57:28.589 response done status=cancelled
23:57:28.589 Speech during response: cancelled, queue flushed
23:57:28.595 TTS generation cancelled (interruption)
```

The conversation app also received the interruption and attempted to flush playback:

```text
23:57:28.638 reachy_mini_conversation_app.console | User intervention: flushing player queue
```

But there was no corresponding `Cleared player queue` log from the SDK, and the robot continued speaking the previously buffered audio.

## Expected behavior

`GstWebRTCClient` should provide a public playback clear method, probably named consistently with the other backend, e.g. `clear_player()` or `clear_output_buffer()`, that immediately drops queued WebRTC/GStreamer output audio during barge-in.

The conversation app can then call the same clear method across local GStreamer and WebRTC backends.

## Possible fix direction

Add a clear method to `GstWebRTCClient` that flushes or resets the playback pipeline/appsrc strongly enough to drop queued audio already downstream. This should be equivalent in intent to `GStreamerAudio.clear_player()`, but may need to account for WebRTC/GStreamer pipeline buffering beyond `appsrc`.

It would also be useful to log when the clear actually runs, so consumers can distinguish "app attempted clear" from "backend playback buffer was actually flushed".

FabienDanieau:
## Summary

Fixes #1108. `GstWebRTCClient.clear_output_buffer()` was a no-op, so realtime-conversation barge-in could not stop already-buffered robot speech over the WebRTC backend.

In the WebRTC backend audio flows **client → daemon**, and the bulk of buffered speech sits in the **daemon's** incoming-audio playback pipeline (`appsrc → opusdec → queue → audiosink`, `sync=True`), not in the client send chain — so a complete fix is two-sided.

## Changes

- **Client** (`webrtc_client_gstreamer.py`): new `clear_player()` flushes the local audio send-chain appsrc and POSTs `/api/media/clear_incoming_audio` so the daemon drops audio already queued for the speaker. `clear_output_buffer()` deprecated (warn-only).
- **Local backend** (`audio_gstreamer.py`): `clear_output_buffer()` deprecated to match; real flush stays in `clear_player()`.
- **`AudioBase`**: `clear_player()` promoted to an abstract method (both backends implement it), so callers can drop `hasattr` guards.
- **Daemon flush** reachable via **both** transports → one `Backend.clear_incoming_audio()` → `GstMediaServer.clear_incoming_audio()` (flushes the per-peer playback pipeline with `flush_stop(reset_time=False)` to preserve the shared-clock timeline):
  - REST: `POST /api/media/clear_incoming_audio` (used by the Python client).
  - Data channel: `ClearIncomingAudioCmd` (used by the TS SDK).
- **TypeScript SDK**: `clearIncomingAudio()` on `ReachyMini` / `ReachyMiniInstance`, documented in the JS SDK reference.

## Verification

- `ruff check` + `mypy` (strict) clean on changed files.
- End-to-end on a wireless robot: barge-in stops the robot promptly; client logs `Cleared player queue (WebRTC send chain flushed)` and daemon logs `Flushed incoming audio playback`.

The companion app-side change is pollen-robotics/reachy_mini_conversation_app#376.

### DIFF
--- a/docs/source/API/openapi.json
+++ b/docs/source/API/openapi.json
@@ -1314,6 +1314,29 @@
         }
       }
     },
+    "/api/media/clear_incoming_audio": {
+      "post": {
+        "summary": "Clear Incoming Audio",
+        "description": "Drop audio received from WebRTC clients that is queued for the speaker.\n\nUsed for barge-in so the robot stops speaking already-buffered audio.",
+        "operationId": "clear_incoming_audio_api_media_clear_incoming_audio_post",
+        "responses": {
+          "200": {
+            "description": "Successful Response",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "additionalProperties": {
+                    "type": "string"
+                  },
+                  "type": "object",
+                  "title": "Response Clear Incoming Audio Api Media Clear Incoming Audio Post"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
     "/api/media/wobbling/enable": {
       "post": {
         "summary": "Enable Wobbling",
--- a/docs/source/SDK/javascript-sdk.md
+++ b/docs/source/SDK/javascript-sdk.md
@@ -109,6 +109,7 @@ new ReachyMini({
 | `setAntennasDeg(right, left)` | `boolean` | Set antenna positions in degrees (wraps `setTarget`) |
 | `setBodyYawDeg(yaw)` | `boolean` | Set body yaw in degrees (wraps `setTarget`) |
 | `playSound(filename)` | `boolean` | Play a sound file on the robot |
+| `clearIncomingAudio()` | `boolean` | Drop audio queued for the robot speaker (barge-in) |
 | `sendRaw(data)` | `boolean` | Send arbitrary JSON via data channel |
 | `requestState()` | `boolean` | Request a state snapshot |
 | `setAudioMuted(muted)` | — | Mute/unmute robot speaker (local) |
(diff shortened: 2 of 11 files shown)


andimarafioti (approved):
Looking good!!

Issue #569, opened 2025-12-17
[Improvement] louder sound


clmnt:
Model: beta reachy mini lite

When I'm in a noisy environment or I want to show reachy to people who are not super close, the sound of reachy mini (for example in the conversation app) is not strong enough so it would be nice if the next version's microphone was stronger. Can also be useful to play music in a room.

brainwavecoder9:
There is a lot with the microphone that isn't being utilized, which could help a lot with noisy environments.

It's difficult to imagine a lot being done related to the speaker, but I am exploring ways to extend it for external hardware.

oojacoboo:
The speaker leaves a lot to be desired.  It's not even very crisp/clear

adspiceprospice:
Can't the speaker volume be increase through a firmware update, i'm pretty sure it can handle more, now it's not even as loud as a phone speaker. Also, i'm not looking to see improvements in the next version as much as i'm looking to see you guys actually offer recommendations for improvement parts.

Toon-nooT:
I also noticed that the volume level is too low when using it at a booth or outside. The volume between gemini vs openai realtime differs a lot. Gemini realtime is much louder by itself. I was thinking of adding a fixed, daemon-level output gain stage for audio playback, controlled by an environment variable:

- `REACHY_AUDIO_GAIN_DB`

This feature should apply a constant gain in decibels to audio sent to the output sink.
Risk is clipping, but if the variable is not set this stage will not be in effect.

Toon-nooT:
i opened a PR: https://github.com/pollen-robotics/reachy_mini/pull/1187
For me it works very well and solves my problem of Reachy being too quiet onstage/outdoors

FabienDanieau:
There is a new firmware with +6dB gain: https://github.com/pollen-robotics/reachy_mini/tree/fix/sound_volume/src/reachy_mini/assets/firmware
Please let us know how it goes

PR #1187, opened 2026-06-03
feat: add daemon-level audio output gain with limiter

Toon-nooT:
Adds a configurable audio output gain stage to boost the robot's speaker volume at the daemon level.
Referenced Issue: [Improvement] louder sound #569

### What it does

- Reads `REACHY_AUDIO_GAIN_DB` env var at startup (default: 0 dB = no change)
- Inserts a GStreamer `volume` element in all three audio output pipelines (playbin, WebRTC incoming, Python SDK `push_audio_sample()`)
- Adds `rglimiter` after the volume element to prevent clipping distortion at high gain
- Exposes a REST API (`GET/POST /api/audio/gain`) for runtime gain adjustment without restart

### Why

The ALSA master volume on the robot is already at 100%. The speaker/amp combination is simply too quiet at native output level. This provides a clean software gain stage with clipping protection to make the robot audibly louder.

### Recommended usage

Set `REACHY_AUDIO_GAIN_DB=6` in the systemd service for a moderate boost, or up to `12` for approximately 2× perceived loudness. The limiter minimizes distortion.

### Testing

- 19 unit tests covering dB conversion, env parsing, clamping, and runtime set
- Manual verification on Reachy Mini Wireless hardware at +6 dB and +12 dB
- ruff + mypy clean

### AI Disclosure
Assisted-by: Claude:claude-opus-4-6

### DIFF
--- a/docs/source/SDK/audio_output_gain.md
+++ b/docs/source/SDK/audio_output_gain.md
@@ -1,0 +1,153 @@
+# Audio Output Gain (REACHY_AUDIO_GAIN_DB)
+
+## Purpose
+
+Reachy Mini is not well audible in noisy environments, outdoors, convention booths, ... even with volume at 100%.
+Proposed here is a daemon-level output gain stage that boosts audio sent to the robot's speaker. This is intended for repeatable, cross-platform signal-level correction.
+
+## How it works
+
+- At daemon startup the environment variable `REACHY_AUDIO_GAIN_DB` is read and parsed as a floating dB value.
+- The value is converted to a linear multiplier (gain) and applied to a GStreamer `volume` element placed before the platform audio sink.
+- The gain is applied to **all three** audio output paths:
+  1. **Daemon playbin** — used by `play_sound()`, recorded/move audio, emotion sounds (via `_build_audiosink_tee_bin`)
+  2. **WebRTC incoming audio** — browser mic played on the robot's speaker (per-peer pipeline)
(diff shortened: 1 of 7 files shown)


Toon-nooT:
Here is a measurement. Left side is default, right side is with REACHY_AUDIO_GAIN_DB = 12
It shows a ~10dB increase when Gemini is talking.

Screenshot 2026-06-04 085458
With 12 i can hear fidelity loss in the voice, but overall still better hearable because it's much louder.

Sidenote: The speaker has a rating of 5W, i am not impressed with its clarity, maybe an EQ config helps. Maybe a drop in replacement speaker can be recommended by someone?

FabienDanieau:
Thanks for the PR. We're investigating if there is something that can be done at the firmware level of the sound card first. I'll get back to you

FabienDanieau:
Can you try the latest firmware 2.1.4 and tell me if it's better please? 
https://github.com/pollen-robotics/reachy_mini/tree/fix/sound_volume/src/reachy_mini/assets/firmware

Toon-nooT:
@FabienDanieau 
I tested this. First i recorded a 10 second song sample. Then i used a db meter app, fixed on a certain distance from Reachy. This the left measurement. Next i performed the update with `dfu-util -R -e -a 1 -S <serial_nr> -D reachymini_ua_io16_lin_v2.1.4.bin`
i replayed the same audio clip and that is the measurement on the right side.

image

Conclusion: much better. This is actually making this pull request unnecessary, as it is fixed directly at the source and not through a 'gain hack'.

PR #1189, opened 2026-06-03
software AEC for sim/mockup sim mode

FabienDanieau:
replace https://github.com/pollen-robotics/reachy_mini/pull/1162

When reachy mini's audio card is not detected, typically with reachy-mini-daemon --sim/mockup-sim, we enable gstreamer echo cancellation. 

Can be tested with [this branch](https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/389) of the conv app


andimarafioti (dismissed):
Sorry, meant to review 1186

src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -406,15 +486,26 @@ def push_audio_sample(self, data: npt.NDArray[np.float32]) -> None:
...
-        )
+        # Gap-aware PTS for audiomixer: the first buffer of a new utterance
+        # (gap > GAP_RESET_NS) carries DISCONT + a running-time PTS so the
+        # mixer aligns it on the current timeline. Follow-up buffers are
+        # untimestamped (CLOCK_TIME_NONE) so the mixer places them
+        # contiguously by byte offset — avoids drift from per-buffer
+        # stamping that the mixer would otherwise treat as "late".
+        running_time = self._appsrc.get_current_running_time()
+        duration_ns = (int(data.shape[0]) * Gst.SECOND) // self.SAMPLE_RATE
+        new_cue = running_time > self._appsrc_pts + self.GAP_RESET_NS
```
alozowski: `_appsrc_pts` starts at `-1`, so on the first push this evaluates to `running_time > 199ms` – always `False` right after `start_playing()`. The first buffer goes out without a PTS. Is it possible to add the `< 0` guard back, as `_compute_pts had it` in audio_base.py`, L79?

Something like:
```
new_cue = self._appsrc_pts < 0 or running_time > self._appsrc_pts + self.GAP_RESET_NS
```

src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -151,9 +156,15 @@ def _init_pipeline_record(self, pipeline: Gst.Pipeline) -> None:
...
-                    "No specific audio card found, using default audio source."
+                    "No specific audio card found, using default audio source. Enabling webrtcdsp."
                 )
                 audiosrc = Gst.ElementFactory.make("autoaudiosrc")  # use default mic
+                self._webrtcechoprobe = Gst.ElementFactory.make("webrtcechoprobe")
+                webrtcdsp = Gst.ElementFactory.make("webrtcdsp")
+                # Pair probe ↔ dsp so the playback signal is used as the
+                # far-end reference for echo cancellation on the mic path.
+                self._webrtcechoprobe.set_property("name", self.AEC_PROBE_NAME)
+                webrtcdsp.set_property("probe", self.AEC_PROBE_NAME)
```
alozowski: If `gst-plugins-bad` isn't installed, `ElementFactory.make` raises `MissingPluginError` and crashes the daemon at startup with no fallback. Should we add a guard with a try/except and fall back to no-AEC?

alozowski:
Works with https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/389

alozowski (approved):

FabienDanieau:
Need to test with a remote daemon (i.e. using the conv app from the mobile app). The AEC should be done at the daemon level in this case.

FabienDanieau:
This is done, with `reachy-mini-daemon --sim` the conversation from the mobile app can be used

alozowski:
tested with the sim daemon and mobile conv app – it works!

Issue #1165, opened 2026-05-29
Reachy Mini Wireless daemon RSS ~1.8 GB on Pi 4 CM4 4GB; central-signaling-relay reconnect storm + glibc arena bloat — mitigated locally


thc1006:
## Summary

On a Reachy Mini Wireless (CM4 4GB) running SDK `reachy_mini==1.7.3`, the daemon's resident set grows to ~1.8 GB after a few hours of uptime and exhausts Pi RAM. We traced this to two compounding causes: (1) `reachy_mini.media.central_signaling_relay` reconnecting on every `RECONNECT_INTERVAL=5s` cycle (~720 cycles/hour, producing ~3755 relay-related log lines/hour) when the HF central rejects the cached token with `Authentication failed - token may be invalid`, and (2) glibc per-thread arena bloat in a process with 56 threads. A local workaround (remove cached HF token + set `MALLOC_ARENA_MAX=2`) drops daemon RSS from 1791 MB to 178 MB (-91%). Verified the relay source is unchanged in `v1.8.0rc1` — bug persists in pre-release. Filing for upstream visibility.

## Environment

- Hardware: Reachy Mini Wireless, CM4 4GB, Debian 13 aarch64
- SDK: `reachy_mini==1.7.3`
- Use case: long-running deployment (elder-care assistant). A separate Python process on the same Pi consumes `mini.media.get_frame()` and drives motors via the daemon REST API.

## What we observed

- Daemon RSS climbed to **1791–2007 MB** after ~4 h uptime (steady-state reached ~1.8 GB; brief spikes to ~2.0 GB).
- Pi free RAM fell below ~200 MB; an external watchdog cycled the brain process roughly every ~15 min as a result.
- Daemon log (rate-counted, 1 h window): ~720 reconnect cycles/hour (matching `RECONNECT_INTERVAL=5.0s`), generating ~3755 relay-related log lines per hour (each cycle emits state transition + SSE handler exit + close + peer-remove lines).
  - Cumulative `_connect_and_relay()` invocations since boot: in the thousands.
  - Each cycle is preceded by `[Central Relay] State transition: connecting -> error | Authentication failed - token may be invalid`. The HF central listener rejects the cached token on every attempt; the relay has no special handling for repeated auth failures, so it loops indefinitely.
- Thread count: 56 in the daemon process.
- `/proc/<PID>/smaps` shows a long tail of `[anon]` mappings each ~64 MB — consistent with many glibc heap arenas.

[...]

PR #1190, opened 2026-06-03
docs(troubleshooting): add Pi 4 4GB daemon memory mitigations

thc1006:
Follows up on #1165, where @pierre-rouanet was happy to take a PR.

This adds a troubleshooting entry for the Wireless robot running out of memory on a 4GB Pi 4. It documents the two mitigations from the issue:

1. Stop `central_signaling_relay` by moving the Hugging Face token aside (only when the App Store remote view is not in use, and the change is reversible).
2. Cap glibc memory arenas with a `MALLOC_ARENA_MAX=2` systemd drop-in on the daemon, including the edit and reload steps.

Measured on one CM4 unit: daemon RSS went from 1791MB to 178MB, available memory from 762MB to 2612MB, and relay reconnects per boot from over 17,000 to zero. The watchdog restart cycle went from roughly 15 minutes to several hours. The note is clear that these are mitigations rather than a root fix.

The entry follows the existing `<details>` format in the file and only touches `docs/source/troubleshooting.md`.

### DIFF
--- a/docs/source/troubleshooting.md
+++ b/docs/source/troubleshooting.md
@@ -909,7 +909,7 @@ If a part is broken/malfunctioning, Pollen's after-sales team will determine if 
 
 On a 4GB CM4 Wireless unit the daemon can climb to around 1.8GB RSS within a few hours, push the Pi into swap, and trigger a watchdog restart loop. Two settings keep it under control.
 
-First, if you are not using the Hugging Face App Store remote view, stop the signaling relay. When the HF signaling endpoint is unreachable (stale token, rate limit, region policy), `central_signaling_relay` retries about once a second, which was seen at over 17,000 reconnects per boot. The relay does not start without a token, so moving it aside disables it. This is reversible: keep the `.bak` and move it back to re-enable the remote view.
+First, if you are not using the Hugging Face App Store remote view, stop the signaling relay. When the HF signaling endpoint is unreachable (stale token, rate limit, region policy), `central_signaling_relay` retries about once a second, which was seen at over 17,000 reconnects per boot. The relay reads the token through `huggingface_hub.get_token()`, so it picks up the cached file or the `HF_TOKEN` environment variable. Moving the cached file aside disables the relay as long as no `HF_TOKEN` is exported for the daemon (check the service environment if the relay keeps reconnecting). This is reversible: keep the `.bak` and move it back to re-enable the remote view.
 
 ```bash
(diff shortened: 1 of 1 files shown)


FabienDanieau:
These are interesting issues thanks! Would you like to push actual fixes?

thc1006:
Thanks @FabienDanieau! I opened #1193 with the code fix for the glibc arena part: it sets `MALLOC_ARENA_MAX=2` in the Wireless `launcher.sh` (and the systemd unit) so the cap is in place before glibc allocates the main arena.

For the relay reconnect part, the current `central_signaling_relay` loop already has `RECONNECT_INTERVAL` and an attempt counter, so I did not want to change it blindly. If the reconnect storm is still reproducible on a recent build I am happy to dig into that separately, but I could not reproduce it cleanly without removing the token workaround on a production unit.

This PR (#1190) can stay as the troubleshooting docs, and #1193 carries the actual arena fix.

pierre-rouanet:
Hi @thc1006 can you test if https://github.com/pollen-robotics/reachy_mini/pull/1201 fix your issue?

thc1006:
Hi @pierre-rouanet — yes, I looked closely. #1201 fixes the exact path behind the storm I reported in #1165: a token that's present but rejected by central (stale / rate-limited) returns cleanly in the `ERROR` state, so the old loop skipped the backoff and reconnected immediately. #1201 routes that `ERROR` case through the normal backoff, which is the correct fix.

My `mv ~/.cache/huggingface/token` workaround sidesteps the same path a different way (no token → `WAITING_FOR_TOKEN`, 30s), so with #1201 in place that workaround shouldn't be needed anymore. So I'm happy to close #1190 once #1201 lands — its content is superseded at that point.

I can't reproduce the storm on my unit right now (it's in production use), so this is a code-level confirmation; glad to run a before/after on hardware when it frees up if you'd like the empirical check on top. Thanks!

thc1006:
Quick status: now that #1201 has merged, the relay-token workaround documented here is superseded. Since this PR also documents the `MALLOC_ARENA_MAX` workaround, I'll close it once #1193 lands too, so both fixes are in code. Feel free to close it sooner if you prefer. Thanks!

thc1006:
Closing this now that both underlying fixes have landed: #1193 (the `MALLOC_ARENA_MAX` cap, now applied directly in the daemon service) and #1201 (the central-relay backoff, which removes the need for the token workaround). The manual steps documented here are superseded by those changes in code. Thanks @FabienDanieau @pierre-rouanet @alozowski!

Issue #1191, opened 2026-06-04
JS apps on HF Spaces can't detect simulation mode


yaseminozkut:
- Needs to pass a signal/variable name to be used from JS apps to detect if the robot is in simulation mode
- Its useful when simulation mode require different things such as laptop camera instead of MuJoCo camera for testing purposes

PR #1192, opened 2026-06-04
feat: add sim transport flag for JS clients

yaseminozkut:
## What

  - #1191 
  - sets `transport="sim"` in addition to `usb` and `wifi` in the signaling meta when daemon runs with simulation mode
  - documents the `transport` field in the JS SDK docs

  ## For JS apps, it can be used as:
  ```js
  const isSim = r.meta?.transport === 'sim';
  ```


pierre-rouanet:
Pinging @tfrere  on this one.

PR #1193, opened 2026-06-04
fix(daemon): cap glibc malloc arenas on Pi 4 Wireless (MALLOC_ARENA_MAX=2)

thc1006:
Follows up on #1165 and #1190, where @FabienDanieau asked for actual fixes. This is the code-level fix for the glibc arena part of the memory growth on the 4GB Pi 4 Wireless unit.

## Problem

The Wireless daemon runs many long-lived threads (GStreamer, WebRTC, asyncio). On a quad core Pi 4, glibc defaults to `8 * ncpu` malloc arenas of up to 64MB each. Under sustained allocation churn the per-arena heaps grow and are not returned to the OS, so daemon RSS climbs over time and can push the 4GB unit into swap, which contributes to the watchdog restart cycle reported in #1165.

## Fix

Set `MALLOC_ARENA_MAX=2` so glibc reads it before the main arena is allocated, kept consistent in the three places it should appear:

- `launcher.sh`: `export MALLOC_ARENA_MAX=2` alongside the existing exports, before `python` starts. glibc reads the variable at process init, before any threads or allocations, so this is the reliable runtime path.
- `install_service.sh`: add `Environment=MALLOC_ARENA_MAX=2` to the systemd unit it generates. The unit is created here via heredoc (not copied from the repo template), so this is where the cap has to go to take effect on a fresh install.
- `reachy-mini-daemon.service`: the same `Environment=` line in the checked-in unit, so the cap is visible in every representation (per review feedback).

A pure in-process `mallopt(M_ARENA_MAX, 2)` was considered but rejected: by the time Python can call it via ctypes the main arena is already allocated, so the env var route is the dependable one.

## Verification

The cap bounds glibc arena RSS for the long-running multithreaded case. In a glibc reproduction with long-lived threads doing sustained mixed-size allocation churn (the daemon's pattern), peak RSS drops from ~66MB uncapped to ~53MB at `MALLOC_ARENA_MAX=2` (about 20%), monotonic with arena count. The effect is specific to long-lived threads — short-lived threads that allocate and exit have their per-thread arenas released anyway, so the cap only matters for a long-running daemon like this one.

To be accurate about #1165: the ~1.8GB → 178MB reduction reported there was the combined effect of two coupled knobs — disabling the runaway central-relay reconnect (the dominant contributor, documented in #1190) and this arena cap (the relay churn is what fed the arenas). This PR is only the arena cap, so it does not on its own account for the full 1.8GB; it is a conservative, reversible bound that complements the relay mitigation.

Conservative, reversible, and scoped to the Wireless platform files. It does not touch the relay reconnect path.

### DIFF
--- a/src/reachy_mini/daemon/app/services/wireless/install_service.sh
+++ b/src/reachy_mini/daemon/app/services/wireless/install_service.sh
@@ -16,6 +16,8 @@ Wants=network-online.target
 
 [Service]
 Type=simple
+# Cap glibc malloc arenas on the 4GB Pi 4 (see issue #1165 and launcher.sh).
+Environment=MALLOC_ARENA_MAX=2
 ExecStart=$LAUNCHER_PATH
 Restart=on-failure
 RestartSec=3s
--- a/src/reachy_mini/daemon/app/services/wireless/launcher.sh
+++ b/src/reachy_mini/daemon/app/services/wireless/launcher.sh
@@ -6,6 +6,11 @@ export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib/aarch64-linux-gnu/
 export LIBCAMERA_IPA_MODULE_PATH=/usr/local/lib/aarch64-linux-gnu/libcamera/ipa
 export LIBCAMERA_IPA_CONFIG_PATH=/usr/local/share/libcamera/ipa
 
+# Cap glibc malloc arenas. On a quad core Pi 4 the default (8 x ncpu) lets the
+# multithreaded daemon spread allocations across many 64MB arenas, which grows
+# RSS over time on the 4GB Wireless unit. 2 keeps it bounded. See issue #1165.
+export MALLOC_ARENA_MAX=2
+
 # Ensure WiFi is not soft-blocked (can happen after a crash or kernel module reload)
 sudo rfkill unblock wifi
 
--- a/src/reachy_mini/daemon/app/services/wireless/reachy-mini-daemon.service
+++ b/src/reachy_mini/daemon/app/services/wireless/reachy-mini-daemon.service
@@ -5,6 +5,8 @@ Wants=network-online.target
 
 [Service]
 Type=simple
+# Cap glibc malloc arenas on the 4GB Pi 4 (see issue #1165 and launcher.sh).
+Environment=MALLOC_ARENA_MAX=2
 ExecStart=/venvs/mini_daemon/lib/python3.12/site-packages/reachy_mini/daemon/app/services/wireless/launcher.sh
 Restart=on-failure
 RestartSec=3s
--- a/src/reachy_mini/daemon/app/services/wireless/reachy-mini-daemon.service
+++ b/src/reachy_mini/daemon/app/services/wireless/reachy-mini-daemon.service
@@ -5,8 +5,6 @@ Wants=network-online.target
 
 [Service]
 Type=simple
-# Cap glibc malloc arenas on the 4GB Pi 4 (see issue #1165 and launcher.sh).
-Environment=MALLOC_ARENA_MAX=2
 ExecStart=/venvs/mini_daemon/lib/python3.12/site-packages/reachy_mini/daemon/app/services/wireless/launcher.sh
 Restart=on-failure
 RestartSec=3s


thc1006:
Self-review update: I initially also edited `reachy-mini-daemon.service`, but that template is not the install source (the unit is generated by `install_service.sh` via heredoc, with a different `User`/`WorkingDirectory`). I reverted the template and moved the `Environment=MALLOC_ARENA_MAX=2` into `install_service.sh` so it actually lands on a fresh install. The PR now only touches `launcher.sh` and `install_service.sh`.

pierre-rouanet (changes requested):
Could you also add your changes to the service file. It's cleaner if the change is visible everywhere.

Also maybe @FabienDanieau we should to latency test within this context to measure impact?

thc1006:
Thanks a lot for the review, @pierre-rouanet, and for the suggestion.

Done: I've added `Environment=MALLOC_ARENA_MAX=2` (with the same comment) to `reachy-mini-daemon.service` as well, so the cap is now consistent across `launcher.sh`, the unit generated by `install_service.sh`, and the checked-in service file.

On measuring the impact — thank you for raising it, I'd been wanting to characterise it properly. I ran a small glibc reproduction (x86 container, so directional rather than Pi-absolute):

- With **long-lived threads** doing sustained mixed-size allocation churn (the daemon's pattern), peak RSS went from ~66MB uncapped to ~53MB at `MALLOC_ARENA_MAX=2` — about 20%, and monotonic with arena count (cap=4 landed in between).
- With **short-lived threads** that allocate and exit there's no benefit, since glibc releases the per-thread arenas on exit. So the win is specifically the long-running multithreaded case, which is what the daemon is.

On latency: it does trade a little cross-thread malloc concurrency for that bound. For an I/O-bound daemon (GStreamer / WebRTC / asyncio) I'd expect the effect to be small, but I haven't measured latency yet, so I'd rather not put a number on it. If it would help, I'd be very happy to run a before/after on the actual Pi 4 Wireless — for example WebRTC round-trip or command-to-motion — and post the numbers here. Please just let me know which metric would be most convincing for you and @FabienDanieau.

One thing I wanted to be transparent about: the ~1.8GB → 178MB figure from #1165 was actually the combined effect of two coupled knobs — disabling the runaway central-relay reconnect (the dominant one, documented in #1190) and this arena cap. Since this PR is only the cap, I didn't want it to read as if the cap alone produced that result, so I've tightened the PR description accordingly.

Thanks again for taking the time to look at this — much appreciated.

pierre-rouanet:
@FabienDanieau got a procedure to measure video latency, maybe he can tell you more. 
Maybe we should also have more qualitative measure of some apps (the conversation app at least) to see if they behave the same.

thc1006:
Thank you, @pierre-rouanet — that sounds great.

@FabienDanieau, whenever it's convenient, I'd really appreciate it if you could share your video-latency procedure. I'd like to follow your method exactly so the numbers are comparable to yours, then I'll run it on the Pi 4 Wireless with the cap on vs off and post the results here.

On the qualitative side: agreed, that's the right check. For context, the daemon has already been running with `MALLOC_ARENA_MAX=2` in our setup and the conversation flow (STT → LLM → TTS → motion) has behaved the same as before — but I'll do a proper A/B rather than rely on that impression. I have a small end-to-end conversation harness I can run with the cap on and off to compare both behaviour and response time, and I'll post the before/after here alongside the video-latency numbers.

Please let me know if there's a particular app or scenario you'd most like me to cover. Thanks again for the guidance.

thc1006:
Thanks @pierre-rouanet. @FabienDanieau — I'd love to use your procedure; could you share it whenever convenient? I'll follow your method exactly so the numbers line up with yours.

In the meantime I checked the robot so I'm ready to run it: it's GStreamer 1.26.2 with the core tracers available (`latency`, `leaks`, …). GstShark / `interlatency` isn't installed, but I'm happy to add it if you'd prefer per-element attribution. A procedure I can run if it's useful:

- **Intra-pipeline**: the built-in `latency` tracer (`GST_TRACERS=latency`) for per-element / element-to-sink timing inside the daemon pipeline — no rebuild needed.
- **True glass-to-glass**: an external method (camera filming an on-screen millisecond timer, comparing captured vs displayed timestamps) for the real end-to-end figure the tracer can't give on its own.

I'd run both with the conversation app (brain) **on**, so the daemon is under realistic load, comparing `MALLOC_ARENA_MAX=2` vs unset. (A quick check I ran earlier happened to have that load off, so it wasn't discriminative — I'll make sure to fix that.)

That said, I'm very happy to just defer to your procedure — whatever is easiest on your side. Thanks again!

FabienDanieau:
We have a custom board to measure the glass to glass latency (https://github.com/pollen-robotics/G2GDelay, but you can make one with an arduino (see original repo)).
Now this may be overkill for your tests. When you start the daemon the gstreamer latency is reported. It's a good first metric. Then you may use the tracers indeed to check that you don't add a significant latency compared to what we have now.

thc1006:
Thank you @FabienDanieau — that was exactly the right pointer, and good to know the G2GDelay board is there if we ever need true glass-to-glass.

I ran your first metric on the actual Pi 4 Wireless (with the brain on, so the media pipeline is under realistic load), comparing the daemon's reported GStreamer pipeline latency with `MALLOC_ARENA_MAX=2` vs unset (daemon + brain restarted each round, MainPID-verified):

| | min_latency | max_latency |
|---|---|---|
| cap unset | 84 ms | 1024 ms |
| `MALLOC_ARENA_MAX=2` | 84 ms | 1026 ms |

So the arena cap adds no pipeline latency — `min_latency` is identical at 84 ms, and `max_latency` is within run-to-run noise (~2 ms). This is consistent with a separate request-latency A/B I ran earlier on `/api/daemon/status` (controlled for warm-up and for brain load): no measurable difference between cap on and off.

I'm happy to also run the `latency` tracer for per-element attribution if you'd like that extra confirmation — just let me know. Thanks again for the clear guidance!

FabienDanieau (approved):
I don't see any issue with my tests. We can merge this. Thanks!

pierre-rouanet (approved):

PR #1197, opened 2026-06-08
Clean up python dependencies

### LINKED ISSUE #1196: Clean up Python dependency metadata
`pyproject.toml` has a few dependency metadata issues: 
- `requests` is declared twice, 
- `pydantic` is used directly but only pulled transitively, 
- `tornado` is installed as a core dependency even though it is only needed through the `placo`/`meshcat` path

Also, let's unpin `huggingface-hub`

alozowski:
Cleanes up `pyproject.toml` dependency declarations and refreshes `uv.lock`. Closes https://github.com/pollen-robotics/reachy_mini/issues/1196

This makes `pydantic` explicit, deduplicates `request`, and moves the `tornado` security floor to the `placo_kinematics` extra where it is actually needed. Also unpins `huggingface-hub`


pyproject.toml
```diff
@@ -34,14 +35,11 @@ dependencies = [
...
     "rustypot>=1.4.2",
     "pyyaml>=6.0",
-    "requests>=2.28.0",
+    "requests>=2.33.0",
     "zeroconf>=0.131,<1",
     "pycaw>=20240210; sys_platform == 'win32'",
     "pulsectl>=24.1.0; sys_platform == 'linux'",
 
-    "tornado>=6.5.5",   # fix security issue of this subdependency
-    "requests>=2.33.0", # fix security issue of this subdependency
```
andimarafioti, line 43: is tornado no longer necessary? there was that security issue

FabienDanieau:
You can also bump dependencies to fix [security issues](https://github.com/pollen-robotics/reachy_mini/security/dependabot). (and run tests to make sure that there is no regression)

pyproject.toml
alozowski: we don't need to declare `tornado` for the install to work, because `meshcat` (via placo) already pulls it in transitively, but I suggest to keep it with this `>=6.5.5` floor purely as a security guardrail, because `meshcat` itself doesn't constrain the version and a future resolution could otherwise land on an older, vulnerable release

alozowski:
@FabienDanieau I bumped up packages' versions per dependabot, CI is green

FabienDanieau (approved):
Validated by the physical CI.

PR #1199, opened 2026-06-09
feat(host): session-topbar robot identity with live latency + Lite/Wireless tags

tfrere:
## Summary

Brings the host shell's in-session chrome in line with the mobile app, and makes the dev harness usable against a real stream. Three focused commits:

1. **Shared identity pill primitives + picker tag** - port the mobile shell's identity-row primitives into `ts/host/src/ui/design` (`TransportChip`, `MetaPill`/`TagLabel`/`VariantTag`, `LinkQualityBars` + a self-contained `linkQualityLevel`, and a trimmed `IdentityChipBar`). `PickerView` now shows the product `VariantTag` (`Lite`/`Wireless`) instead of the raw USB/Wi-Fi chip. (Assumed duplication, same stance as `lib/tokens.ts`, since the host is a separate package.)
2. **Topbar robot identity + live latency** - in session the topbar reads like the mobile overlay: running app as the headline, with a secondary sub-line carrying the robot name, its `Lite`/`Wireless` tag and a live latency pill.
   - Robot identity (name + transport) is **snapshotted at selection time**, because `useRobots` is disabled the moment we leave the picker (to protect the 1:1 `token -> peer` handoff invariant) and clears its list - so it can't be looked up during the session (this was surfacing the peer-id hash instead of the name).
   - Latency is a **true live value**: the embed measures RTT on its own WebRTC candidate pair (the host released its slot on handoff) and reports it via an additive `rttMs` field on `embed:app-state`. `embed/index.ts` samples a rolling-min RTT once live, `useHostBridge` forwards it, and the topbar shows the pill only when a value is present (never a frozen number).
3. **Dev harness live camera** - the minimal, dependency-free harness embed now renders the robot's live camera (via `media.attachVideo`) instead of a text-only status, so the topbar work can be exercised against a real stream. Auto-wake on connect and runtime sleep + torque release on leave are unchanged.

## Notes

- Additive protocol change only (`rttMs?` on `embed:app-state`), no version bump; hosts that don't care (the standalone web shell, which measures its own link) simply ignore it.
- No persona avatars on the host: personality profiles are not stored on the robot, so the host can't know them.

## Test plan

- [ ] `npm run typecheck:host` passes (verified locally).
- [ ] Preview chrome: `?host-preview=connecting` shows app headline + robot sub-line (name, Wireless tag, latency pill).
- [ ] Real session: robot **name** (not the peer-id hash) shows in the topbar; `Lite`/`Wireless` tag appears when central advertises `meta.transport`; latency pill appears a few seconds after going live.
- [ ] Dev harness embed shows the live camera; End-session tears down cleanly.

### DIFF
--- a/ts/host/dev/main.ts
+++ b/ts/host/dev/main.ts
@@ -86,11 +86,19 @@ async function bootEmbed(): Promise<void> {
   const { connectToHost } = await import('@/embed');
 
   const root = document.getElementById('root') ?? document.body;
+  // Still intentionally minimal, but now it FEELS like a real app:
+  // the robot's live camera fills the view instead of a text stub.
+  // Everything else (auto-wake on connect, runtime sleep + torque
+  // release on leave) is handled by `connectToHost()` / the leave
+  // contract - this page only has to render the stream the SDK
+  // already negotiated.
   root.innerHTML = `
     <main style="font-family:system-ui,sans-serif;display:flex;flex-direction:column;
                  align-items:center;justify-content:center;height:100%;gap:16px;padding:24px;
-                 text-align:center;color:#111">
-      <h1 style="margin:0;font-size:1.25rem">Minimal embed</h1>
+                 text-align:center;color:#111;box-sizing:border-box">
+      <video id="robot-video" autoplay playsinline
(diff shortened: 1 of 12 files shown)


pierre-rouanet (approved):
Tested, works fine!

PR #1201, opened 2026-06-10
Back off central relay reconnect loop on 401 from stale HF token

### LINKED ISSUE #1198: Investigate long usage (>5H+)
- [x] Run low level rusty pot soak test - https://github.com/pollen-robotics/rustypot/blob/master/examples/soak_test.rs
- [x] Run Reachy-mini-motor-controller soak test

alozowski:
Addresses the stale-token reconnect loop suspected in https://github.com/pollen-robotics/reachy_mini/issues/1198

When central rejects a stored HF token with 401, for example, stale/invalid token, the relay now applies the same backoff used for connection failures instead of retrying immediately, preventing log spam and connection churn against the local GStreamer signaling server during long-running sessions.

### DIFF
--- a/src/reachy_mini/media/central_signaling_relay.py
+++ b/src/reachy_mini/media/central_signaling_relay.py
@@ -687,8 +687,7 @@ class CentralSignalingRelay:
                         )
 
                 if self._running and not had_exception and self._state == RelayState.ERROR:
-                    # Connect returned cleanly but left us in ERROR (e.g. a 401
-                    # from a stale token) - back off like a connection failure.
+                    # Clean return but ERROR (e.g. 401) - back off like a failure.
                     had_exception = True
                     self._connection_attempts += 1
 
--- a/src/reachy_mini/media/central_signaling_relay.py
+++ b/src/reachy_mini/media/central_signaling_relay.py
@@ -686,6 +686,12 @@ class CentralSignalingRelay:
                             f"Connection failed after {self._connection_attempts} attempts: {e}",
                         )
 
+                if self._running and not had_exception and self._state == RelayState.ERROR:
+                    # Connect returned cleanly but left us in ERROR (e.g. a 401
+                    # from a stale token) - back off like a connection failure.
+                    had_exception = True
+                    self._connection_attempts += 1
+
                 if self._running and had_exception:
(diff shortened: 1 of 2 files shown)


FabienDanieau:
related to #1165

alozowski:
In this case I will close this my PR once https://github.com/pollen-robotics/reachy_mini/issues/1165 is finished and merged

thc1006:
From the production side — this is the 401-immediate-retry path behind the relay churn in #1165 on a CM4 4GB Wireless unit, and this is the right fix. A stale-but-present token returning cleanly in the `ERROR` state was exactly what made the old loop skip the backoff and reconnect immediately; routing it through the existing backoff addresses the root cause, and it means the `mv token` workaround is no longer required. Nice and minimal.

One optional follow-up, only if the >5h soak in #1198 shows residual growth: the backoff here is the fixed `RECONNECT_INTERVAL` (5s), so a persistently-rejected token still reconnects ~720×/hour. If that still adds up over long sessions, a capped exponential backoff (and/or pausing retries until `update_token()` fires) would take it close to zero. Happy to contribute that as a follow-up if the soak data warrants it — but this PR already removes the worst of it.

alozowski:
Hello @thc1006,

Sounds good, let's get this PR merged on its own first, what do you think, @FabienDanieau?
Plus, if the >5h soak in #1198 still shows residual reconnects after this PR + #1193 are merged, a follow-up PR adding exponential backoff (capped) and/or an auth-failure circuit breaker would be very welcome 🙂

FabienDanieau (approved):
LGTM.
>  a follow-up PR adding exponential backoff (capped) and/or an auth-failure circuit breaker would be very welcome 
agree 

@thc1006 do you want to do it? thanks!

thc1006:
Thanks @FabienDanieau — yes, I'm happy to do it. Let's let #1193 land and then watch a soak with both fixes in place: if it still shows residual reconnects from a persistently-rejected token, I'll open the follow-up with the capped exponential backoff and the auth-failure circuit breaker. If the fixed 5s backoff already flattens it on its own, we may not even need it — either way I'll keep an eye on it once #1193 is in. Thanks, and nice work on this fix @alozowski!

PR #1195, opened 2026-06-08
feat(media): harden conversation audio against Wi-Fi packet loss & jitter

tfrere:
## Summary

Make the phone-brokered conversation audio resilient to Wi-Fi loss and jitter, on **both** legs of the robot's WebRTC media path. No plugin forks, no internals: only standard GStreamer element properties and the supported `webrtcsink` `encoder-setup` extension point.

### Context

The conversation relays audio through the robot's `webrtcsink` (producer) and a per-consumer playback pipeline. Two GStreamer defaults made this fragile on a jittery Wi-Fi link:

- **Phone -> robot (assistant voice on the speaker):** the playback `opusdec` ran with `use-inband-fec=False` and `plc=False`, so any dropped packet glitched the speaker even though the browser encoder ships redundancy; and the consumer `webrtcbin` used the default 200ms jitter buffer, which underruns under jitter spikes.
- **Robot mic -> phone (audio that feeds the realtime backend / STT):** `webrtcsink` builds its `opusenc` with defaults (`inband-fec=False`, `packet-loss-percentage=0`), so a dropped mic packet could only be concealed by the browser decoder, never reconstructed. This degrades recognition on bad Wi-Fi.

### Changes (`media/media_server.py`)

- **Receive (`opusdec`)**: enable `use-inband-fec` + `plc` so the decoder reconstructs/conceals lost packets (~one frame of look-ahead).
- **Receive jitter buffer**: raise the consumer `webrtcbin` `latency` to `300ms` (`RX_JITTER_LATENCY_MS`), capped there because end-to-end latency is the metric we watch closest.
- **Send (`opusenc`)**: hook `webrtcsink`'s `encoder-setup` signal to set `inband-fec=True` + `packet-loss-percentage=20` (`TX_OPUS_FEC_LOSS_PERC`) on the mic encoder. Returns `False` so webrtcsink keeps its congestion-controlled bitrate; `find_property` guards keep it safe across GStreamer versions.

### Test plan

- [x] Syntax / AST checks pass.
- [x] Deployed on a wireless Reachy Mini (1.8.0), daemon restarts cleanly, no GStreamer property errors at startup.
- [x] Phone -> robot: assistant voice no longer stutters on a jittery Wi-Fi link (confirmed by ear).
- [ ] Robot mic -> phone: confirm `opusenc tuned ... inband-fec=True, packet-loss-percentage=20` is logged on consumer connect, and recognition holds up under induced loss.
[...]

PR #1202, opened 2026-06-10
test(media): jitter buffer & Opus FEC resilience measurement harnesses

tfrere:
## Context

Follow-up to #1195 (`harden conversation audio against Wi-Fi packet loss & jitter`, already merged). That PR raised the RX jitter buffer to 300 ms and enabled Opus in-band FEC on both audio legs. The review asked: *did you compare jitter buffer vs FEC?*

This PR adds the **standalone measurement harnesses** I used to characterise that tuning, plus a README documenting methodology, results and limitations. They live in `tests/integration_tests/media_resilience/`, rebuild a representative WebRTC receive path with the same GStreamer elements, and **do not touch the running daemon**.

## TL;DR

- **Jitter buffer vs FEC is not an either/or** - they fix different problems (delay-absorption vs loss-recovery). Keeping both is correct.
- **The GStreamer `rtpjitterbuffer` is static, not adaptive.** Measured start delay tracks the configured depth almost exactly even on a zero-jitter link, so 200 -> 300 ms is a ~100 ms permanent latency cost on the phone -> robot leg, not something that shrinks on good Wi-Fi.

  | latency config | measured start delay |
  |---:|---:|
  | 0 ms | 1.5 ms |
  | 40 ms | 36.8 ms |
  | 200 ms (default) | 197.4 ms |
  | 300 ms (PR #1195) | 296.3 ms |

- **FEC recovery quality could not be quantified objectively offline.** Two independent metrics (waveform SNR, per-frame dominant frequency) are both confounded by the same root cause: concealment (PLC and FEC alike) does not preserve exact sample phase/timeline, and `opusdec use-inband-fec` adds a variable 1-frame shift only when it actually uses FEC. A trustworthy number needs a perceptual metric (PESQ/ViSQOL) on speech, or instrumenting `opusdec` to count real `decode_fec` reconstructions. Enabling FEC still follows the WebRTC recommendation (RFC 8854) and was validated subjectively in #1195.

## Files

| File | Status |
|------|--------|
| `measure_jitter_fec.py` | conclusive (latency) |
| `measure_fec_quality.py` | exploratory (confounded), kept to document the dead end |
| `measure_fec_spectral.py` | exploratory (confounded), kept to document the dead end |
| `README.md` | methodology + results + limitations |

cc @FabienDanieau

### DIFF
--- a/tests/integration_tests/media_resilience/README.md
+++ b/tests/integration_tests/media_resilience/README.md
@@ -1,0 +1,119 @@
+# Media resilience measurements (jitter buffer & Opus FEC)
+
+Experimental harnesses used to characterise the Wi-Fi resilience tuning added
+to the WebRTC media server in PR #1195
+(`feat(media): harden conversation audio against Wi-Fi packet loss & jitter`).
+
+That PR added, in `src/reachy_mini/media/media_server.py`:
+
+| Knob | Leg | What |
+|------|-----|------|
+| `RX_JITTER_LATENCY_MS = 300` | phone → robot | consumer `webrtcbin` jitter buffer depth |
+| `opusdec use-inband-fec=True, plc=True` | phone → robot | decoder loss recovery |
+| `opusenc inband-fec=True, packet-loss-percentage=20` | robot mic → phone | encoder loss resilience |
+
+These scripts answer the review question *"did you compare jitter buffer vs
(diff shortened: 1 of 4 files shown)


FabienDanieau:
Thanks I've copied this into our notion.

PR #1205, opened 2026-06-11
daemon: bind loopback off-wireless; remove dead --localhost-only flag

FabienDanieau:
The --localhost-only flag never affected the socket bind. uvicorn always bound args.fastapi_host (default 0.0.0.0), and the flag only gated whether self._status.wlan_ip was populated, so the Lite/desktop/simulation daemons exposed the unauthenticated HTTP API on all interfaces despite the flag defaulting to "localhost only".

Derive the bind address from wireless_version instead: bind 0.0.0.0 only on the wireless version (which must be reachable over the LAN) and 127.0.0.1 everywhere else. An explicit --fastapi-host still overrides. Remove the misleading, unused --localhost-only/--no-localhost-only flags and the localhost_only parameter threaded through Daemon.start()/restart().

### DIFF
--- a/src/reachy_mini/daemon/app/main.py
+++ b/src/reachy_mini/daemon/app/main.py
@@ -90,19 +90,27 @@ class Args:
 
     robot_name: str = "reachy_mini"
 
-    fastapi_host: str = "0.0.0.0"
+    # None means "auto": bind 0.0.0.0 on the wireless version (must be reachable
+    # over Wi-Fi) and 127.0.0.1 everywhere else. See _resolve_bind_host().
+    fastapi_host: str | None = None
     fastapi_port: int = 8000
 
-    localhost_only: bool | None = None
+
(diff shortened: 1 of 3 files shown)


pierre-rouanet (commented):
Overall change looks good!

But shall we entirely remove the --localhost-only argument or simply deprecated it and add a warning to say that the argument will be ignored?
Right now removing it break the physical ci (this we can easily fix), but I'm not sure if this may break other things.
What do you think?

FabienDanieau:
> Overall change looks good!
> 
> But shall we entirely remove the --localhost-only argument or simply deprecated it and add a warning to say that the argument will be ignored? Right now removing it break the physical ci (this we can easily fix), but I'm not sure if this may break other things. What do you think?

I'd prefer to update the physical CI. I don't think it was used elsewhere

pierre-rouanet (approved):

PR #1208, opened 2026-06-12
feat(daemon): add start_update WebRTC data channel command

tfrere:
## Summary

Adds a `start_update` command to the typed WebRTC data channel so Central-routed (remote) peers can trigger a daemon PyPI update without an LAN HTTP path. Today the update is only reachable via the REST `POST /update/start` endpoint on the robot's port 8000 (LAN-only); the data channel exposed `restart_daemon` but no update.

This mirrors the existing `restart_daemon` plumbing end to end.

## Changes

- **`io/protocol.py`**: new `StartUpdateCmd` (`type: "start_update"`, optional `pre_release`), added to the `AnyCommand` union.
- **`daemon/backend/abstract.py`**: `_start_update_callback` slot, `set_start_update_callback()`, and a handler in `_handle_command` that acks then fires the callback (errors if no host wired it).
- **`daemon/daemon.py`**: `_spawn_webrtc_update()` wired via `set_start_update_callback`; runs `update_reachy_mini` on a background thread (same pattern as `_spawn_webrtc_restart` / REST `/update/start`).
- **`ts/lib/reachy-mini.ts`**: `startDaemonUpdate({ preRelease })` SDK method.

## Semantics

Fire-and-ack, identical to `restart_daemon`: the update ends with a `systemctl restart` that tears the transport down, so the daemon sends a single ack immediately and the consumer reconnects afterwards. No progress is streamed on this channel; consumers that want logs can pair it with `subscribe_logs`.

## Test plan

- [ ] Send `{ "type": "start_update" }` over the data channel and confirm the `{"status":"ok","command":"start_update"}` ack, then the daemon updates + restarts.
- [ ] Confirm `pre_release: true` installs a pre-release.
- [ ] Confirm a backend host without the callback wired returns the `start_update not supported` error.
- [ ] Confirm reconnection after the post-update restart.

### DIFF
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -334,8 +334,11 @@ class Backend:
         # followed by a restart. Wired in by `Daemon`, same fire-and-ack
         # contract as `_restart_daemon_callback`: the update ends with a
         # `systemctl restart` that tears the transport down, so the
-        # callback must return promptly (it spawns its own thread).
-        self._start_update_callback: Optional[Callable[[bool], None]] = None
+        # callback must return promptly (it spawns its own thread). It runs
+        # cheap pre-checks (wireless robot, update available, not already
+        # running) synchronously and returns a refusal reason string when it
+        # declines, or ``None`` once the update job has been accepted.
+        self._start_update_callback: Optional[Callable[[bool], Optional[str]]] = None
 
     # Life cycle methods
     def wrapped_run(self) -> None:
@@ -1394,9 +1397,14 @@ class Backend:
                 self.logger.error(f"restart_daemon callback failed: {e}")
 
         elif isinstance(cmd, StartUpdateCmd):
-            # Same fire-and-ack contract as `restart_daemon`: the update
-            # ends with a `systemctl restart` that tears this transport
-            # down, so ack first and let the consumer reconnect.
+            # Same fire-and-ack contract as `restart_daemon`: a successful
+            # update ends with a `systemctl restart` that tears this
+            # transport down. The callback runs cheap pre-checks (wireless
+            # robot, an update is available, none already running)
+            # synchronously and returns a refusal reason if it declined; we
+            # only ack ok once the update job has actually been accepted, so
+            # the consumer can surface a real error instead of waiting for a
+            # reconnect that never comes.
             if self._start_update_callback is None:
                 send_response(
                     {
@@ -1405,11 +1413,18 @@ class Backend:
                     }
                 )
                 return
-            send_response({"status": "ok", "command": "start_update"})
             try:
-                self._start_update_callback(cmd.pre_release)
+                refusal = self._start_update_callback(cmd.pre_release)
             except Exception as e:
                 self.logger.error(f"start_update callback failed: {e}")
+                send_response(
+                    {"error": f"start_update failed: {e}", "command": "start_update"}
+                )
+                return
+            if refusal is not None:
+                send_response({"error": refusal, "command": "start_update"})
+                return
+            send_response({"status": "ok", "command": "start_update"})
 
         elif isinstance(cmd, UploadMoveStartCmd):
             self._handle_upload_start(cmd)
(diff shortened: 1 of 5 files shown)


src/reachy_mini/daemon/daemon.py
```diff
@@ -559,6 +561,66 @@ def _run() -> None:
 
         Thread(target=_run, daemon=True, name="webrtc-restart-daemon").start()
 
+    def _spawn_webrtc_update(self, pre_release: bool = False) -> None:
```
pierre-rouanet: We should make sure we acquire the lock from the update router: https://github.com/pollen-robotics/reachy_mini/blob/main/src/reachy_mini/daemon/app/routers/update.py#L23
Otherwise, you could start an update via webrtc and start another one in the middle via HTTP which will most likely corrupt the venv.

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -1385,6 +1393,24 @@ def _maybe_ignore(field: str) -> bool:
             except Exception as e:
                 self.logger.error(f"restart_daemon callback failed: {e}")
 
+        elif isinstance(cmd, StartUpdateCmd):
```
pierre-rouanet, line 1399: We should first check if there is an update available. Otherwise the update will force a reinstall of the same version.

src/reachy_mini/io/protocol.py
```diff
@@ -378,6 +379,35 @@ class RestartDaemonCmd(BaseModel):
...
+# Remote counterpart of ``POST /update/start`` (``routers/update.py``):
+# upgrades the ``reachy_mini`` package in the daemon venv from PyPI.
+# Exposed over the typed transport so Central-routed peers can trigger
+# an update without an LAN HTTP path.
+#
+# Like ``restart_daemon``, this is fire-and-ack: the update ends with a
+# ``systemctl restart`` that tears the transport down, so the daemon
+# sends a single ack (``{"status": "ok", "command": "start_update"}``)
+# immediately and the consumer is expected to reconnect once the daemon
+# is back up. No completion / log message is streamed over this channel;
```
pierre-rouanet: You have update log messages no?

src/reachy_mini/daemon/backend/abstract.py
pierre-rouanet, line 1399: We should also check if we are a wireless daemon. Otherwise, if we are a lite, we should return an error code directly.

pierre-rouanet (changes requested):

src/reachy_mini/daemon/daemon.py
tfrere: Good point - fixed in f17d6f08. `_spawn_webrtc_update` now imports and acquires the shared `busy_lock` from `routers/update.py` (non-blocking, so it rejects immediately if one is already held). The worker thread owns the lock and releases it; a successful update restarts the process before release. A WebRTC update and a concurrent REST update can no longer run at the same time.

src/reachy_mini/daemon/backend/abstract.py
tfrere: Fixed in f17d6f08. The callback now calls `is_update_available("reachy_mini", pre_release)` before accepting and rejects with a `No update available` error ack instead of forcing a needless reinstall + restart of the running version (mirrors the REST endpoint's check).

src/reachy_mini/io/protocol.py
tfrere: Good catch - that comment was stale from the first commit. The follow-up commit (cc776db7) does stream the job's log lines as `update_progress` broadcasts (one per line, see `UpdateProgressMsg`). I've updated the `StartUpdateCmd` comment/docstring in f17d6f08 to match, so it no longer claims nothing is streamed.

src/reachy_mini/daemon/backend/abstract.py
tfrere: Fixed in f17d6f08. `_spawn_webrtc_update` now returns early with a `start_update is only supported on Reachy Mini Wireless` error ack when `not self.wireless_version`, so a Lite gets a direct error instead of failing deep inside `update_reachy_mini`. The callback returns the refusal reason so the backend acks the error rather than `ok`.

pierre-rouanet (approved):

FabienDanieau:
Works, we can merge. But there are a lot of comments in this PR. Please keep it minimal. Longs comments go to the function docstring, avoid repetition, make minimal inline comments.

PR #1209, opened 2026-06-12
Fix media sounds upload validation

FabienDanieau:
## What

The `POST /api/media/sounds/upload` endpoint accepted any file with no
extension, content, or size validation, letting an unauthenticated client
write arbitrary files (e.g. shell scripts) into `/tmp/reachy_mini_sounds/`.
This addresses GHSA-m2pc-3q4q-w6jr (CWE-434, medium).

## Changes

- **Extension allow-list** — only `wav/mp3/ogg/oga/opus/flac/m4a/aac`.
- **Size cap** (25 MiB), streamed to a temp file in bounded chunks instead of
  reading the whole body into memory.
- **Content validation** via GStreamer's pbutils discoverer — the same
  machinery that plays the file, so "accepted ⇒ playable" — requiring at least
  one audio stream. Rejects mislabelled/non-audio payloads.
- Validated file is moved into place atomically; partial/invalid uploads are
  cleaned up.

Shared helpers `is_valid_audio_file()` / `audio_duration_seconds()` now live in
`media/gstreamer_utils.py`. The `sound_play` example switched from `--wav` to
`--file` (any container) and the docs were updated to match.

## Testing

New `tests/unit_tests/test_media_upload_api.py` (7 tests): the `.sh` PoC and
other bad extensions are rejected, oversized uploads return 413, path traversal
is stripped, valid audio is stored, and a script disguised as `.wav` fails
content validation.

### DIFF
--- a/src/reachy_mini/daemon/app/routers/media.py
+++ b/src/reachy_mini/daemon/app/routers/media.py
@@ -8,6 +8,7 @@ so that WebRTC clients can upload, play, list and delete sound files on
 the daemon.
 """
 
+import asyncio
 import os
 import tempfile
 from pathlib import Path
@@ -203,7 +204,8 @@ async def upload_sound(
             while chunk := await file.read(1 << 20):
                 f.write(chunk)
 
-        if not is_valid_audio_file(tmp_path):
+        # Offload the blocking GStreamer probe (up to 5 s) off the event loop.
+        if not await asyncio.to_thread(is_valid_audio_file, tmp_path):
             raise HTTPException(
                 status_code=400, detail="Unsupported or invalid audio file"
             )
--- a/src/reachy_mini/daemon/app/routers/media.py
+++ b/src/reachy_mini/daemon/app/routers/media.py
@@ -195,23 +195,12 @@ async def upload_sound(
     os.makedirs(SOUNDS_TMP_DIR, exist_ok=True)
     dest = os.path.join(SOUNDS_TMP_DIR, filename)
 
-    # Stream to a temp file in the same directory, enforcing the size cap, then
-    # validate the content before atomically moving it into place. This avoids
-    # buffering the whole body in memory and never leaves a partial or invalid
-    # file at the public destination name.
+    # Probe a temp copy, then atomically move it into place so no partial or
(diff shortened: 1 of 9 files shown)


src/reachy_mini/daemon/app/routers/media.py
```diff
@@ -165,12 +182,48 @@ async def upload_sound(
...
+            ),
+        )
+
     os.makedirs(SOUNDS_TMP_DIR, exist_ok=True)
     dest = os.path.join(SOUNDS_TMP_DIR, filename)
 
-    content = await file.read()
-    with open(dest, "wb") as f:
-        f.write(content)
+    # Stream to a temp file in the same directory, enforcing the size cap, then
```
pierre-rouanet: Whole file is still download entirely no matter its size, no?
Can we add some enforcing directly in the route itself? Something such as https://github.com/fastapi/fastapi/issues/362?

src/reachy_mini/media/gstreamer_utils.py
```diff
@@ -73,3 +74,54 @@ def get_sample(appsink: GstApp.AppSink, logger: logging.Logger) -> Optional[byte
...
+
+    Returns:
+        ``True`` if the file is a decodable media file with at least one audio
+        stream, ``False`` otherwise.
+
+    """
+    Gst.init([])  # idempotent
+    try:
+        discoverer = GstPbutils.Discoverer.new(5 * Gst.SECOND)
+        info = discoverer.discover_uri(Gst.filename_to_uri(path))
```
pierre-rouanet, line 100: Claude tells me this can block for up to 5s. I'm not sure but if it's the case, should we make it run in a thread not to block the whole async loop?

src/reachy_mini/daemon/app/routers/media.py
FabienDanieau: done

src/reachy_mini/media/gstreamer_utils.py
FabienDanieau: yes, moved to asyncio.to_thread

pierre-rouanet (approved):

PR #1219, opened 2026-06-17
feat(daemon): robot onboarding commands - first wake-up, rename, BLE identify

tfrere:
## Summary

Adds three robot-onboarding/identity commands to the daemon, exposed over the
WebRTC data channel (and one over BLE), so the mobile/desktop apps can run a
first-time setup flow and let a user manage and identify their robot.

1. **First wake-up wizard flag** - a persisted, robot-wide boolean so the
   post-connection setup wizard ("first wake-up") shows exactly once per robot,
   whichever client connects.
2. **Robot rename** - a persisted, robot-wide display name that overrides the
   `--robot-name` CLI default at the next daemon start (advertised over mDNS /
   central relay, shown in the apps' robot list).
3. **BLE identify** - a public BLE command that plays a short sound on the
   robot's speaker so a user can tell which physical Reachy maps to an entry in
   the BLE scan list, *before* connecting.

### Data-channel commands (`io/protocol.py`, `daemon/backend/abstract.py`)

- `get_first_wake_up` -> `{ "command", "is_completed" }`
- `set_first_wake_up` (`is_completed: bool = true`) -> `{ "command", "status", "is_completed" }`
- `get_robot_name` -> `{ "command", "name" }` (`name` is null when unset)
- `set_robot_name` (`name: str`, 1..64) -> `{ "command", "status", "name" }`

Persistence lives in `utils/first_wake_up.py` and `utils/robot_name.py`, each a
tiny JSON file under `~/.config/reachy_mini/`. Both are fail-safe: every read/
write error degrades gracefully (default `False` / `None`) and never raises, so
a storage problem can't break the command loop. `app/main.py` loads the
persisted name at startup so a rename survives reboots.

### BLE identify (`daemon/app/services/bluetooth/bluetooth_service.py`)

New `IDENTIFY` command. It is **public** (no PIN): identification happens before
the user authenticates, the action is benign, and BLE range already bounds who
can trigger it. Like the WiFi/update commands it proxies to the daemon
(`/api/media/play_sound`) off the GLib mainloop, returning `OK: working` and
delivering the real result as a later RESPONSE notification. Documented in
`BLE_WIFI_PROVISIONING.md`.

### JS SDK (`ts/lib/reachy-mini.ts`, `ts/lib/types.ts`)

`getFirstWakeUp` / `setFirstWakeUp` / `getRobotName` / `setRobotName`, using the
existing single-slot round-trip pattern. All resolve `null` on a closed channel
or a daemon that predates the command (callers fail-open).

## Tests

- `test_robot_name.py`, `test_first_wake_up.py`: round-trip, trim/length-cap,
  empty/invalid rejection, fail-safe on corrupt JSON and write errors.
- `test_ble_identify_proxy.py`: `_identify` relay + error mapping (503 ->
  "Audio not ready", URLError -> "Daemon unreachable"), Linux-only like the
  other BLE proxy tests.
- `test_js_protocol_sync.py` (existing) auto-verifies the new commands' protocol
  <-> TS parity.

## Notes

Two `style:` commits isolate the pre-existing ruff-format normalization that the
pre-commit hook flags when touching `abstract.py` / `protocol.py` /
`bluetooth_service.py`, so the feature commits stay reviewable.

## Test plan

- [ ] `set_*` then `get_*` round-trips over the data channel
- [ ] Rename survives a daemon restart (persisted name overrides `--robot-name`)
- [ ] `IDENTIFY` over BLE plays a sound and replies `OK: working` then a result
- [ ] Read/write errors (e.g. read-only home) degrade gracefully without raising
- [ ] Unknown/legacy clients are unaffected (new command types only)


src/reachy_mini/utils/first_wake_up.py
```diff
@@ -0,0 +1,48 @@
...
+
+from __future__ import annotations
+
+import json
+import logging
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+_STATE_PATH = Path.home() / ".config" / "reachy_mini" / "first_wake_up.json"
```
FabienDanieau, line 21: Shall we use the new config file: https://github.com/pollen-robotics/reachy_mini/pull/1254 ?

src/reachy_mini/utils/robot_name.py
```diff
@@ -0,0 +1,63 @@
...
+
+from __future__ import annotations
+
+import json
+import logging
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+_STATE_PATH = Path.home() / ".config" / "reachy_mini" / "robot_name.json"
```
FabienDanieau, line 22: same comment.

src/reachy_mini/daemon/app/services/bluetooth/bluetooth_service.py
```diff
@@ -1274,6 +1295,33 @@ def _daemon_request(
             return raw.decode("utf-8", errors="replace")
 
 
+# --- Robot identification proxy (reuses _daemon_request above) ---------------
+# Plays a short sound on the robot speaker so the user can pick their Reachy out
+# of the BLE scan list before connecting. Proxies to the daemon's
+# /api/media/play_sound route — note the /api prefix: the media router is
+# mounted under the prefixed APIRouter, unlike the /wifi and /update routes.
+_IDENTIFY_SOUND_FILE = "wake_up.wav"
```
FabienDanieau: I would use another file to have a sound unique for this procedure (or at least less used than the wake up)

PR #1220, opened 2026-06-18
feat(audio): let moves carry non-wav sound files

FabienDanieau:
GStreamer playbin already sniffs content, so non-wav audio plays fine; the only wav assumptions were in naming/lookup, not playback.

- recorded_move: dataset sidecar lookup tries all containers the daemon can play (wav/mp3/ogg/oga/opus/flac/m4a/aac), wav winning ties, instead of only .wav. This is what actually unblocks bundled non-wav moves.
- protocol: generalize UploadAudioStartCmd.encoding from wav-base64 to any "<container>-base64"; wav-base64 stays the default so older clients are unaffected.
- daemon: _handle_audio_finish derives the written extension from the declared encoding instead of hard-coding .wav.

The TS SDK still sends encoding='wav-base64', so live uploads stay wav until it's updated; this only widens what the daemon accepts.

Assisted-by: Claude:claude-opus-4-8

### DIFF
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -1633,7 +1633,9 @@ class Backend:
             import base64
             raw = base64.b64decode(payload, validate=False)
             os.makedirs(self._audio_temp_dir, exist_ok=True)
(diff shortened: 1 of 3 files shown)


RemiFabre (approved):
Thanks!

PR #1222, opened 2026-06-18
Resolve robot_name to its daemon via mDNS when connecting

### LINKED ISSUE #1221: robot_name is ignored when connecting to a robot
`ReachyMini` accepts a `robot_name` argument but never uses it when establishing the connection. As a result, the name has no effect: with multiple daemons on the same subnet there is no way to target a specific robot by name, and the client connects regardless of the value passed.

alozowski:
Fixes the issue where `robot_name` had no effect on connection. A non-default `robot_name` is now resolved to its daemon through mDNS discovery, so clients can target a specific robot by name when several daemons share a subnet. The default-name path is unchanged.

Closes https://github.com/pollen-robotics/reachy_mini/issues/1221


FabienDanieau:
I think @pierre-rouanet will be a better reviewer than me for this.

src/reachy_mini/reachy_mini.py
```diff
@@ -436,6 +442,25 @@ def _initialize_client(
...
         requested_mode = cast(ConnectionMode, requested_mode.lower())
+
+        # Only mDNS discovery can pick one named daemon out of several on a subnet.
+        if self.robot_name != DEFAULT_ROBOT_NAME and requested_mode != "localhost_only":
+            from reachy_mini.utils.discovery import find_robots
+
+            matches = [
+                robot
+                for robot in find_robots(timeout=timeout)
+                if robot.name == self.robot_name
```
pierre-rouanet, line 495: Should the robot name also used in the mdns service exposed by the robot?

alozowski: Added and verified it on a wireless with a custom name: the robot was advertised, discovered, and connected to successfully

pierre-rouanet:
Pushed a follow-up commit (`dada132d5`) — heads-up since it makes a small design call on your branch.

**Problem:** for a non-default `robot_name`, connection resolved purely through mDNS, so the `host` argument was silently ignored. `ReachyMini(robot_name="mybot", host="192.168.1.5")` couldn't connect when mDNS is blocked / cross-subnet / over a VPN, and `connection_mode="network"` failed outright with "not found on the network".

**Change:**
- `self.host` is now appended as a last-resort candidate after the mDNS-advertised addresses.
- To keep that safe, every dialed host is name-validated via `get_status` — factored into a new `_connect_and_verify_name` helper shared with the localhost path. So falling back to the default `host` can't silently connect you to a differently-named daemon; a mismatch raises.
- Bonus: multiple daemons sharing a name now emit a `warning` instead of being picked silently.

Added two tests: host fallback when mDNS misses, and the fallback rejecting a wrong-named daemon.

Note this now name-validates the mDNS path too (previously it trusted the TXT record), which also closes a small TOCTOU. Happy to split that out or drop the dup-name warning if you'd rather keep the diff minimal — feel free to adjust.

pierre-rouanet (approved):
LGTM, thanks @alozowski !

PR #1223, opened 2026-06-19
feat(daemon): device-code OAuth login with automatic token refresh

tfrere:
## Summary

Adds an **RFC 8628 device-code OAuth login** path to the daemon, alongside the existing authorization-code flow. This is the foundation for letting the robot's Hugging Face token **refresh itself without any user interaction**.

Two problems it solves:

1. **No more manual re-login.** Today the daemon stores only an `access_token` (JWT, ~30-day `exp`). When it expires, the central relay starts returning `401` and the user has to log in again by hand. The authorization-code flow HF uses here returns **no refresh token** ([HF docs](https://huggingface.co/docs/hub/oauth): *"No refresh tokens are provided"*). The device-code flow **does** return one: `huggingface_hub` persists it next to the token and `get_token()` (already used everywhere in the daemon) transparently renews the access token when less than a day of validity remains.
2. **No more `reachy-mini.local` dependency for login.** The device-code flow has **no redirect URI** - the phone just shows a short code + URL and the robot polls HF. So it works regardless of how the robot is addressed on the network (the long-standing mDNS-vs-IP pain during first-time setup).

It uses Hugging Face's first-party device-code OAuth client shipped in `huggingface_hub` (`DEVICE_CODE_OAUTH_CLIENT_ID`), so it works even when `HF_OAUTH_CLIENT_ID` is not configured.

## What's in this PR (daemon, additive, non-breaking)

- **`apps/sources/hf_auth.py`**: device-code session lifecycle (`start_device_code_login`, background polling via `asyncio.to_thread`, `get_device_code_session_status`, `cancel_device_code_session`), refresh-aware token persistence (delegates to `huggingface_hub._login._save_oauth_token` so the stored-token format stays in lock-step with `get_token()`), and central-relay notification on success.
- **`daemon/app/routers/hf_auth.py`**: 
  - `POST /api/hf-auth/oauth/device/start` -> `{ user_code, verification_uri, verification_uri_complete, session_id, interval, expires_in }`
  - `GET /api/hf-auth/oauth/device/status/{session_id}` -> `pending` / `authorized` (+ `username`) / `error` / `expired`; starts the central relay once on first success (mirrors the OAuth callback).
  - `DELETE /api/hf-auth/oauth/device/session/{session_id}`
- **`pyproject.toml`**: bump `huggingface-hub` to `>=1.20.0` (device-code login + transparent refresh landed in 1.20.0).
- **Tests**: `tests/unit_tests/test_hf_auth_device_code.py` covers the orchestration (session lifecycle, status polling, relay notification, error mapping) with the `huggingface_hub` device layer stubbed, so they run deterministically on any hub version.

The existing authorization-code flow (`/oauth/begin`, `/oauth/callback`, `/oauth/start`) is untouched, so nothing regresses.

## Notes / follow-ups

- **Mobile app (separate PR):** the setup wizard needs to call `/oauth/device/start`, display the `user_code` + `verification_uri`, and poll `/oauth/device/status/{id}` instead of opening the redirect-based `/oauth/begin`. Until then these endpoints are available but not yet consumed.
- **Scopes:** the device-code flow yields HF's standard device-login token (same as `hf auth login`), not the Pollen app's explicit scope list. Worth confirming it grants what the robot needs (inference + repo access).
- **Private hub helpers:** uses `huggingface_hub.utils._oauth_device` and `huggingface_hub._login._save_oauth_token`. These are private but stable within the pinned `>=1.20,<2` range; the persisted stored-tokens contract that `get_token()` relies on is the stable surface.

## Test plan

- [x] `pytest tests/unit_tests/test_hf_auth_device_code.py` (10 passed)
- [x] `ruff check` + `ruff format --check` on changed src files
- [x] Verified `request_device_code` / `poll_device_token` / `_save_oauth_token` / `DeviceCodeError` import and signatures against `huggingface_hub==1.20.1`
- [ ] End-to-end on a robot once the mobile app calls the new endpoints (token persists with refresh metadata, survives `expires_at` crossing without re-login)


pierre-rouanet:
### Current state for reviewers

- **Feature:** adds a device-code (RFC 8628) HF OAuth flow — redirect-free (no `reachy-mini.local` dependency) and refresh-capable, so `get_token()` transparently renews without re-auth. Backend + endpoints only; no UI wiring yet.
- **Dependency:** `huggingface-hub` range-pinned `>=1.20.1,<2.0.0`. It uses **private** hub internals (`utils._oauth_device.{request,poll}`, `_login._save_oauth_token`); floor is 1.20.1 because the `_oauth_device` module moved there from 1.20.0. Verified import- and signature-stable through 1.24.0.
- **Drift guard:** `tests/unit_tests/test_hf_hub_private_api_contract.py` asserts those private symbols exist and are callable as we call them; a dedicated CI job (`hf-api-contract.yml`) runs it against the **latest** allowed hub (PRs + weekly), so an upstream break surfaces in CI, not on a robot.
- **Robustness:** authorized sessions now expire via a TTL (no in-memory leak); `cancel` sets a `threading.Event` observed by `poll_device_token`'s `on_pending` hook, so it actually stops the worker thread (~1 poll interval) instead of leaking it for ~15 min.
- **Green:** branch is up to date with `main`; all CI passes (pytest macos+ubuntu, mypy, ruff, check-api-docs incl. the 3 new endpoints, uv-lock-check, contract-latest, docs build).

PR #1224, opened 2026-06-21
Use capped exponential backoff for central relay reconnect

thc1006:
Follow-up to #1201, addressing #1165.

#1201 routes the 401/ERROR reconnect through the existing fixed `RECONNECT_INTERVAL` (5 s). That stops the storm, but a persistently unreachable central server is still retried every 5 s. This swaps the fixed wait for a capped exponential backoff (5 s → … → `MAX_RECONNECT_INTERVAL`, 60 s) with a small additive jitter to de-synchronize reconnects across a fleet.

`_connection_attempts` already resets to 0 on a successful connection (and via `update_token` / `force_reconnect`), so the backoff restarts from 5 s after recovery, and the `_token_updated` event still short-circuits the wait — a login / token refresh reconnects immediately.

On the auth-failure circuit breaker also floated in #1165: I left it out on purpose. The loop already re-fetches the token on every attempt and resets the counter on success, so a token that becomes valid self-heals on the next retry even without `update_token` firing. A hard "stop retrying until token update" breaker would regress that and could get stuck on a transient 401; the capped backoff stops the hammering while keeping recovery automatic. Happy to add the breaker if you'd prefer it.

Added a unit test covering the backoff growth and cap; the existing relay tests still pass. `MAX_RECONNECT_INTERVAL` and the jitter fraction are easy to tune.

### DIFF
--- a/src/reachy_mini/media/central_signaling_relay.py
+++ b/src/reachy_mini/media/central_signaling_relay.py
@@ -13,6 +13,7 @@ import asyncio
 import json
 import logging
 import os
+import random
 import threading
 from enum import Enum
 from typing import Any, Callable, Optional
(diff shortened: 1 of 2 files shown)


FabienDanieau (approved):
LGTM

PR #1225, opened 2026-06-22
Fix : Prevent yaw from being interpolated through 180° when using goto.

scienceetonnante:
## Problem

When going from say yaw=120° to yaw=-120°, the linear shortest path interpolation tries to go through the "back route" turning at 180°. This is not possible with Reachy Mini and leads to jumpy behavior

Example code:
```
import time

from reachy_mini.utils import create_head_pose

with ReachyMini() as mini:
    for yaw in (0, 120, -120, 0):
        mini.goto_target(head=create_head_pose(yaw=yaw, degrees=True), duration=2.0)
        time.sleep(0.3)
```

## Solution

Instead of interpolating yaw as an angle, we treat it as a scalar.

This is done by adding a yaw_as_scalar flag to `linear_pose_interpolation`, and interpolating yaw as a signed scalar (instead of an angle) and let linear interpolation happen on pitch/yaw/translation.

NOTE: I added it as an option to keep the previous behavior available. But we might consider this should be the only way to interpolate, in which case I can modify the function and just remove the flag.

### DIFF
--- a/src/reachy_mini/motion/goto.py
+++ b/src/reachy_mini/motion/goto.py
@@ -57,7 +57,10 @@ class GotoMove(Move):
         interp_time = time_trajectory(t / self.duration, method=self.method)
 
         interp_head_pose = linear_pose_interpolation(
-            self.start_head_pose, self.target_head_pose, interp_time
+            self.start_head_pose,
+            self.target_head_pose,
+            interp_time,
+            yaw_as_scalar=True,
         )
         interp_antennas_joint = (
             self.start_antennas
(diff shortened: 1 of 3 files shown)


FabienDanieau:
I can reproduce the issue indeed. Looks like it should be the default control mode. Any opinion from the control people @pierre-rouanet @askuric ?

pierre-rouanet:
LGTM!

scienceetonnante:
Let me know if I should rewrite this as the only mode (removing the flag to
opt in/out)

Le mer. 24 juin 2026 à 13:55, Pierre Rouanet ***@***.***> a
écrit :

pierre-rouanet:
I feel like having the flag is a good reminder of this issue! I'd go with the way you've done it right now personally.

FabienDanieau (approved):
Let's merge then!

PR #1228, opened 2026-06-22
Add daemon off requirement in motor_diagnosis.md

### LINKED ISSUE #1227: Doc, daemon off before running testbench app
Troubleshooting page should mention that the daemon needs to be turned off to have the second interface of the testbench app and be able to debug motors ralated issues.

augustin-crampette:


FabienDanieau (approved):
Thanks!

PR #1233, opened 2026-06-24
feat(sdk): forward audio container type on upload instead of forcing wav

### LINKED ISSUE #1232: TS SDK should accept the same audio formats the daemon does (and not just wav)

RemiFabre:
The daemon (UploadAudioStartCmd.encoding) already accepts any <container>-base64 and GStreamer playbin decodes whatever it's given, so the TS SDK shouldn't narrow valid audio to WAV. Derive the upload encoding from the audioBlob's MIME type, defaulting to wav-base64 for untyped/legacy blobs. Unblocks bundling Opus/Ogg move audio from the Marionette apps.

Closes #1232

### DIFF
--- a/ts/lib/reachy-mini.ts
+++ b/ts/lib/reachy-mini.ts
@@ -24,6 +24,7 @@ import {
     bytesToBase64,
     gzipBase64,
     clampVolume,
+    audioUploadEncoding,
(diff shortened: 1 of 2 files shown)


FabienDanieau (approved):
LGTM

PR #1237, opened 2026-06-25
Stop /state/ws/full streaming loop when backend is no longer ready

### LINKED ISSUE #1236: Unhandled AssertionError on /state/ws/full after daemon stop
On Wireless, stopping the daemon while a client is streaming the robot state over the`/state/ws/full` WebSocket produces an unhandled exception in the logs.

The daemon-stop job completes normally:
```
reachy_mini.daemon.daemon - INFO - Central signaling relay stopped
reachy_mini.daemon.daemon - INFO - Daemon stopped successfully.
logs_job_... - INFO - Job 'daemon-stop' completed successfully
```
Immediately after, the still-open state stream throws:
```
uvicorn.error - ERROR - Exception in ASGI application
Traceback (most recent call last):
...
File ".../reachy_mini/daemon/app/routers/state.py", line 153, in ws_full_state
full_state = await get_full_state(
File ".../reachy_mini/daemon/app/routers/state.py", line 104, in get_full_state
result["head_joints"] = backend.get_present_head_joint_positions()
File ".../reachy_mini/daemon/backend/robot/backend.py", line 462, in get_all_joint_positions
assert self.c is not None, "Motor controller not initialized or already closed."
AssertionError: Motor controller not initialized or already closed.
```
The daemon itself stays up and keeps serving requests (`GET /api/daemon/status` returns 200 afterwards), so this is not a crash, but the shutdown leaves a noisy unhandled traceback in the logs every time a state stream is open.

alozowski:
Gate the `/state/ws/full` streaming loop on `backend.ready` so it exits cleanly when the daemon stops, instead of reading from the closed backend and throwing the AssertionError from the linked issue.
Closes https://github.com/pollen-robotics/reachy_mini/issues/1236

### DIFF
--- a/src/reachy_mini/daemon/app/routers/state.py
+++ b/src/reachy_mini/daemon/app/routers/state.py
@@ -149,7 +149,7 @@ async def ws_full_state(
     period = 1.0 / frequency
 
     try:
-        while True:
+        while backend.ready.is_set():
             full_state = await get_full_state(
                 with_head_pose=with_head_pose,
                 with_target_head_pose=with_target_head_pose,


pierre-rouanet (approved):
I don't know if it's hard or not but if you can come up with a unites associated that would be awesome!
But LGTM anyway!

alozowski:
@pierre-rouanet thank you for checking this! 

> I don't know if it's hard or not but if you can come up with a unites associated that would be awesome!
let me think!

PR #1242, opened 2026-06-29
Audio firmware 2.1.4

FabienDanieau:
Increases the max volume by 6dB
Fixes issue when microphone is not outputting audio anymore, probably after a usb reset


FabienDanieau:
Added the 6-channel firmware binary and documented it (changelog, media controls guide, and troubleshooting FAQ) for accessing the raw microphone output.

alozowski (approved):

PR #1244, opened 2026-07-01
Start default app from antenna touch

andimarafioti:
## Summary

Adds an idle-only antenna watcher for the configured startup app. When the daemon has a startup app installed and no managed local app or remote WebRTC session owns the robot, touching either antenna starts the startup app again.

If Reachy is asleep, the watcher detects antenna movement against the current antenna idle pose, enables motor control, runs the normal wake-up sequence, and then starts the startup app if the app slot is still free.

If Reachy is already awake, the watcher only plays `wake_up.wav`, then starts the startup app. It does not move the head, so it does not fight the user holding the antenna down.

After an app stops, the detector re-baselines on the real current antenna pose instead of the commanded target. That lets it re-arm even when the antenna is held slightly away from target by physical counter-pressure.

The watcher is designed to stay cheap on the Raspberry Pi 4: it sleeps while an app/session is active, samples cached antenna state at 10 Hz only while idle, and uses hysteresis against the idle pose to avoid repeat triggers while held.

## Details

- Adds a non-evicting local acquire path to `RobotAppLock`.
- Lets `AppManager.start_app(..., evict_remote=False)` fail instead of displacing a remote session.
- Moves startup-app install/launch/watch helpers into `daemon/app/startup_app.py`.
- Starts/cancels the startup-app antenna watcher with the daemon lifespan.
- Adds tests for idle-pose baseline mode, sleep wake-up, awake startup sound without movement, offset re-arming, idle-only startup behavior, and the new lock path.

## Validation

- `.venv/bin/python -m pytest tests/unit_tests/test_autostart_app.py tests/unit_tests/test_robot_app_lock.py tests/unit_tests/test_app.py::test_app_manager tests/unit_tests/test_app.py::test_faulty_app -q` — 39 passed
- `.venv/bin/python -m ruff check src/reachy_mini/daemon/app/main.py src/reachy_mini/daemon/app/startup_app.py src/reachy_mini/apps/manager.py src/reachy_mini/daemon/robot_app_lock.py`
- `.venv/bin/python -m mypy src/reachy_mini/daemon/app/main.py src/reachy_mini/daemon/app/startup_app.py src/reachy_mini/apps/manager.py src/reachy_mini/daemon/robot_app_lock.py`

Note: `uv run --locked ...` did not reach pytest because this local `uv` rejected the repo existing `exclude-newer = "7 days"` setting and wanted to update `uv.lock`. Running `tests/unit_tests/test_app.py::test_app` in the existing `.venv` also fails before this change path because the MuJoCo extra is not installed.

### DIFF
--- a/src/reachy_mini/daemon/app/startup_app.py
+++ b/src/reachy_mini/daemon/app/startup_app.py
@@ -6,9 +6,6 @@ from collections.abc import Callable
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
-import numpy as np
-from scipy.spatial.transform import Rotation as R
-
 from reachy_mini.apps import SourceKind
 from reachy_mini.apps.manager import AppManager
 from reachy_mini.daemon.robot_app_lock import RobotAppLockState
@@ -167,19 +164,8 @@ async def start_startup_app_if_idle(
 
 
 async def play_awake_startup_cue(backend: Any) -> None:
-    """Play the wake sound and head tilt without running the full wake-up pose."""
-    try:
-        start_pose = np.array(backend.get_current_head_pose(), dtype=float).copy()
-    except Exception:
-        start_pose = np.array(backend.INIT_HEAD_POSE, dtype=float).copy()
-
-    tilt_pose = start_pose.copy()
-    roll = R.from_euler("xyz", [20, 0, 0], degrees=True).as_matrix()
-    tilt_pose[:3, :3] = roll @ start_pose[:3, :3]
-
+    """Play the wake sound without running the full wake-up pose."""
     backend.play_sound("wake_up.wav")
-    await backend.goto_target(tilt_pose, duration=0.2)
-    await backend.goto_target(start_pose, duration=0.2)
 
 
 async def wake_or_start_startup_app_if_idle(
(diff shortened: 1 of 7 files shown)


src/reachy_mini/daemon/app/startup_app.py
```diff
@@ -0,0 +1,244 @@
...
+
+    backend = daemon.backend
+    if backend is None:
+        return False
+
+    try:
+        if backend.get_motor_control_mode() == MotorControlMode.Disabled:
+            logger.info(f"Waking up from antenna touch before starting app: {name}")
+            backend.set_motor_control_mode(MotorControlMode.Enabled)
+            await backend.wake_up()
```
FabienDanieau, line 186: It seems that there is a race condition issue. This appears
```
reachy_mini.daemon.robot_app_lock - INFO - RobotAppLock: local acquire refused (state=local_app holder='reachy_mini_conversation_app' requester='reachy_mini_conversation_app')
reachy_mini.daemon.app.startup_app - ERROR - Failed to auto-start app 'reachy_mini_conversation_app': The robot app slot is already in use
reachy_mini.apps.manager.runner - INFO - App reachy_mini_conversation_app is running
```
The app is launched twice I suppose

src/reachy_mini/daemon/app/startup_app.py
```diff
@@ -0,0 +1,244 @@
...
+                continue
+
+            present = _read_current_antenna_pair(backend)
+            if present is None:
+                detector.reset()
+                await asyncio.sleep(idle_poll_interval_s)
+                continue
+
+            if detector.update(present):
+                await wake_or_start_startup_app_if_idle(app_manager, daemon, name)
```
FabienDanieau: This may won't return an exception if there is an issue. We may want to add a  `detector.reset() ` in case of something went wrong.

FabienDanieau (commented):
Should we add a new flag or we consider that --startup-app explicitly uses this feature?

FabienDanieau (approved):

PR #1246, opened 2026-07-01
Proper sleep: keep the robot asleep (and limp) after an app stops

FabienDanieau:
## Summary
Make the robot settle into a real sleep — the same compliant state as a fresh `--no-wake-up-on-start` daemon start — instead of standing back up after an app ends.

- `goto_sleep()` now disables motor torque once it reaches the sleep pose, so the robot rests limp (soft antennas) on every sleep path (app exit, the go-to-sleep tool, the sleep button, REST, daemon stop).
- The app manager no longer returns the robot to the zero pose after an app stops **if it's already in the sleep pose** — so a sleep the app requested isn't immediately overridden.

## Why
Previously, an app ending would send the robot back to the neutral/zero pose with motors held, so it never actually went to sleep and the antennas stayed stiff.

## Notes
- Complements conversation-app PR [#438](https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/438) (which routes the app's exit paths through `goto_sleep`): app asks to sleep → daemon honors it → robot rests like at boot.
- Motor position feedback still works with torque off, so pose reads (e.g. the antenna-touch watcher) are unaffected.
- The stay-asleep check uses a 10 magic-mm tolerance around the sleep pose; if the head droops when it goes limp, that's the knob to loosen.

### DIFF
--- a/src/reachy_mini/apps/manager.py
+++ b/src/reachy_mini/apps/manager.py
@@ -13,6 +13,7 @@ import psutil
 from pydantic import BaseModel
 
 from reachy_mini.daemon.backend.robot import RobotBackend
+from reachy_mini.io.protocol import MotorControlMode
 from reachy_mini.utils.interpolation import distance_between_poses
 
 from . import AppInfo, SourceKind
@@ -308,8 +309,10 @@ class AppManager:
                 backend.get_current_head_pose(), backend.SLEEP_HEAD_POSE
             )
             if dist_to_sleep <= SLEEP_POSE_MAGIC_DISTANCE:
+                # pose check only; ensure limp (idempotent)
+                backend.set_motor_control_mode(MotorControlMode.Disabled)
                 self.logger.getChild("runner").info(
-                    "Robot is asleep; leaving it in the sleep pose."
+                    "Robot is asleep; leaving it limp in the sleep pose."
                 )
             else:
                 if isinstance(backend, RobotBackend):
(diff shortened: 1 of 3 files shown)


src/reachy_mini/apps/manager.py
```diff
@@ -297,27 +301,38 @@ async def stop_current_app(self, timeout: float | None = 20.0) -> None:
...
-                    INIT_ANTENNAS_JOINT_POSITIONS,
-                    INIT_HEAD_POSE,
+            backend = self.daemon.backend
+            _, _, dist_to_sleep = distance_between_poses(
+                backend.get_current_head_pose(), backend.SLEEP_HEAD_POSE
+            )
+            if dist_to_sleep <= SLEEP_POSE_MAGIC_DISTANCE:
+                self.logger.getChild("runner").info(
+                    "Robot is asleep; leaving it in the sleep pose."
                 )
```
andimarafioti, line 316: Should we call `backend.disable_motors()` here? or check that they are disabled?

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -1032,6 +1032,9 @@ async def goto_sleep(self) -> None:
         self._last_head_pose = self.SLEEP_HEAD_POSE
         await asyncio.sleep(sleep_time)
 
+        # Rest limp at the sleep pose, like a fresh boot.
+        self.set_motor_control_mode(MotorControlMode.Disabled)
```
andimarafioti, line 1036: Or are you doing it here anyway?

andimarafioti (approved):
Just a question, but looks good 👍

PR #1250, opened 2026-07-02
Add get_frame_jpeg() to encode camera frames as jpeg via gstreamer

### LINKED ISSUE #1249: No way to get an encoded camera frame from the media API
`MediaManager.get_frame()` returns camera frames only as raw BGR numpy arrays. Any consumer that needs a compressed image, such as a jpeg still to send to a vision model, has to add and maintain its own encoding dependency to convert the frame.

alozowski:
Adds `MediaManager.get_frame_jpeg()` / `CameraBase.read_jpeg()` encoding the current BGR frame to JPEG with `jpegenc`, so downstream apps don't need their own encoder. Works across LOCAL, WEBRTC and simulation (all deliver BGR at the client). Reuses `get_sample`.

Closes https://github.com/pollen-robotics/reachy_mini/issues/1249

### DIFF
--- a/src/reachy_mini/media/camera_base.py
+++ b/src/reachy_mini/media/camera_base.py
@@ -24,6 +24,7 @@ from reachy_mini.media.camera_constants import (
     MujocoCameraSpecs,
 )
 from reachy_mini.media.camera_utils import scale_intrinsics
+from reachy_mini.media.gstreamer_utils import encode_bgr_to_jpeg
 
 
 class CameraBase(ABC):
@@ -157,6 +158,13 @@ class CameraBase(ABC):
 
         """
         ...
+
+    def read_jpeg(self) -> Optional[bytes]:
+        """Pull the latest frame encoded as JPEG bytes, or ``None`` if unavailable."""
+        frame = self.read()
+        if frame is None:
+            return None
+        return encode_bgr_to_jpeg(frame, self.logger)
 
     @abstractmethod
     def close(self) -> None:
--- a/src/reachy_mini/media/camera_base.py
+++ b/src/reachy_mini/media/camera_base.py
@@ -35,8 +35,9 @@ except ImportError as e:
     ) from e
 
 gi.require_version("Gst", "1.0")
+gi.require_version("GstApp", "1.0")
 
-from gi.repository import Gst  # noqa: E402
+from gi.repository import Gst, GstApp  # noqa: E402
 
 
 class CameraBase(ABC):
@@ -57,9 +58,9 @@ class CameraBase(ABC):
         self.camera_specs: Optional[CameraSpecs] = None
         self.resized_K: Optional[npt.NDArray[np.float64]] = None
         self._jpeg_pipeline: Gst.Pipeline = None
-        self._jpeg_appsrc: Gst.Element = None
-        self._jpeg_appsink: Gst.Element = None
-        self._jpeg_size: Optional[tuple[int, int]] = None
+        self._jpeg_appsrc: GstApp.AppSrc = None
+        self._jpeg_appsink: GstApp.AppSink = None
+        self._jpeg_resolution: Optional[tuple[int, int]] = None
 
     @property
     def resolution(self) -> tuple[int, int]:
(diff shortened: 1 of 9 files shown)


src/reachy_mini/media/gstreamer_utils.py
```diff
@@ -76,6 +79,38 @@ def get_sample(appsink: GstApp.AppSink, logger: logging.Logger) -> Optional[byte
     return data
 
 
+def encode_bgr_to_jpeg(
+    frame: npt.NDArray[np.uint8], logger: logging.Logger
+) -> Optional[bytes]:
+    """Encode a BGR frame as JPEG bytes, or ``None`` if encoding fails."""
+    Gst.init([])  # idempotent
```
FabienDanieau: Actually this could be moved to camerabase directly (and audiobase) since all child are using it

src/reachy_mini/media/gstreamer_utils.py
```diff
@@ -76,6 +79,38 @@ def get_sample(appsink: GstApp.AppSink, logger: logging.Logger) -> Optional[byte
...
+    height, width = frame.shape[:2]
+    try:
+        pipeline = Gst.parse_launch(
+            "appsrc name=src format=time is-live=false "
+            "! videoconvert ! jpegenc ! appsink name=sink sync=false"
```
FabienDanieau: Please create separate elements, they are easier to configure and to debug

src/reachy_mini/media/camera_base.py
```diff
@@ -158,6 +159,13 @@ def read(self) -> Optional[npt.NDArray[np.uint8]]:
         """
         ...
 
+    def read_jpeg(self) -> Optional[bytes]:
+        """Pull the latest frame encoded as JPEG bytes, or ``None`` if unavailable."""
```
FabienDanieau: Specific that this is for *occasionally* converting images. This is not optimized for video

src/reachy_mini/media/gstreamer_utils.py
```diff
@@ -76,6 +79,38 @@ def get_sample(appsink: GstApp.AppSink, logger: logging.Logger) -> Optional[byte
...
 
 
+def encode_bgr_to_jpeg(
+    frame: npt.NDArray[np.uint8], logger: logging.Logger
+) -> Optional[bytes]:
+    """Encode a BGR frame as JPEG bytes, or ``None`` if encoding fails."""
+    Gst.init([])  # idempotent
+    height, width = frame.shape[:2]
+    try:
+        pipeline = Gst.parse_launch(
```
FabienDanieau: We'll create a pipeline each time we convert an image. Why not initializing one first and keeping a ref?
The state can be just paused and resumed when needed, and we tear it down at the end.

FabienDanieau (changes requested):
We may note somewhere that the lite already [outputs a jpeg](https://github.com/pollen-robotics/reachy_mini/blob/83df7f9743e774fa4d14593497971d320a528a18/src/reachy_mini/media/media_server.py#L825), but to align with the wireless and the osx lite platform we decode it.
An alternative approach would be to output the jpeg in the main [read](https://github.com/pollen-robotics/reachy_mini/blob/83df7f9743e774fa4d14593497971d320a528a18/src/reachy_mini/media/camera_gstreamer.py#L257). The user would have to stop the media, and restart it with a jpeg parameter in a way similar we do for changing resolution. Wdyt?

alozowski:
On the jpeg-mode `read()`: since `read()` is client-side and the daemon already decodes to raw before the IPC tee, a jpeg mode there would just re-encode the raw (no passthrough), and the stop/restart makes it costly per still, so I kept the on-demand `read_jpeg` for occasional stills

src/reachy_mini/media/camera_base.py
```diff
@@ -37,11 +50,16 @@ class CameraBase(ABC):
...
+        Gst.init([])
         self.logger = logging.getLogger(type(self).__module__)
         self.logger.setLevel(log_level)
         self._resolution: Optional[CameraResolution] = None
         self.camera_specs: Optional[CameraSpecs] = None
         self.resized_K: Optional[npt.NDArray[np.float64]] = None
+        self._jpeg_pipeline: Gst.Pipeline = None
+        self._jpeg_appsrc: Gst.Element = None
+        self._jpeg_appsink: Gst.Element = None
+        self._jpeg_size: Optional[tuple[int, int]] = None
```
FabienDanieau: I'd prefer resolution, since size could be misleading (size of the jpeg file)

src/reachy_mini/media/camera_base.py
```diff
@@ -158,6 +176,60 @@ def read(self) -> Optional[npt.NDArray[np.uint8]]:
...
+        """
+        frame = self.read()
+        if frame is None:
+            return None
+        height, width = frame.shape[:2]
+        if self._jpeg_pipeline is None or self._jpeg_size != (width, height):
+            self._release_jpeg_encoder()
+            self._build_jpeg_encoder(width, height)
+        self._jpeg_pipeline.set_state(Gst.State.PLAYING)
+        self._jpeg_appsrc.emit("push-buffer", Gst.Buffer.new_wrapped(frame.tobytes()))
```
FabienDanieau: you can use self._jpeg_appsrc.push_buffer() directly. You need to import GstApp

FabienDanieau (approved):
Works! Added some optional comments

PR #1240, opened 2026-06-26
Feat/auto start app

FabienDanieau:
Auto-start an app after wake-up

Adds a `--startup-app` flag so the daemon launches a chosen app once the robot wakes up. It's **opt-in** (off by default) — a first step toward having the conversation app run out of the box.

### How it works
- **Install at boot:** if the app isn't already installed, it's fetched from the Hugging Face catalog and installed when the daemon starts (no robot needed to download into `apps_venv`), so it's ready before the robot wakes.
- **Start on first wake-up:** the app is launched by a one-shot hook fired when `wake_up()` completes — so it works however the wake is triggered (on-start, the desktop app, the button, or the REST endpoint). It fires only on the **first** wake; later wakes are ignored.
- Every step is best-effort: a bad app name or a failed install is logged and never blocks daemon startup.
- Includes a small fix: HfApi space metadata is normalized to JSON-plain so installing from the catalog in-process doesn't choke on non-serializable objects.

### Try it
The wireless unit boots asleep (`--no-wake-up-on-start`), so start the daemon with:

```bash
reachy-mini-daemon --wireless-version --no-wake-up-on-start --startup-app reachy_mini_conversation_app
```

Then wake it manually (motors must be enabled **first**, then play the wake animation):

```
http://reachy-mini.local:8000/docs#/default/set_motor_mode_api_motors_set_mode__mode__post
http://reachy-mini.local:8000/docs#/default/play_wake_up_api_move_play_wake_up_post
```

Or just open the **desktop app** — it enables motors, wakes the robot, and you'll see the conversation app running on the robot.

### Open questions / follow-ups
- **Persist the selected startup app** in a config file, tunable from the desktop/mobile apps (rather than a CLI flag). Likely a follow-up PR.
- **Wake-on-word / touch:** let the robot wake on its own so you don't need to open an app at all — the natural next step.

PR #1254, opened 2026-07-03
Feat/1247 configure startup app

### LINKED ISSUE #1247: Configure the startup app
Allow configuration of the startup app, save the config, expose to the desktop / mobile app

FabienDanieau:
## Rationale

Closes #1247. PR #1240 added a `--startup-app` **CLI flag**, but the choice lived only in the arg (nothing set it, so no app ever auto-started) and the desktop/mobile apps couldn't read or change it.

This makes the startup app a **persisted, API-configurable** setting:

- **Persisted per-user** as JSON, so it survives reboots/app updates.
- **REST API** `GET`/`PUT /api/apps/startup-app`.
- **Applied live** — a change re-arms the running daemon (wake hook + antenna watcher), no restart needed.
- The `--startup-app` CLI flag is **removed** (config is the single source of truth).
- Antenna-touch launch (from #1244) honors the configured app, and no longer false-triggers on the go-to-sleep motion (the sleep swing was read as a touch).
- OpenAPI spec regenerated (`docs/source/API/openapi.json`).

## Config file location

`platformdirs.user_config_dir("reachy_mini") / "daemon_config.json"` → on Linux `~/.config/reachy_mini/daemon_config.json` (on the wireless robot: `/home/pollen/.config/reachy_mini/daemon_config.json`). Shape:

```json
{ "startup_app": "reachy_mini_conversation_app" }
```

## How to test

Against a running daemon (`localhost:8000`):

```bash
# 1. current value (null when unset)
curl -s localhost:8000/api/apps/startup-app          # -> {"startup_app":null}

# 2. an app that isn't installed is rejected
curl -s -o /dev/null -w '%{http_code}\n' -X PUT localhost:8000/api/apps/startup-app \
  -H 'content-type: application/json' -d '{"startup_app":"not-installed"}'   # -> 400

# 3. pick one that IS installed
curl -s localhost:8000/api/apps/list-available/installed | python3 -m json.tool
curl -s -X PUT localhost:8000/api/apps/startup-app \
  -H 'content-type: application/json' -d '{"startup_app":"reachy_mini_conversation_app"}'

# 4. it persisted to disk (no restart)
cat ~/.config/reachy_mini/daemon_config.json

# 5. clear it
curl -s -X PUT localhost:8000/api/apps/startup-app \
  -H 'content-type: application/json' -d '{"startup_app":null}'
```

Install-then-set for an app that isn't installed yet: `POST /api/apps/install` → poll `GET /api/apps/job-status/{job_id}` → then the `PUT` above.

**Live-apply / hardware checks:**
- After a `PUT`, no daemon restart: on the next wake-up (or an antenna touch on an idle robot) the newly chosen app launches.
- Send the robot to sleep (turn-off) with a startup app set → it completes the sleep motion and **stays asleep** (the sleep swing no longer relaunches the app); a real antenna touch still wakes it and starts the app.

### DIFF
--- a/docs/source/API/openapi.json
+++ b/docs/source/API/openapi.json
@@ -302,6 +302,62 @@
         }
       }
     },
+    "/api/apps/startup-app": {
+      "get": {
+        "summary": "Get Startup App",
+        "description": "Get the app configured to auto-start on wake-up (null if unset).",
+        "operationId": "get_startup_app_api_apps_startup_app_get",
+        "responses": {
+          "200": {
+            "description": "Successful Response",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/StartupApp"
+                }
+              }
+            }
+          }
+        }
+      },
+      "put": {
+        "summary": "Set Startup App",
+        "description": "Set (or clear, with null) the app that auto-starts on wake-up.\n\nMust already be installed (install via /apps/install first). Applied live, so\nno daemon restart is needed.",
+        "operationId": "set_startup_app_api_apps_startup_app_put",
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/StartupApp"
+              }
+            }
+          },
+          "required": true
+        },
+        "responses": {
+          "200": {
+            "description": "Successful Response",
+            "content": {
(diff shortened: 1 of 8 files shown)


alozowski (approved):
Everything works 👍🏻

PR #1256, opened 2026-07-07
1245 improve the first interaction after building the robot

### LINKED ISSUE #1245: Improve the first interaction after building the robot
- Maybe the robot could play a move
- Maybe we can connect to the wifi by just showing a QR code

RemiFabre:
(just to use the online diff tool)


RemiFabre:
Hello,
While the traction on socials was ~decent, I'm now convinced the secret handshake should **not** be a default feature on Mini for now.
Some people find it weird to hold the head of the robot and pull on the antennas (you don't do that to your pet companion).
Also the added value doesn't seem to warrant the risk of confusion. Path B is more solid and should be the main path.

In any case, this is a working prototype with:
- A way to interact with the robot as soon as it's built
- A way to connect to your wifi without the need of an app (cf [this demo](https://discord.com/channels/519098054377340948/1519311872787087380/1524168258935193600))
- A robust way to detect "antenna presses" and "antenna collisions" + a simple state machine to detect codes

Everything can be found here:
1245-improve-the-first-interaction-after-building-the-robot

I'd argue that sharing the WiFi throught a QR code is a nice thing to integrate though. Maybe add the option through the smartphone app and reuse my code for the QR detection on the robot?
I think the idea of interacting with Mini with 0 screens (no smartphone, no laptop) is an exciting one, but it's an entire new thing. I'd be down to work on this if at some point we think it's worth it

PR #1259, opened 2026-07-09
perf(motion): replace speech tapper sample deque with a numpy window

ai-ag2026:
## Summary

`SwayRollRT.feed()` keeps a 10 s `deque[float]` of samples and rebuilds the analysis frame with `np.fromiter(islice(...))` on **every hop**, although only the most recent `frame` samples (`FRAME_MS` = 20 ms) are ever read. The per-hop `tolist()` / `fromiter` round-trip runs inside the GStreamer audio callback and is surprisingly expensive on small boards.

This PR keeps just the rolling window as a `float32` array instead of the deque. No API change (`samples` keeps its name; its type changes from `deque[float]` to `NDArray[np.float32]` — internal analysis state).

## Why

Profiling a Reachy Mini Wireless (CM4) while assistant audio plays, py-spy attributes ~20 % of the app process's Python CPU to `speech_tapper.feed` (native leaf: `PyArray_FromIter`). On a 4-core CM4 that headroom matters.

## Validation

- **Bit-identical outputs:** old vs new implementation fed with 30 s of synthetic speech/silence in 1024-sample chunks → 599 hop dicts, max per-key delta < 1e-12.
- `pytest tests/unit_tests/test_head_wobbler.py` → 24 passed.
- Microbenchmark (same 30 s stream): ~508 ms → ~10 ms for the feed loop on x86 (~50×); the relative win on the CM4 is larger since `fromiter` dominates there.

## Notes

- `deque`/`islice` imports drop out; no new dependencies.
- The 10 s deque history was never read beyond the last 20 ms frame, so no behavioral change.

### DIFF
--- a/src/reachy_mini/motion/speech_tapper.py
+++ b/src/reachy_mini/motion/speech_tapper.py
@@ -9,8 +9,6 @@ Ported from *reachy_mini_conversation_app*.
 from __future__ import annotations
 
 import math
-from collections import deque
-from itertools import islice
 
 import numpy as np
 from numpy.typing import NDArray
@@ -91,7 +89,11 @@ class SwayRollRT:
         self.sample_rate = int(sample_rate)
         self.frame = int(self.sample_rate * FRAME_MS / 1000)
         self.hop = int(self.sample_rate * HOP_MS / 1000)
-        self.samples: deque[float] = deque(maxlen=10 * self.sample_rate)
+        # Rolling analysis window: only the most recent ``frame`` samples are ever
+        # read, so keep just those as a float32 array instead of a 10 s deque of
+        # Python floats (the per-hop tolist()/fromiter round-trip dominated CPU on
+        # Raspberry Pi CM4 class hardware).
+        self.samples: NDArray[np.float32] = np.zeros(0, dtype=np.float32)
         self.carry: NDArray[np.float32] = np.zeros(0, dtype=np.float32)
 
         self.vad_on = False
@@ -113,7 +115,7 @@ class SwayRollRT:
 
     def reset(self) -> None:
         """Reset state (VAD/env/buffers/time) but keep initial phases/seed."""
-        self.samples.clear()
+        self.samples = np.zeros(0, dtype=np.float32)
         self.carry = np.zeros(0, dtype=np.float32)
         self.vad_on = False
         self.vad_above = 0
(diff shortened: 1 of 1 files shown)


src/reachy_mini/motion/speech_tapper.py
```diff
@@ -147,16 +149,15 @@ def feed(self, pcm: NDArray[np.float32]) -> list[dict[str, float]]:
             hop = self.carry[:self.hop]
             self.carry = self.carry[self.hop:]
 
-            self.samples.extend(hop.tolist())
-            if len(self.samples) < self.frame:
+            if self.samples.size:
```
FabienDanieau, line 154: With the [current constants](https://github.com/ai-ag2026/reachy_mini/blob/fix/speech-tapper-ring-buffer/src/reachy_mini/motion/speech_tapper.py#L19C1-L20C12) `FRAME_MS = 20 HOP_MS = 50`, we always fall into the else. Why not commenting the if just to keep in mind that if we change the constants we need to mind this condition?

src/reachy_mini/motion/speech_tapper.py
```diff
@@ -147,16 +149,15 @@ def feed(self, pcm: NDArray[np.float32]) -> list[dict[str, float]]:
...
+                self.samples = np.concatenate([self.samples, hop])[-self.frame :]
+            else:
+                self.samples = hop[-self.frame :].copy()
+            if self.samples.size < self.frame:
```
FabienDanieau, line 160: same

FabienDanieau (commented):
Thanks for the PR! Just a small comment

src/reachy_mini/motion/speech_tapper.py
ai-ag2026: Addressed in 04493634: added the requested comment explaining that the branch starts empty with the current HOP_MS > FRAME_MS defaults and is retained for future overlapping-window configurations.

src/reachy_mini/motion/speech_tapper.py
ai-ag2026: Addressed in 04493634 as well: documented that the default hop fills a frame and the short-frame branch remains for future overlapping-window configurations.

ai-ag2026:
@FabienDanieau Both requested clarifications are implemented at `0449363479704ca4d7166a6bac8b24752a81d825`; 24 focused tests and Ruff pass. Ready for re-review.

FabienDanieau (approved):
perfect thanks!

PR #1260, opened 2026-07-10
perf(sdk): disable WebSocket compression on loopback connections

ai-ag2026:
## Motivation

`WSClient` connects to the daemon's `/ws/sdk` endpoint via
`websockets.sync.client`, which negotiates permessage-deflate compression
by default. The daemon streams full robot state over this socket at 50 Hz.
When the SDK client and the daemon run on the same host — the common case
for on-robot Python apps, where both processes live on the CM4 — compression
buys nothing (there's no network link to save bandwidth on) but still costs
real CPU to deflate/inflate every frame.

We hit this in production on a CM4-class Reachy Mini: pinning
`compression=None` on our application-side WebSocket client was one of
several changes in a CPU optimization pass that brought steady-state load
from ~275% to ~87.7% (a 3.1x reduction). Since every Python SDK consumer
that talks to a local daemon pays this cost by default, it seemed worth
fixing upstream in the SDK client itself rather than in each app.

## Change

- `src/reachy_mini/io/ws_client.py`: added `_is_loopback_host()`, which
  recognizes `localhost`, `127.0.0.1`, and `::1` (case-insensitive, trimmed).
  `WSClient.__init__` now passes `compression=None` to `ws_sync.connect()`
  when connecting to a loopback host, and leaves the library default
  (`"deflate"`) untouched for any other host — so remote/cross-host
  connections are unaffected.
- No new configuration surface was added; this mirrors the one existing
  connection call site (`WSClient(host, port)` in `reachy_mini.py`) and
  there's currently no options/config pattern on `WSClient` to hook into,
  so autodetection seemed more in keeping with the existing code than
  introducing a new constructor parameter.

## Test evidence

New file `tests/unit_tests/test_ws_client.py` (14 tests): parametrized
coverage of `_is_loopback_host()` for loopback and non-loopback inputs
(including the `0.0.0.0` edge case, which is intentionally *not* treated as
loopback), plus three tests that mock `ws_sync.connect` and assert the
exact `compression` kwarg passed for `localhost`, `127.0.0.1`, and a remote
IP.

Commands run in a `uv`-managed Python 3.11 venv inside the branch worktree:

```
uv venv --python 3.11 .venv
uv sync --group dev
.venv/bin/python -m pytest --collect-only tests/unit_tests/test_ws_client.py -q   # 14 collected, no collection errors
.venv/bin/python -m pytest tests/unit_tests/test_ws_client.py -v                  # 14 passed
.venv/bin/ruff check src/reachy_mini/io/ws_client.py tests/unit_tests/test_ws_client.py   # All checks passed
.venv/bin/ruff format --check src/reachy_mini/io/ws_client.py tests/unit_tests/test_ws_client.py  # 2 files already formatted
.venv/bin/mypy src/reachy_mini/io/ws_client.py             # Success: no issues found
.venv/bin/python -m pytest tests/unit_tests -q --ignore=tests/unit_tests/test_wifi_connect_retry.py
```

Full-suite result: **21 failed, 318 passed, 20 skipped** — identical set of
21 pre-existing failures as the baseline run *before* this change (same
test names; all attributable to sandbox/environment gaps unrelated to this
diff — e.g. `test_placo.py::test_load_kinematics` needs the `placo` extra,
several `test_daemon.py`/`test_wireless.py`/`test_app.py` cases need the
`mujoco` extra or real audio/network hardware). Passed count increased by
exactly 14, matching the new test file. `test_wifi_connect_retry.py` was
excluded from both baseline and post-change runs because it fails to
*collect* in this environment (`ModuleNotFoundError: No module named
'nmcli'`, an optional Linux-only extra not installed) — unrelated to this
change.

No physical robot was used for benchmarking; the CPU evidence above is a
pointer to our existing production measurement, not a fresh benchmark.

## Breaking changes

None. Behavior for non-loopback hosts is byte-for-byte unchanged (still
negotiates `permessage-deflate`). For loopback hosts, the wire protocol is
unaffected — WebSocket compression is a transport-layer negotiation; peers
that don't offer/support it simply fall back to uncompressed frames, which
is exactly what happens today when *either* side declines. The daemon-side
`FastAPI`/`uvicorn` WebSocket endpoint was not touched.

### DIFF
--- a/src/reachy_mini/io/ws_client.py
+++ b/src/reachy_mini/io/ws_client.py
@@ -35,6 +35,16 @@ from reachy_mini.io.protocol import (
 
 logger = logging.getLogger(__name__)
 
+# Hosts that resolve to the local machine. On loopback, permessage-deflate
+# compression only adds CPU overhead (there's no network bandwidth to save),
+# so we skip negotiating it for these.
+_LOOPBACK_HOSTS = {"localhost", "127.0.0.1", "::1"}
+
+
+def _is_loopback_host(host: str) -> bool:
+    """Return True if `host` is a loopback address (localhost/127.0.0.1/::1)."""
+    return host.strip().lower() in _LOOPBACK_HOSTS
+
 
 class WSClient(AbstractClient):
     """WebSocket client for Reachy Mini."""
@@ -72,8 +82,11 @@ class WSClient(AbstractClient):
         self._heartbeat = threading.Event()
 
         uri = f"ws://{host}:{port}/ws/sdk"
+        # permessage-deflate buys nothing on loopback (no network to save
+        # bandwidth on) and costs CPU at the daemon's 50 Hz state stream rate.
+        compression = None if _is_loopback_host(host) else "deflate"
         try:
-            self._ws = ws_sync.connect(uri)
+            self._ws = ws_sync.connect(uri, compression=compression)
         except (OSError, websockets.exceptions.InvalidHandshake, TimeoutError) as e:
             raise ConnectionError(f"Failed to connect to {uri}: {e}") from e
 
--- a/src/reachy_mini/io/ws_client.py
+++ b/src/reachy_mini/io/ws_client.py
@@ -4,6 +4,7 @@ Connects to the daemon's /ws/sdk endpoint and provides cached state,
 fire-and-forget commands, and task request/progress tracking.
 """
 
+import ipaddress
 import logging
 import threading
 import time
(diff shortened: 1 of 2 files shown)


src/reachy_mini/io/ws_client.py
```diff
@@ -35,6 +35,16 @@
 
 logger = logging.getLogger(__name__)
 
+# Hosts that resolve to the local machine. On loopback, permessage-deflate
+# compression only adds CPU overhead (there's no network bandwidth to save),
+# so we skip negotiating it for these.
+_LOOPBACK_HOSTS = {"localhost", "127.0.0.1", "::1"}
+
+
+def _is_loopback_host(host: str) -> bool:
```
FabienDanieau, line 39: Why not using the standard lib?
```
import ipaddress
def _is_loopback_host(host: str) -> bool:
    h = host.strip().strip("[]").lower()
    if h in {"localhost"}:
        return True
    try:
        return ipaddress.ip_address(h).is_loopback
    except ValueError:
        return False
```

FabienDanieau (commented):
Good catch! Just a minor comment to make it more robust

src/reachy_mini/io/ws_client.py
ai-ag2026: Implemented in 907e9c33 using stdlib `ipaddress.ip_address(...).is_loopback`, while preserving `localhost` and bracketed IPv6 handling. Added loopback/malformed-host coverage; 17 focused tests pass.

ai-ag2026:
@FabienDanieau Your robustness suggestion is implemented at `907e9c33d4bf6f9ab7c75b2a36fe8b0a99a75f29` using stdlib `ipaddress` with IPv4/IPv6/malformed-host coverage; 17 focused tests and Ruff pass. Ready for re-review.

FabienDanieau (approved):
Thanks a lot!

PR #1266, opened 2026-07-13
Expose the apps control API over WebRTC (JSON-RPC 2.0 relay + SDK)

### LINKED ISSUE #1252: Expose apps API to webrtc
For instance the conv app can be remotely controlled. https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/436
But this won't be possible from the mobile app that relies on the webrtc connection.

FabienDanieau:
Implements #1252 — expose the apps control API over WebRTC as **JSON-RPC 2.0**, with the daemon acting as a relay. One control protocol for driving/observing on-robot apps over the WebRTC DataChannel (and `/ws/sdk`).

## What
- **`io/jsonrpc.py`** — shared JSON-RPC 2.0 envelope (request/response/notification). One deliberate deviation: the stable machine-branchable error string lives in `error.data.reason`; `error.code` stays an int.
- **`apps/jsonrpc_server.py`** — small `/rpc` WebSocket server an app mounts (`@method`, `broadcast` / `broadcast_threadsafe`).
- **`daemon/jsonrpc_relay.py`** — routes by namespace: `apps.*` handled locally against the `AppManager` (`start`/`stop`/`status`/`install`); everything else relayed to the running app's `/rpc` over one persistent WS, whose reader also fans the app's notifications out to every connected client.
- Transport wiring: `Backend._handle_webrtc_message` branches on `looks_like_jsonrpc`; `Daemon` sets up the relay on the main loop.
- **Robot-lock coexistence** so a phone can drive a local app without losing its session: `start_app(keep_remote=True)` + `RobotAppLock.acquire_local_keeping_remote`, and the central signaling relay accepts a peer as **control-only** while a local app holds the robot.
- **TS SDK** (`ts/lib/reachy-mini.ts`): `rpcCall(method, params, {timeoutMs})` + `onNotification(method, cb)`.
- Fix: `custom_app_url` resolution now works for **editable (`-e`) installs** (importlib fallback), so the relay can reach the app's `/rpc` in dev.

## Tests
`test_jsonrpc_relay`, `test_jsonrpc_server`, `test_robot_app_lock`, `test_central_signaling_relay`, `test_custom_app_url` — all green; ruff + mypy clean.

## Companion PRs
- conversation app: `1252-rpc-over-webrtc` (app `/rpc` surface)
- mobile app: `feat/robot-conversation-api` (drives the robot over this API)

⚠️ The TS SDK (`rpcCall`/`onNotification`) needs publishing before the mobile app can consume it from npm.


pierre-rouanet (commented):
1. Overall approach seems like the right one to me! I'll dig into implementation details and leave comments directly on the code.
2. My only concern is that it kind of duplicates some of the legacy custom command system. But I'd actually vote for using the JSON-RPC approach everywhere and remove this legacy commands in another PR later. Basically: ```legacy commands get wrapped as robot.* JSON-RPC methods, and the legacy {"type": …} grammar is deprecated, then deleted. JSON-RPC stays.```

src/reachy_mini/apps/sources/local_common_venv.py
```diff
@@ -153,6 +154,41 @@ def get_app_python(
...
+    )
+    if site_packages and site_packages.exists():
+        main_file = site_packages / app_name / "main.py"
+        if main_file.exists():
+            return main_file
+
+    try:
+        spec = importlib.util.find_spec(app_name)
+    except (ImportError, ValueError):
+        spec = None
```
pierre-rouanet: Why not returning None directly here?

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -2401,10 +2182,27 @@ def _threadsafe_handler(peer_id: str, message: str) -> None:
...
+
     def _handle_webrtc_message(self, peer_id: str, message: str) -> None:
         def send(resp: dict[str, Any]) -> None:
             self._send_webrtc_response(peer_id, resp)
 
+        # JSON-RPC frames go to the app relay; legacy {"type": ...} commands
+        # fall through to command_adapter. Both share the DataChannel.
+        if self._jsonrpc_handler is not None:
+            try:
+                obj = json.loads(message)
```
pierre-rouanet, line 2416: This can be quite slow on the cm4. Can we rely on pydantic here instead (where json ser/deser is done in rust much more efficiently)?

src/reachy_mini/apps/jsonrpc_server.py
```diff
@@ -0,0 +1,156 @@
...
+logger = logging.getLogger(__name__)
+
+Handler = Callable[[dict[str, Any]], Union[Any, Awaitable[Any]]]
+# Unbounded so the decorator preserves each handler's exact signature (keeps
+# --disallow-untyped-decorators happy); the handler is cast to Handler on
+# registration.
+F = TypeVar("F")
+
+
+class JsonRpcServer:
```
pierre-rouanet, line 50: Here as well, it'd be nice to rely on pydantic for ser/deser. I'm not sure it's feasible though.

src/reachy_mini/daemon/jsonrpc_relay.py
```diff
@@ -0,0 +1,280 @@
...
+
+        if req.is_notification:
+            # Fire-and-forget: forward verbatim, nothing to correlate.
+            await ws.send(req.model_dump_json())
+            return
+
+        self._counter += 1
+        relay_id = f"relay-{self._counter}"
+        original_id = req.id
+        self._pending[relay_id] = reply
```
pierre-rouanet, line 180: Claude says (I've not yet dived enough in implementation to be sure):

> self._pending[relay_id] = reply          # registered BEFORE the await
frame["id"] = relay_id
await ws.send(json.dumps(frame))          # ← yields; reader task can run here
self._pending_original_id[relay_id] = original_id   # registered AFTER the await
> 
> The reader (_on_app_frame, same loop) matches on _pending then does _pending_original_id.pop(rid, None). On a loopback WS the app can answer during await ws.send(...); the reader then finds _pending set but _pending_original_id not yet set, so pop(...) returns None. The response goes back with id=null, the SDK's _handleRpcMessage requires data.id != null, drops it → the caller only rejects at the 20 s timeout despite an instant reply.
> 
> Fix: register _pending_original_id[relay_id] before the send (mirror _pending), and pop it in the except:
> 
> self._pending[relay_id] = reply
self._pending_original_id[relay_id] = original_id
frame["id"] = relay_id
try:
    await ws.send(json.dumps(frame))
except Exception as e:
    self._pending.pop(relay_id, None)
    self._pending_original_id.pop(relay_id, None)
    raise JsonRpcError(...) from e

src/reachy_mini/daemon/jsonrpc_relay.py
```diff
@@ -0,0 +1,280 @@
...
+    make_result,
+    parse_request,
+)
+
+Reply = Callable[[dict[str, Any]], None]
+
+_CONNECT_TIMEOUT_S = 5.0
+
+
+class JsonRpcRelay:
```
pierre-rouanet, line 47: Same for the relay. I'd feel better if we could rely on pydantic for parsing.

src/reachy_mini/daemon/jsonrpc_relay.py
```diff
@@ -0,0 +1,280 @@
...
+    async def handle(self, raw: str, reply: Reply) -> None:
+        """Route one incoming frame; ``reply`` sends a response to its sender."""
+        try:
+            req = parse_request(raw)
+        except JsonRpcError as e:
+            reply(error_from_exc(None, e))
+            return
+
+        try:
+            if req.namespace == "apps":
```
pierre-rouanet, line 90: I feel the "apps" field confusing. Could we use "apps_manager" instead for instance?

pierre-rouanet (approved):
I've left the two minor comments @FabienDanieau . It is mergeable in the current state for me.

CI fix in #1273

PR #1267, opened 2026-07-15
feat(media): speaker EQ to correct the head-shell "boxy" resonance

FabienDanieau:
## Rationale

Audio from Reachy Mini's speaker sounds a bit "boxy": the plastic head shell acts as a small enclosure that colors the output, a low-mid resonance (bass boom) plus a muffled 1-8 kHz presence range. This adds a fixed graphic-EQ correction on the daemon's speaker output to compensate.

## What it does

- Inserts a GStreamer [`equalizer-10bands`](https://gstreamer.freedesktop.org/documentation/equalizer/equalizer-10bands.html) on **every** speaker output branch, `GStreamerAudio` (voice push + file playback) and `GstMediaServer` (WebRTC + file playback), always **after the wobbler tee**, so head motion is still driven by the uncorrected signal.
- Gains come from the daemon config (`speaker_eq_gains` in `daemon_config.json`, reusing the store added in #1254), falling back to a tested default. **All-zero gains bypass the equalizer** (pipeline byte-identical to before).
- The default cuts the head-cavity bass boom and lifts the muffled presence range.

## Calibration

The default gains were derived offline from a **differential** measurement (speaker response *with* vs *without* the shell, so mic/amp/room coloration cancels) using [hifiscan](https://github.com/erdewit/hifiscan). The reusable tool lives in `src/reachy_mini/tools/speaker_eq_calibration/` (mirroring `camera_calibration/`), it is not used at runtime.

Speaker EQ calibration

*Top: measured speaker response, shell on vs off. Bottom: the shell coloration (16 kHz and 44.1 kHz measurements agree) and the applied 10-band correction, its mirror image.*

## Tuning

Override or disable via `~/.config/reachy_mini/daemon_config.json` (Wireless: `/home/pollen/.config/...`):

```json
{ "speaker_eq_gains": [0.0, -13.21, -5.55, -4.28, -4.32, 5.80, 4.65, 4.90, 3.41, 0.0] }
```

Ten dB values, one per band; all `0.0` disables the EQ. Restart the daemon to apply. Documented in *Advanced Media Controls → Speaker equalization* and a *Troubleshooting → Vision & Audio* entry.

## Testing

- New unit tests (`test_speaker_eq.py`, `test_speaker_eq_calibration.py`): gains resolution + config validation, `equalizer-10bands` element build (default active / all-zero no-op), and the calibration band math. `uv run pytest` green (pre-existing WEBRTC/network-dependent tests aside).
- Default gains derived from averaged hifiscan sweeps (shell on/off, at both 16 kHz and 44.1 kHz) on a **Reachy Mini Lite**.
- To A/B: set `speaker_eq_gains` to ten zeros and restart to hear the uncorrected output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -288,6 +288,3 @@ ts/*.d.ts
 ts/*.d.ts.map
 ts/host/src/**/*.d.ts.map
 !ts/host/src/assets/svg.d.ts
-
-# Speaker EQ calibration measurements (hifiscan exports)
-corr_*
--- a/.gitignore
+++ b/.gitignore
@@ -288,3 +288,6 @@ ts/*.d.ts
 ts/*.d.ts.map
 ts/host/src/**/*.d.ts.map
 !ts/host/src/assets/svg.d.ts
+
+# Speaker EQ calibration measurements (hifiscan exports)
+corr_*
--- a/src/reachy_mini/daemon/startup_app_config.py
+++ b/src/reachy_mini/daemon/startup_app_config.py
@@ -8,6 +8,7 @@ be set over the REST API instead of only via a CLI flag.
 
 import json
 import logging
+import math
 from pathlib import Path
 
 import platformdirs
@@ -16,6 +17,18 @@ logger = logging.getLogger(__name__)
 
 _KEY = "startup_app"
 _EQ_KEY = "speaker_eq_gains"
+# equalizer-10bands accepts per-band gains in [-24, +12] dB.
+_EQ_GAIN_MIN, _EQ_GAIN_MAX = -24.0, 12.0
+
+
+def _is_valid_gain(value: object) -> bool:
+    """Return True for a finite real number within the equalizer dB range."""
+    return (
+        isinstance(value, (int, float))
+        and not isinstance(value, bool)
+        and math.isfinite(value)
+        and _EQ_GAIN_MIN <= value <= _EQ_GAIN_MAX
+    )
 
 
 def _config_path() -> Path:
@@ -46,14 +59,15 @@ def get_startup_app() -> str | None:
 def get_speaker_eq_gains() -> list[float] | None:
     """Return the 10 speaker-EQ band gains (dB), or None if unset/invalid.
 
-    Invalid values (wrong length, non-numeric) are treated as unset so the
-    caller falls back to its built-in default.
+    Invalid values (wrong length, non-numeric, NaN/inf, or outside the
+    equalizer-10bands [-24, +12] dB range) are treated as unset so the caller
+    falls back to its built-in default.
     """
     value = _read().get(_EQ_KEY)
     if (
         isinstance(value, list)
         and len(value) == 10
-        and all(isinstance(x, (int, float)) and not isinstance(x, bool) for x in value)
+        and all(_is_valid_gain(x) for x in value)
     ):
         return [float(x) for x in value]
     return None
--- a/src/reachy_mini/daemon/startup_app_config.py
+++ b/src/reachy_mini/daemon/startup_app_config.py
@@ -63,13 +63,24 @@ def get_speaker_eq_gains() -> list[float] | None:
     equalizer-10bands [-24, +12] dB range) are treated as unset so the caller
     falls back to its built-in default.
     """
-    value = _read().get(_EQ_KEY)
+    config = _read()
+    if _EQ_KEY not in config:
+        return None
+    value = config[_EQ_KEY]
     if (
         isinstance(value, list)
         and len(value) == 10
         and all(_is_valid_gain(x) for x in value)
     ):
         return [float(x) for x in value]
+    # Present but malformed: warn so the user knows their values were ignored.
+    logger.warning(
+        "Ignoring invalid '%s' in daemon config (need 10 finite dB gains in "
+        "[%g, %g]); using the built-in defaults.",
+        _EQ_KEY,
+        _EQ_GAIN_MIN,
+        _EQ_GAIN_MAX,
+    )
     return None
 
 
--- a/src/reachy_mini/daemon/startup_app_config.py
+++ b/src/reachy_mini/daemon/startup_app_config.py
@@ -15,6 +15,7 @@ import platformdirs
 logger = logging.getLogger(__name__)
 
 _KEY = "startup_app"
+_EQ_KEY = "speaker_eq_gains"
 
 
 def _config_path() -> Path:
@@ -40,6 +41,22 @@ def get_startup_app() -> str | None:
     """Return the persisted startup app name, or None if unset."""
     value = _read().get(_KEY)
     return value if isinstance(value, str) else None
+
+
+def get_speaker_eq_gains() -> list[float] | None:
+    """Return the 10 speaker-EQ band gains (dB), or None if unset/invalid.
+
+    Invalid values (wrong length, non-numeric) are treated as unset so the
+    caller falls back to its built-in default.
+    """
+    value = _read().get(_EQ_KEY)
+    if (
+        isinstance(value, list)
+        and len(value) == 10
+        and all(isinstance(x, (int, float)) and not isinstance(x, bool) for x in value)
+    ):
+        return [float(x) for x in value]
+    return None
 
 
 def set_startup_app(name: str | None) -> None:
--- a/src/reachy_mini/daemon/startup_app_config.py
+++ b/src/reachy_mini/daemon/startup_app_config.py
@@ -8,7 +8,6 @@ be set over the REST API instead of only via a CLI flag.
 
 import json
 import logging
-import math
 from pathlib import Path
 
 import platformdirs
@@ -22,11 +21,15 @@ _EQ_GAIN_MIN, _EQ_GAIN_MAX = -24.0, 12.0
 
 
 def _is_valid_gain(value: object) -> bool:
-    """Return True for a finite real number within the equalizer dB range."""
+    """Return True for a real number within the equalizer dB range.
+
+    The range comparison also rejects NaN and infinities (they compare False)
+    and oversized ints (exact int/float compare, so no OverflowError) without
+    converting the value.
+    """
     return (
         isinstance(value, (int, float))
         and not isinstance(value, bool)
-        and math.isfinite(value)
         and _EQ_GAIN_MIN <= value <= _EQ_GAIN_MAX
     )
 
--- a/src/reachy_mini/media/audio_base.py
+++ b/src/reachy_mini/media/audio_base.py
@@ -30,7 +30,11 @@ from reachy_mini.media.audio_control_utils import (
     init_respeaker_usb,
 )
 from reachy_mini.media.audio_doa import AudioDoA
-from reachy_mini.media.audio_utils import resolve_speaker_eq_gains
+from reachy_mini.media.audio_utils import (
+    has_reachymini_asoundrc,
+    resolve_speaker_eq_gains,
+)
+from reachy_mini.media.device_detection import get_audio_device
 from reachy_mini.media.gstreamer_utils import get_sample, handle_default_bus_message
 
 gi.require_version("Gst", "1.0")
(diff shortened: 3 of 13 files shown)


src/reachy_mini/media/audio_base.py
```diff
@@ -43,6 +44,26 @@
...
+    """
+    gains = resolve_speaker_eq_gains()
+    if not any(gains):
+        return None
+    eq = Gst.ElementFactory.make("equalizer-10bands")
+    if eq is None:
+        logger.warning("equalizer-10bands unavailable; skipping speaker EQ")
+        return None
+    for i, gain in enumerate(gains):
+        eq.set_property(f"band{i}", float(gain))
```
alozowski, line 73: GStreamer calls this a [10-band IIR equalizer](https://gstreamer.freedesktop.org/documentation/equalizer/index.html), and its [1.28.3 code applies all ten bands to the same audio](https://github.com/GStreamer/gstreamer/blob/1.28.3/subprojects/gst-plugins-good/gst/equalizer/gstiirequalizer.c#L770-L814). I tested the default values with 16 kHz audio, at 3,770 Hz, the EQ raised the level by 7.87 dB, so I think this extra gain can make loud audio clip. Could we reduce the level before the EQ or add a limiter after it? Should we also add an automated test with loud input audio?

src/reachy_mini/tools/speaker_eq_calibration/calibrate.py
```diff
@@ -0,0 +1,261 @@
...
+        coloration(f) = resp_on(f) - resp_off(f)
+
+    and the correcting gain is its negative. We recenter by the median of the
+    trusted bands (the two measurements differ by an arbitrary overall level we
+    must not apply as broadband gain), clamp to the element's range, and zero
+    bands above ``zero_above`` (untrusted treble and/or beyond the voice-path
+    Nyquist). ``max_boost`` caps positive gains (noisy differentials can ask for
+    unrealistic boosts that only add hiss and eat headroom).
+    """
+    gains = bin_to_bands(freqs, resp_off_db - resp_on_db)  # = -coloration
```
alozowski, line 100: This code takes the average for each frequency range and uses it directly as one EQ gain. However, [each GStreamer band covers a range of nearby frequencies](https://github.com/GStreamer/gstreamer/blob/1.28.3/subprojects/gst-plugins-good/gst/equalizer/gstiirequalizer.c#L683-L697), and [the code applies all the bands to the same audio](https://github.com/GStreamer/gstreamer/blob/1.28.3/subprojects/gst-plugins-good/gst/equalizer/gstiirequalizer.c#L805-L813). I think, could we calculate the values from the real output of the GStreamer EQ, then measure again with the EQ enabled to confirm the result?

src/reachy_mini/daemon/startup_app_config.py
```diff
@@ -42,6 +43,22 @@ def get_startup_app() -> str | None:
...
+    """Return the 10 speaker-EQ band gains (dB), or None if unset/invalid.
+
+    Invalid values (wrong length, non-numeric) are treated as unset so the
+    caller falls back to its built-in default.
+    """
+    value = _read().get(_EQ_KEY)
+    if (
+        isinstance(value, list)
+        and len(value) == 10
+        and all(isinstance(x, (int, float)) and not isinstance(x, bool) for x in value)
```
alozowski: Python's json parser accepts `NaN` and infinity values, while GStreamer only accepts EQ gains from [-24 dB to +12 dB](https://gstreamer.freedesktop.org/documentation/equalizer/equalizer-10bands.html). A value can therefore pass this check and still fail when it is applied to GStreamer. Could we reject `NaN`, infinity, and values outside this range? Should we also add tests for them?

src/reachy_mini/media/audio_gstreamer.py
```diff
@@ -316,6 +321,7 @@ def _build_audiosink_tee_bin(self) -> Gst.Bin:
         queue_speaker = Gst.ElementFactory.make("queue")
         ac_speaker = Gst.ElementFactory.make("audioconvert")
         ar_speaker = Gst.ElementFactory.make("audioresample")
+        eq_speaker = make_speaker_eq(self.logger)
```
alozowski, line 324: When the code cannot find Reachy Mini Audio, it sends the sound to the default output instead. The Reachy EQ is still active, so it may change the sound of a different speaker and both playback paths have this behavior. Could we add the EQ only after we have confirmed that the selected output is Reachy Mini Audio?

tests/unit_tests/test_speaker_eq.py
```diff
@@ -0,0 +1,65 @@
...
+    assert startup_app_config.get_speaker_eq_gains() == [float(i) for i in range(10)]
+
+    cfg.write_text(json.dumps({"speaker_eq_gains": [1.0, 2.0, 3.0]}))  # wrong length
+    assert startup_app_config.get_speaker_eq_gains() is None
+
+    cfg.write_text(json.dumps({"speaker_eq_gains": "nope"}))  # not a list
+    assert startup_app_config.get_speaker_eq_gains() is None
+
+
+def test_make_speaker_eq_default_active(monkeypatch: pytest.MonkeyPatch) -> None:
```
alozowski, line 105: These tests check the EQ values and the EQ element by themselves, they don't create either real playback path. The tests could therefore pass even if the EQ is missing or connected in the wrong place. Could we add a test that builds the playback path and confirms that the EQ is used for non-zero values and skipped when all values are zero?

src/reachy_mini/daemon/startup_app_config.py
```diff
@@ -42,6 +43,22 @@ def get_startup_app() -> str | None:
...
+    ):
+        return [float(x) for x in value]
+    return None
```
alozowski, line 87: If `speaker_eq_gains` contains an invalid value, this returns `None`, then `resolve_speaker_eq_gains()` loads the built-in values, but the user doesn't see a warning. They may think that their custom values were applied. Could we log a warning before returning `None`?

src/reachy_mini/media/audio_utils.py
```diff
@@ -30,6 +30,30 @@
 import numpy as np
 import numpy.typing as npt
 
+# Speaker EQ gains (dB) correcting the head-shell "boxy" resonance, derived
```
alozowski: The constant name and the calibration `README` already explain this. Should we remove this three-line comment and keep the full explanation in the README? My idea is to keep all comments short in scripts, because it's easier to keep them updated

src/reachy_mini/tools/speaker_eq_calibration/calibrate.py
```diff
@@ -0,0 +1,261 @@
...
+         corr_on.pkl corr_off.pkl corr_on_16k.pkl corr_off_16k.pkl \
+         -o docs/assets/speaker_eq_calibration.png
+
+Caveats: a 10-band graphic EQ has fixed octave bands, so a narrow resonance may
+fall between centers; it corrects the magnitude response only (not time-domain
+ringing or mechanical buzz); and on the 16 kHz voice path only bands <= 8 kHz
+carry signal (top bands are zeroed via ``--zero-above``).
+"""
+
+from __future__ import annotations
```
alozowski: This file has no forward type references, so it doesn't need `from __future__ import annotations`, could we remove it?

.gitignore
```diff
@@ -287,4 +287,7 @@ ts/lib/*.d.ts.map
 ts/*.d.ts
 ts/*.d.ts.map
 ts/host/src/**/*.d.ts.map
-!ts/host/src/assets/svg.d.ts
\ No newline at end of file
+!ts/host/src/assets/svg.d.ts
+
+# Speaker EQ calibration measurements (hifiscan exports)
+corr_*
```
alozowski: Could we limit it to the folder where the calibration files are created? Or do we need this gitignore rule in general? I find it too broad and it can catch unrelated files whose names start with `corr_`

alozowski (approved):
Tested this PR on a physical Reachy Mini Wireless with the standard audio firmware v2.1.4 from this PR: https://github.com/pollen-robotics/reachy_mini/pull/1242

This PR works on the robot, the EQ made a clear audible difference at the same mixer level, so approving it with some comments

src/reachy_mini/media/audio_base.py
FabienDanieau: Fixed in 6bf9e1b5. \`make_speaker_eq\` now returns a float-internal bin \`capsfilter(F32LE) → equalizer-10bands → audiodynamic → audioconvert\`: the EQ runs in float and a memoryless \`audiodynamic\` limiter (threshold 0.9, ratio 0.05) caps peaks *before* the int conversion, so boosts cannot clip for any gains, loudness preserved. Added a pipeline test that pushes a full-scale 3770 Hz tone through the bin and asserts the output does not exceed full scale (5d051a25).

src/reachy_mini/tools/speaker_eq_calibration/calibrate.py
FabienDanieau: Acknowledged: the per-octave average is an approximation and the IIR bands do overlap/sum. Kept the heuristic (offline tool, validated by ear and on your hardware) and documented the verify-by-re-measure loop in the README (1edd87d5): apply the gains, re-measure with the EQ active, confirm the residual is flatter. Noted a closed-loop fit against the element's real summed response as a future improvement.

src/reachy_mini/daemon/startup_app_config.py
FabienDanieau: Fixed in 28ad6f5e. \`get_speaker_eq_gains\` now also requires each value to be finite (\`math.isfinite\`) and within the equalizer's [-24, +12] dB range; otherwise the config is treated as unset. Added tests for NaN, Infinity, and both out-of-range bounds.

src/reachy_mini/media/audio_gstreamer.py
FabienDanieau: Fixed in a625714d. \`make_speaker_eq\` is now gated on the same detection the sink builders use (\`has_reachymini_asoundrc() or get_audio_device("Sink") is not None\`), so on a fallback output (autoaudiosink) the EQ is skipped. The check lives in the shared helper, so both playback paths are covered. Added a test for the skip case.

tests/unit_tests/test_speaker_eq.py
FabienDanieau: Done in 5d051a25. Added tests that build the real speaker tee bin and assert the \`equalizer-10bands\` is present for non-zero gains and absent for all-zero gains, plus the full-scale no-clip pipeline test.

src/reachy_mini/daemon/startup_app_config.py
FabienDanieau: Fixed in 7d294f83. When \`speaker_eq_gains\` is present but invalid we now log a warning before falling back to the defaults; an absent key stays silent (the normal path). Covered by a test.

src/reachy_mini/media/audio_utils.py
FabienDanieau: Done in 8403864d: reduced to a single line pointing at the tool README.

src/reachy_mini/tools/speaker_eq_calibration/calibrate.py
FabienDanieau: Done in fefdfe34: removed; no forward references needed on Python 3.12.

.gitignore
FabienDanieau: Done in b2e90d28: you're right removed.

src/reachy_mini/daemon/startup_app_config.py
```diff
@@ -8,13 +8,27 @@
...
+# equalizer-10bands accepts per-band gains in [-24, +12] dB.
+_EQ_GAIN_MIN, _EQ_GAIN_MAX = -24.0, 12.0
+
+
+def _is_valid_gain(value: object) -> bool:
+    """Return True for a finite real number within the equalizer dB range."""
+    return (
+        isinstance(value, (int, float))
+        and not isinstance(value, bool)
+        and math.isfinite(value)
```
alozowski: A json integer can be larger than a python float. I tested `10**400`, and this line raised `OverflowError` instead of rejecting the configuration. Could we check the allowed range before calling `math.isfinite`, or avoid the conversion, and add this case to the validation test?

src/reachy_mini/media/audio_base.py
```diff
@@ -43,6 +48,57 @@
...
+    in_caps = Gst.ElementFactory.make("capsfilter")
+    limiter = Gst.ElementFactory.make("audiodynamic")
+    out_conv = Gst.ElementFactory.make("audioconvert")
+    if in_caps is None or limiter is None or out_conv is None:
+        logger.warning("audiodynamic unavailable; speaker EQ runs without a limiter")
+        return eq
+    in_caps.set_property("caps", Gst.Caps.from_string("audio/x-raw,format=F32LE"))
+    # Hard-knee compressor with a low ratio near full scale = a brickwall limiter.
+    limiter.set_property("threshold", 0.9)
+    limiter.set_property("ratio", 0.05)
```
alozowski: I'm thinking would it be better to set the ratio to `0.0`?

src/reachy_mini/daemon/startup_app_config.py
FabienDanieau: Fixed in a413ed09. Dropped the `math.isfinite` call (which raised `OverflowError` on `10**400`) and rely on the range comparison instead: it rejects NaN, infinities and oversized ints via an exact int/float compare, no conversion. Added a `10**400` test case.

src/reachy_mini/media/audio_base.py
FabienDanieau: Good call, done in 2be857fd. Set the ratio to 0.0 so the limiter is a true brickwall at the threshold; the output can no longer exceed full scale for any gains.

PR #1268, opened 2026-07-15
feat: add adaptive face tracking filter

jrubiosainz:
## Summary

Ports the adaptive smoothing ideas from pollen-robotics/reachy_mini_conversation_app#339 to the SDK's current daemon-side face-tracking architecture, following the maintainer guidance in that PR.

The SDK now filters normalized face centers before they reach `look_at_image_pose()`, reducing small target jitter while retaining a faster response to meaningful movement.

## Changes

- Add a small stateful `AdaptiveCenterFilter` with:
  - EMA smoothing for ordinary motion
  - a faster alpha for large frame-to-frame movement
  - a radial dead zone for sub-threshold jitter
  - explicit reset behavior on target loss or tracking pause
- Wire the filter into the actual `FaceTracker` detector path
- Preserve filter state across transient misses, but reset it when the tracker drops its target association
- Remove the backend's duplicate center deadband so target filtering has one owner
- Add focused unit tests for slow smoothing, fast adaptation, dead-zone behavior, and reacquisition/reset

## Architecture notes

The old conversation-app patch also changed camera loop rate, tracking gain, loss delay, and neutral interpolation. Those parts are intentionally not ported:

- camera IPC and detector pacing now own capture cadence;
- the daemon already owns transient-loss holding and sustained-loss recentering;
- calibrated `look_at_image_pose()` plus the existing tracking weight replace the old raw translation/rotation gain.

## Validation

- `pytest tests/unit_tests/test_face_tracking.py tests/unit_tests/test_head_tracking_backend.py -q` — 19 passed
- Ruff lint — passed
- Ruff format check — passed
- targeted mypy — passed
- `git diff --check` — passed

Supersedes pollen-robotics/reachy_mini_conversation_app#339.

### DIFF
--- a/src/reachy_mini/vision/face_tracking.py
+++ b/src/reachy_mini/vision/face_tracking.py
@@ -41,34 +41,15 @@ class FaceObservation:
     timestamp: float
 
 
-class AdaptiveCenterFilter:
-    """Smooth normalized face centers while remaining responsive to large motion.
+class _AdaptiveCenterFilter:
+    """Internal face-center smoother with fixed tracking parameters."""
 
-    Small changes are held inside a dead zone. Outside it, an exponential moving
-    average uses ``alpha`` for ordinary motion and ``fast_alpha`` when consecutive
-    raw observations move farther than ``movement_threshold``.
-    """
+    _ALPHA = 0.3
+    _FAST_ALPHA = 0.6
+    _MOVEMENT_THRESHOLD = 0.15
+    _DEAD_ZONE = 0.02
 
-    def __init__(
-        self,
-        alpha: float = 0.3,
-        fast_alpha: float = 0.6,
-        movement_threshold: float = 0.15,
-        dead_zone: float = 0.02,
-    ) -> None:
-        """Configure the filter in normalized image coordinates."""
-        if not 0.0 < alpha <= 1.0:
-            raise ValueError("alpha must be in (0, 1]")
-        if not 0.0 < fast_alpha <= 1.0:
-            raise ValueError("fast_alpha must be in (0, 1]")
-        if movement_threshold < 0.0:
-            raise ValueError("movement_threshold must be non-negative")
-        if dead_zone < 0.0:
-            raise ValueError("dead_zone must be non-negative")
-        self._alpha = alpha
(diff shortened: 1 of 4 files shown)


src/reachy_mini/vision/face_tracking.py
```diff
@@ -41,6 +41,61 @@ class FaceObservation:
     timestamp: float
 
 
+class AdaptiveCenterFilter:
```
alozowski: This filter is internal to `FaceTracker`, could we make it private and keep its fixed parameters internal instead of exposing unused constructor configuration and validation?

tests/unit_tests/test_face_tracking.py
```diff
@@ -44,6 +49,51 @@ def test_to_observation_normalizes_nose_and_roll() -> None:
     assert obs.roll == float(np.arctan2(2.0, 2.0))
 
 
+def test_adaptive_filter_smooths_ordinary_motion() -> None:
```
alozowski: These tests still pass if `FaceTracker` stops applying or resetting the filter. Could we replace them with a tracker-level test that feeds a detection sequence and checks the emitted observations?

alozowski (changes requested):
Nice work! I tested on a physical wireless and it works. I left two small comments on keeping the filter internal and covering its `FaceTracker` wiring

src/reachy_mini/vision/face_tracking.py
jrubiosainz: Addressed in cbeac7d9: the filter is now a private implementation detail with fixed internal parameters and no public configuration/validation surface.

tests/unit_tests/test_face_tracking.py
jrubiosainz: Addressed in cbeac7d9: replaced the direct filter tests with FaceTracker-level detection-sequence tests that assert emitted smoothing/dead-zone behavior and reset after target loss.

alozowski (approved):
Approved!

jrubiosainz:
Thanks a lot @alozowski !! So happy to contribute! 🎉

FabienDanieau (approved):
Smooth, thanks!

PR #1270, opened 2026-07-16
jsonrpc: model responses/notifications with pydantic + TypeAdapter parsing

pierre-rouanet:
Refactor of the JSON-RPC envelope introduced in #1266, targeting that PR's branch (`1252-expose-apps-api-to-webrtc`), not `main`.

## Motivation

The envelope modeled only *requests* as pydantic. Responses, errors and notifications were assembled as hand-rolled dicts, and the daemon relay classified/rewrote app frames by poking raw dicts (`json.loads` + `dict.get`, `obj["id"] = ...`). That's inconsistent with the repo's own protocol idiom, which parses both directions of the legacy protocol through `TypeAdapter` unions (`command_adapter`, `server_msg_adapter`). This aligns the new protocol with that pattern and removes the manual ser/deser.

## What changed

- **`io/jsonrpc.py`** — model every frame (`RpcSuccess` / `RpcErrorResponse` / `RpcNotification` / `RpcErrorObj`) and add an `AnyRpcInbound` `TypeAdapter` union (`parse_inbound`) so inbound frames are parsed and classified on required fields, not shape-sniffed.
- The `make_*` / `error_from_exc` helpers keep returning `dict` (call sites unchanged) but now build through the models, so the wire shape has a single source of truth. The `parse_error` (-32700) vs `invalid_request` (-32600) distinction is preserved via a shared `_decode` step.
- **`daemon/jsonrpc_relay.py`** — replace the raw-dict correlation in `_relay_to_app` / `_on_app_frame` with `model_copy` + `parse_inbound`; drop the now-unused `json` import.

## Tests

Added envelope round-trip and `parse_inbound` classification/rejection tests. ruff + mypy clean; the jsonrpc unit tests are green.

### DIFF
--- a/src/reachy_mini/apps/jsonrpc_server.py
+++ b/src/reachy_mini/apps/jsonrpc_server.py
@@ -32,8 +32,8 @@ from fastapi import FastAPI, WebSocket, WebSocketDisconnect
 
 from reachy_mini.io.jsonrpc import (
     JsonRpcError,
+    RpcNotification,
     error_from_exc,
-    make_notification,
     make_result,
     parse_request,
 )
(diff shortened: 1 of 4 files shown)


FabienDanieau (approved):
Nice, pydantic is a better way indeed

PR #1271, opened 2026-07-16
ci: test audio, camera & media with virtual devices

FabienDanieau:
## What

CI ran `pytest -m 'not audio and not video and not wireless'`, so the **entire audio + video + media surface was skipped** (it assumed real hardware). This makes that surface testable in CI using **userspace virtual devices** — no `/dev/snd`, no `/dev/video`, no kernel modules (which the GitHub Azure-kernel runner doesn't provide).

## How it's simulated

| Area | Fake | Runs in CI |
|------|------|:---:|
| Audio playback/record | PulseAudio `module-null-sink` + ALSA `type pulse` `.asoundrc` → real `alsasrc`/`alsasink` fast-path (`has_reachymini_asoundrc`) | ✅ |
| "Did audio actually play?" | null-sink `.monitor` loopback → assert captured signal is non-silent | ✅ |
| ReSpeaker/XMOS + DoA | fake `usb.core.Device` (in-memory register file) behind `ReSpeaker` | ✅ |
| Camera frames | existing `ipc_video_source` fixture: `videotestsrc → unixfdsink` into the IPC socket `GStreamerCamera` reads | ✅ |
| UDP sim camera | `GStreamerUDPCamera` → localhost RTP loopback → receive | ✅ |
| WebRTC signalling helpers | mocked websocket | ✅ |

## Changes

- **Marker split**: `respeaker` (needs the physical XMOS/USB board), `loopback` (needs a virtual audio loopback). Board/DoA/config tests stay gated behind `respeaker` for on-hardware runs.
- **Fake ReSpeaker** (`conftest.py`): the whole board is reached through one `ctrl_transfer` seam, so an in-memory `(resid, cmdid)` register file reproduces the vendor protocol (writes round-trip through reads). Drives `read_values` decode, `apply_audio_config` verify, `AudioDoA.get_DoA`, and the REST audio-config router — no hardware.
- **New tests**: `test_play_sound_reaches_sink` (loopback energy assertion), simulated XMOS/DoA/config tests, `test_udp_camera.py` (RTP round-trip), `test_webrtc_utils.py` (producer-list helpers).
- **CI** (`pytest.yml`): install `gstreamer1.0-alsa`, `gstreamer1.0-plugins-bad`, `libasound2-plugins`, `pulseaudio`; add a step that brings up the null-sink virtual card and runs the audio + video tests.
- **Coverage** (`allure.yml`): bring up the same card + include audio/video so the published `reachy_mini` coverage reflects the media modules (was 35% with all of this skipped).
- `volume_control` fixture skips cleanly when no real ALSA card is present (see below).

## Not covered (and why — all need something the runner can't provide)

- **`media_server.py`** (daemon) — its constructor requires the `webrtcsink` element (gst-plugins-rs).
- **`webrtc_client_gstreamer.py` / `central_signaling_relay.py`** — full WebRTC peer + signalling loopback (separate, larger effort).
- **volume control** & **daemon v4l2 capture** — need a real ALSA card / `/dev/video` (`snd-*`/`v4l2loopback` kernel modules aren't on the Azure runner).

## Verified

Locally in an `ubuntu:24.04` container mirroring the runner, and confirmed green on CI:
audio media (6) + playback loopback (1) + simulated XMOS/DoA/config + camera IPC (5) + UDP camera + webrtc_utils (4). macOS leg unchanged.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -45,8 +45,11 @@ jobs:
         run: |
           # The Azure runner kernel ships snd-dummy/snd-aloop in the
           # linux-modules-extra package, which isn't installed by default.
+          # depmod -a rebuilds modules.dep so modprobe can find the new .ko.
           sudo apt-get update
           sudo apt-get install -y "linux-modules-extra-$(uname -r)"
+          find "/lib/modules/$(uname -r)" \( -name 'snd-dummy*' -o -name 'snd-aloop*' \) -print || true
+          sudo depmod -a
           sudo modprobe snd-dummy id=respeaker
           arecord -l
           uv run python -c "from reachy_mini.media.audio_utils import write_asoundrc_to_home; write_asoundrc_to_home()"
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -43,6 +43,10 @@ jobs:
       - name: Set up dummy ALSA card (Ubuntu)
         if: matrix.os == 'ubuntu-latest'
         run: |
+          # The Azure runner kernel ships snd-dummy/snd-aloop in the
+          # linux-modules-extra package, which isn't installed by default.
+          sudo apt-get update
+          sudo apt-get install -y "linux-modules-extra-$(uname -r)"
           sudo modprobe snd-dummy id=respeaker
           arecord -l
           uv run python -c "from reachy_mini.media.audio_utils import write_asoundrc_to_home; write_asoundrc_to_home()"
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -22,7 +22,7 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         uses: awalsh128/cache-apt-pkgs-action@acb598e5ddbc6f68a970c5da0688d2f3a9f04d05  # v1.6.0
         with:
-          packages: libcairo2-dev libgirepository1.0-dev python3-gi python3-gi-cairo gstreamer1.0-plugins-base gstreamer1.0-plugins-good gir1.2-gst-plugins-base-1.0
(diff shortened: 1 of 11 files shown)


tests/unit_tests/test_audio.py
```diff
@@ -222,6 +223,60 @@ def test_DoA(backend: MediaBackend) -> None:
...
 
 
+@pytest.mark.audio
+@pytest.mark.loopback
+def test_play_sound_reaches_sink(audio_loopback: None) -> None:
+    """Play a sound and confirm real, non-silent audio reaches the sink.
+
+    Unlike test_play_sound (which only checks no-exception, and on a
+    discard-only sink can't tell), this records a virtual loopback of the sink
+    while playing, so silence means playback never reached the speaker path.
```
pierre-rouanet, line 251: Could we even check that the recorded data is "close enough" of the wake_up wav?

.github/workflows/pytest.yml
```diff
@@ -36,3 +36,29 @@ jobs:
...
+      # socket GStreamerCamera reads. Board/wireless tests stay excluded.
+      # Ubuntu only.
+      - name: Run audio + video tests on a virtual sound card (Ubuntu)
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          pulseaudio -D --exit-idle-time=-1 --disable-shm \
+            --load="module-native-protocol-unix socket=/tmp/pulse.sock" \
+            --load="module-null-sink sink_name=vspeaker"
+          export PULSE_SERVER=unix:/tmp/pulse.sock
+          pactl info | grep "Server Name"
```
pierre-rouanet: Claude pointed out that this sanity check is not done in allure.yml. It was recommending to consider a small composite action (.github/actions/virtual-sound-card) so the two can't diverge further.

pierre-rouanet (approved):
Really nice! Thanks @FabienDanieau.
I just add two minor comments. But feel free to merge it directly.

PR #1272, opened 2026-07-16
jsonrpc relay: fix reply-correlation race (caller id restored as null)

pierre-rouanet:
Stacked on #1270 (base: `refactor/jsonrpc-pydantic-envelope`) so this shows only the race fix. Addresses finding #1 from the review of #1266.

## The bug

`JsonRpcRelay._relay_to_app` registered the relay-id → original-id mapping (`_pending_original_id`) **after** `await ws.send(...)`:

```python
self._pending[relay_id] = reply            # set before await
...
await ws.send(outgoing.model_dump_json())  # yields to the loop
self._pending_original_id[relay_id] = req.id   # set AFTER await  <-- race
```

The reader task runs on the same loop and can process the app's reply **during** that `await` (a fast reply on the loopback `/rpc` WS). At that moment `_on_app_frame` finds the pending entry but `_pending_original_id.pop(relay_id, None)` returns `None`, so the response is handed back with `id=null`. The SDK requires `data.id != null` to match a pending call, drops it, and the caller only rejects at its timeout (~20 s) — even though the app answered instantly.

## The fix

Record both correlation entries **before** the send (and drop both if the send fails). No wire change.

## Test

Adds `test_original_id_recorded_before_send`, a deterministic regression test asserting the original-id mapping is present at send time. Verified it fails with the old ordering and passes with the fix. ruff + mypy clean; jsonrpc unit tests green.

### DIFF
--- a/src/reachy_mini/daemon/jsonrpc_relay.py
+++ b/src/reachy_mini/daemon/jsonrpc_relay.py
@@ -172,8 +172,13 @@ class JsonRpcRelay:
 
         self._counter += 1
         relay_id = f"relay-{self._counter}"
-        original_id = req.id
+        # Record both correlation entries BEFORE the await: the reader task runs
+        # on this same loop and can process the app's reply *during*
+        # ``ws.send`` (a fast reply on the loopback WS). If `original_id` were
+        # set after the await, that reply would restore id=None and the caller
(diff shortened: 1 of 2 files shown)


FabienDanieau (approved):
Good catch!

PR #1273, opened 2026-07-17
ci(npm-publish): normalize PEP 440 version to semver before npm version

pierre-rouanet:
## Problem

The `publish` job fails on any push/PR touching `ts/**` (or the workflow file) whenever `main`'s `pyproject.toml` version is a PEP 440 dev/prerelease form:

```
Computed version: 1.10.0.dev0-main.<sha> (event=pull_request, release_tag=rc)
npm error Invalid version: 1.10.0.dev0-main.<sha>
```

`pyproject.toml` is the single source of truth for the version and uses **PEP 440** (`1.10.0.dev0`, `1.9.0rc1`). The workflow fed that string straight to `npm version`, which requires **semver** and rejects those forms. It only ever produced valid semver for a clean `X.Y.Z`.

This currently red-lights CI on PR #1266 (first PR to touch `ts/**` since `main` was bumped to `1.10.0.dev0`).

Note: PRs never actually publish — the `npm publish` steps are gated to `release`/`push`. The PR path only builds + `npm pack --dry-run`. The version-compute step just happened to be fatal for everyone.

## Fix

Normalize the PEP 440 prerelease/dev suffix into a single dot-separated semver prerelease before `npm version`:

| pyproject (PEP 440) | npm (semver) |
|---|---|
| `1.10.0.dev0` | `1.10.0-dev.0` |
| `1.9.0rc1` | `1.9.0-rc.1` |
| `1.9.0rc1.dev0` | `1.9.0-rc.1.dev.0` |
| `1.10.0` | `1.10.0` (unchanged) |

The `-main.<sha>` build marker is appended as extra dot-separated identifiers when a prerelease already exists, so we never emit a second `-` (which semver forbids).

## Verification

Ran the exact workflow step (YAML-loaded, heredoc indentation preserved) across `{1.10.0.dev0, 1.9.0rc1, 1.10.0}` × `{pull_request, release, push}` — all produce valid semver and pass `npm version`. Confirmed the old string `1.10.0.dev0-main.<sha>` still fails `npm version`, i.e. the fix directly addresses the failure.

Once merged, PR #1266's `publish` job picks up the fixed workflow automatically (it reads the version from the PR-merged-into-`main` tree).

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -76,11 +76,40 @@ jobs:
         id: ver
         run: |
           py_ver=$(grep -E '^version' ../pyproject.toml | head -1 | sed -E 's/version *= *"([^"]+)"/\1/')
+          # pyproject.toml uses PEP 440 (e.g. 1.10.0.dev0, 1.9.0rc1); npm
+          # requires semver. Normalize the prerelease/dev suffix into a
+          # single dot-separated semver prerelease before `npm version`:
+          #   1.10.0.dev0    -> 1.10.0-dev.0
+          #   1.9.0rc1       -> 1.9.0-rc.1
+          #   1.9.0rc1.dev0  -> 1.9.0-rc.1.dev.0
+          #   1.10.0         -> 1.10.0 (unchanged)
+          semver=$(python3 - "$py_ver" <<'PY'
+          import re, sys
+          v = sys.argv[1]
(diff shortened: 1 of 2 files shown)


FabienDanieau (approved):
Thanks for the fix! One follow-up while we're here: could you add a note to `RELEASE.md` that `main` intentionally carries a PEP 440 `.dev0` version, so any tooling that reads the version from `pyproject.toml` must handle non-semver forms?

PR #1274, opened 2026-07-17
test: WebRTC coverage — signalling relay, client, media server

FabienDanieau:
## What

Follow-up to #1271, targeting the biggest remaining media coverage gaps — all WebRTC:

| module | before |
|---|---|
| `central_signaling_relay.py` | 25% (523 missing) |
| `webrtc_client_gstreamer.py` | 0% (321 stmt) |
| `media_server.py` | 26% (570 missing) |

## How

**Phase 1 — `central_signaling_relay.py` (pure async, no plugin).** It's a client relay (SSE ↔ local webrtcsink websocket), no GStreamer. Table-driven tests of both message dispatchers + session-id translation (welcome/startSession with single-session + robot-lock gating/peer/endSession, both legs), `_set_state`, `get_relay_status` — driven with `asyncio.run` + a send-journal stub.

**Phases 2 & 3 use the prebuilt `libgstrswebrtc.so`** (webrtcsrc/webrtcsink) from the public `reachy-mini-desktop-app` repo — fetched in CI with no secret (pinned commit + sha256), via a `webrtc-plugin` composite action:

- **Phase 2 — `webrtc_client_gstreamer.py`**: construct the real `GstWebRTCClient` and unit-test the hardware-free surface (daemon REST helpers, transceiver/deep-element/bus callbacks, `_iterate_gst`, `_apply_resolution`, `_configure_webrtcsrc`).
- **Phase 3 — real loopback + daemon build**: a plain `gst-launch-1.0 webrtcsink` pipeline stands in for the robot (raw A/V + embedded signalling server); the real client discovers the producer, negotiates over WebRTC, and receives a **video frame + non-silent audio** — exercising the live paths. Plus `GstMediaServer(MUJOCO)` construction, covering the daemon's whole pipeline-building path.

The `webrtc`-marked tests are gated by a plugin skip-guard, so they run where the `.so` is present and skip cleanly elsewhere (e.g. macOS).

**CI hygiene:** the apt package list is factored into an `install-system-deps` composite action shared by `pytest.yml` and `allure.yml` (single source of truth, keeps the apt cache shared); `allure.yml` also loads the plugin so the webrtc tests count toward coverage; and a few unused imports were removed from existing tests (`ruff --select F401`, which `tests/*` is otherwise excluded from).

## Verified

In an `ubuntu:24.04` / GStreamer-1.24 container mirroring the runner: the `.so` loads (`webrtcsrc`/`webrtcsink` register, libs resolved); Phase 1+2 = 69 passed; loopback stable over repeated runs (~2.3 s); the full `-m webrtc` selection = 23 passed after fetching the `.so`. Green on the runner (ubuntu + macOS + mypy + ruff).

## Not included (possible follow-ups)

- Live audio/DoA over WebRTC stays `wireless` (needs a real robot). The `.so` is x86_64 Linux, so the webrtc step is Ubuntu-only.
- Bigger remaining coverage levers: `central_signaling_relay`'s async I/O (`_connect_and_relay`, `_handle_central_sse`, send/reconnect loops) and a full `GstMediaServer`↔client daemon loopback for `media_server.py`'s runtime paths.


pierre-rouanet (commented):
LGTM! 

My only concern is that if webrtcsink fails to be installed/setup correctly in CI, tests will be skipped instead of shown as failed. 
Maybe we can do something like:
```python
def require_webrtc_plugin(*elements: str) -> None:
    """Skip — or, when the job requires it, fail — unless every element registers."""
    Gst.init([])
    missing = [e for e in elements if Gst.ElementFactory.find(e) is None]
    if not missing:
        return

    reason = (
        "gst-plugins-rs webrtc plugin unavailable — missing element(s): "
        f"{', '.join(missing)} (libgstrswebrtc.so not loaded)"
    )
    if os.environ.get("REACHY_MINI_REQUIRE_WEBRTC") == "1":
        # This job declared the plugin present; a skip here would hide a broken
        # plugin and silently drop coverage, so fail loudly instead.
        pytest.fail(reason, pytrace=False)
    pytest.skip(reason, allow_module_level=True)
```

What do you think @FabienDanieau ?

pierre-rouanet (approved):

PR #1275, opened 2026-07-17
test: coverage sweep — daemon backend + low-% routers/utils

FabienDanieau:
## What

Raises CI unit-test coverage for the biggest remaining low-% modules, following the same constraint as the prior coverage PRs (#1271, #1274): **all new tests run in the default pytest job with no robot hardware and no real network** (mocks, `MockupSimBackend`, FastAPI `TestClient` + `dependency_overrides`, `tmp_path`).

Overall coverage was **44%** before this PR.

## Scope

**Deep — `daemon/backend/abstract.py`** (was 34%, 650 missing — the biggest single lever), driven through a new headless `MockupSimBackend(use_audio=False)` fixture:
- `process_command` dispatcher (all Set*/Get*/motor/gravity/wobbling/head-tracking/recording/volume/audio-config/state/version/hardware-id/log/restart/update branches).
- Move + audio upload handlers (chunk reassembly, out-of-order / count-mismatch / unknown-encoding / too-many-slots / TTL eviction) and play/cancel move+audio.
- Target setters, IK compose, recording buffer, present-pose getters, async goto/wake/sleep wrappers.

**Broad — near-0% routers and pure-logic utils:**
- Routers via `TestClient` + `dependency_overrides` (a new seam): `cache` (0%), `update` (0%), `volume`, `hf_auth`.
- Utils: `hf_space` Space→AppInfo normalization, `wireless_version/*` (0% — build-command / semver / pypi / install-source / async update), `apps/sources/hf_auth` (OAuth sessions + token fns), `daemon/utils`.

## Shared fixtures (`conftest.py`)

- `sim_backend` — headless `MockupSimBackend(use_audio=False)` with a mock media server.
- `router_app` — mounts one router on a bare FastAPI app and wires backend/daemon stubs via `dependency_overrides` + `app.state`.

## Not in this PR

Deferred (heavier, a clean follow-up): `volume_control_linux` ALSA parsing, `startup_check`, routers `state`/`move`/`media`/`logs`. Skipped: `volume_control_macos/windows` (ctypes/pycaw — won't import on Linux CI), `parse_urdf_for_kinematics` (would just exercise Placo's solver, already covered).

## Testing

- ~240 new tests, one file per target module (one commit per file).
- Full suite: `610 passed, 64 deselected` under the default CI selection (`-m 'not audio and not video and not wireless and not webrtc'`); ruff check + format clean.
- Coverage % lands on the CI Allure/coverage run (can't be measured locally — PyGObject + coverage.py crash).

### DIFF
--- a/src/reachy_mini/utils/wireless_version/update_available.py
+++ b/src/reachy_mini/utils/wireless_version/update_available.py
@@ -4,6 +4,7 @@ For now, this only checks if a new version of "reachy_mini" is available on PyPI
 """
 
 import json
+import re
 from importlib.metadata import distribution, version
 
 import requests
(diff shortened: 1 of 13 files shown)


pierre-rouanet:
While reviewing, the new `test_wireless_version.py` surfaced two real `ValueError` crashes on PEP 440 versions this repo actually produces — pushed a test+fix pair for each: `_semver_version` choking on `.dev0` (breaks `get_local_version` / update checks on any from-source install) and `get_pypi_version` comparing against a raw `info.version` string. Both hit dev/from-ref installs, not PyPI-wheel users. These touch `src/`, so split them out if you'd rather keep the sweep test-only; the `releases[-1]` ordering assumption is left as a noted follow-up.

pierre-rouanet (approved):
Added a few more tests on version (when getting a non semver version). The rest LGTM!

PR #1280, opened 2026-07-19
fix: publish live robot backend status

### LINKED ISSUE #1278: Robot backend status never publishes live readiness
### Description
`RobotBackend._status.ready` is initialized to `False`, while the control loop sets `self.ready`, but `get_status()` only refreshes `error` and `motor_control_mode`. As a result, `/api/daemon/status` can continue reporting `backend_status.ready: false` after the real backend is ready. `last_alive` is similarly left at its initial `None` value.

### Reproduction
On a DGX Spark ARM64 native daemon, the backend control loop was operating and the dashboard was usable, but a health check against `/api/daemon/status` failed because the serialized backend status remained not ready. Updating `get_status()` from `self.ready`, `self.should_stop`, and `self.last_alive` corrected the API response.

### Proposed fix
Refresh `ready` and `last_alive` in `RobotBackend.get_status()` and add a unit regression test that changes the live events after constructing the cached status object.

am423:
## Summary
- refresh cached robot backend readiness from the live lifecycle events
- publish the latest control-loop heartbeat through `last_alive`
- add a regression test for ready and stopping transitions

Closes #1278.

## Verification
- `uv run pytest tests/unit_tests/test_robot_backend_status.py -q` — passed
- `uv run ruff check <changed files>` — passed
- `uv run mypy src/reachy_mini/daemon/backend/robot/backend.py` — passed
- `git diff --check` — passed

The broader unit collection on this ARM64 host reaches an unrelated environment dependency: `test_wifi_connect_retry.py` imports optional `nmcli`; installing the complete wireless extra then requires the system `liblgpio` library. Hosted CI should exercise that environment-specific lane.

### DIFF
--- a/src/reachy_mini/daemon/backend/robot/backend.py
+++ b/src/reachy_mini/daemon/backend/robot/backend.py
@@ -321,6 +321,8 @@ class RobotBackend(Backend):
 
     def get_status(self) -> "RobotBackendStatus":
         """Get the current status of the robot backend."""
+        self._status.ready = self.ready.is_set() and not self.should_stop.is_set()
+        self._status.last_alive = self.last_alive
         self._status.error = self.error
         self._status.motor_control_mode = self.motor_control_mode
         return self._status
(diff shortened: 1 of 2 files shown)


pierre-rouanet (approved):
Thanks @am423!

PR #1281, opened 2026-07-19
fix: honor Hugging Face token path

### LINKED ISSUE #1279: OAuth token persistence ignores HF_HOME and HF_TOKEN_PATH
### Description
`exchange_code_for_token()` writes OAuth credentials directly to `~/.cache/huggingface/token`. This bypasses `huggingface_hub`'s configured `HF_HOME` and `HF_TOKEN_PATH` locations.

This breaks isolated deployments such as a systemd user service with a private `HF_HOME`: OAuth reports success, but subsequent hub operations look in the configured location and do not find the token.

### Proposed fix
Use `huggingface_hub.constants.HF_TOKEN_PATH`, create its parent directory, write the token with user-only permissions, and add a regression test with a temporary configured token path.

am423:
## Summary
- persist OAuth credentials to the path selected by `huggingface_hub`
- create the configured parent directory and restrict the token file to mode `0600`
- add an async regression test covering the complete token exchange persistence path

Closes #1279.

## Verification
- `uv run pytest tests/unit_tests/test_hf_auth.py -q` — passed
- `uv run ruff check <changed files>` — passed
- `uv run mypy src/reachy_mini/apps/sources/hf_auth.py` — passed
- `git diff --check` — passed

### DIFF
--- a/src/reachy_mini/apps/sources/hf_auth.py
+++ b/src/reachy_mini/apps/sources/hf_auth.py
@@ -6,10 +6,12 @@ import os
 import secrets
 import time
 from dataclasses import dataclass, field
+from pathlib import Path
 from typing import Any, Optional
 
 import aiohttp
 from huggingface_hub import HfApi, get_token, login, logout, whoami
+from huggingface_hub.constants import HF_TOKEN_PATH
 from huggingface_hub.errors import HfHubHTTPError
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


am423:
Addressed an independent security review finding in `3f693abd3154e5f5634e8d83c269fcd13cc62aa4`:

- token bytes are written only after a same-directory temporary file has mode `0600`
- the temporary file is flushed and `fsync`ed before atomic `os.replace`
- failures clean up the temporary file and return an OAuth persistence error
- regression coverage now verifies mode before replacement and verifies that a permission-enforcement failure retains no token bytes

Verification: 2 focused tests passed; Ruff, mypy, and `git diff --check` passed.

pierre-rouanet:
Maybe @cdussieux you want to have a look at this?

cdussieux (approved):
we might need to do the same thing with `.cache/huggingface` no ?
can be treated in another PR

PR #1282, opened 2026-07-20
Reduce sdk startup latency

alozowski:
Defers daemon-only imports and local camera readiness, reuses the media pipeline built during daemon construction, and sends the current daemon status immediately to new clients. Adds focused regression coverage for import isolation, camera lifecycle, media-server restart, and initial status ordering.


src/reachy_mini/io/__init__.py
```diff
@@ -1,9 +1,28 @@
...
+if TYPE_CHECKING:
+    from .ws_server import WSServer
 
 __all__ = [
     "WSClient",
     "WSServer",
 ]
+
+
+def __getattr__(name: str) -> object:
```
pierre-rouanet: I feel this is kind of hacky @alozowski . Is this really making a big difference on the robot? 
I mean most of the heavy dependencies are still required anyways no (eg. fastapi)?

pierre-rouanet (commented):
Except from the ws_server import defer that I'm a bit afraid of, rest seems good @alozowski  👌.
 But maybe we should double check with @FabienDanieau as this mostly touches media pipes.

src/reachy_mini/io/__init__.py
alozowski: Agree! I removed the lazy `__getattr__` and restored the normal WSServer export

alozowski:
Measured startup on the wireless with the conversation app from https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/506 (they aren't related, but I was curious to test them together), using seven samples for each case:
- when I started the app directly while the robot was already awake, the time until the app settings and event stream were ready went from 9.772 s to 7.641 s (21.8% faster).
- when I woke the robot and let the daemon start the app automatically, it went from 11.975 s to 10.089 s (15.8% faster).

I didn't use the first spoken word as a metric because it mostly depends on the remote backend

alozowski:
I added an isolated benchmark to test this branch with `python scripts/benchmark_startup.py` 
These results are medians of three runs without an active IPC camera source

#### macOS

| metric | baseline | pr_1282 |
|---|---:|---:|
| SDK import | 0.419 s | 0.335 s |
| Camera `open()` | 2.072 s | 0.0004 s |
| First camera `read()` | 0.025 s | 2.005 s |

#### Wireless

| metric | baseline | pr_1282 |
|---|---:|---:|
| SDK import | 3.382 s | 2.746 s |
| Camera `open()` | 2.011 s | 0.0086 s |
| First camera `read()` | 0.021 s | 2.001 s |

SDK import is about 19-20% faster and camera `open()` now starts the pipeline without waiting for a frame. The first `read()` waits only when necessary. This improves startup for apps that do not immediately use the camera, while time to the first frame remains roughly unchanged for apps that read immediately (can be discussed)

alozowski:
I removed lazy DoA discovery because the lock around USB reads could block cached daemon status reads. Making camera `open()` return immediately also exposed an open/close timing issue, so I fixed that and made sure a new WebSocket client cannot receive an older queued status after its fresh initial status

The current measured numbers:
#### macOS

| metric | baseline | pr_1282 |
|---|---:|---:|
| SDK import | 0.4467 s | 0.3616 s |
| Camera `open()` | 2.0747 s | 0.0005 s |
| First camera `read()` | 0.0253 s | 2.0038 s |
| Open + first read | 2.1000 s | 2.0043 s |

#### Wireless

| metric | baseline | pr_1282 |
|---|---:|---:|
| SDK import | 3.3729 s | 2.7460 s |
| Camera `open()` | 2.0109 s | 0.0086 s |
| First camera `read()` | 0.0206 s | 2.0006 s |
| Open + first read | 2.0315 s | 2.0093 s |

PR #1287, opened 2026-07-21
feat(control): add atomic non-evicting app start route

brygal1:
## Summary

- expose a dedicated Control route that starts a local app with `evict_remote=False`
- preserve the existing start route and its remote-eviction behavior
- cover successful delegation and busy-slot HTTP 400 behavior

This gives boot/recovery clients a fail-closed, atomic native operation. A separate lock-status preflight is insufficient because a remote WebRTC session can acquire the managed slot between the read and the existing evicting start request.

## Validation

- Ruff format/check on changed files
- 59 focused app-lock, startup watcher, and route tests

### DIFF
--- a/docs/source/API/openapi.json
+++ b/docs/source/API/openapi.json
@@ -239,6 +239,46 @@
         }
       }
     },
+    "/api/apps/start-app/{app_name}/no-evict": {
+      "post": {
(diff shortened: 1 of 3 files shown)


pierre-rouanet (approved):
Thanks @brygal1 !

Issue #1263, opened 2026-07-10
Make the local IPC camera feed framerate cap configurable


alozowski:
Currently the daemon serves the local camera IPC feed at 10 fps, hardcoded (`IPC_FPS` in `media_server.py`). The cap exists because every local client pays a conversion for each frame it is served, but the right value depends on the platform and on what the local apps actually do with the frames. Right now changing it means editing the source.

domotick:
Oh @alozowski just seen your issue! Yes, merge it quickly please 🙏

https://github.com/pollen-robotics/reachy_mini/issues/1285#issuecomment-5044338718

PR #1288, opened 2026-07-22
Cap FaceTracker detection rate independently of the IPC feed (hardening for #1263)

johannwest7012:
## Context

Follow-up to the discussion below. The original framing of this PR — that `FaceTracker` runs detection at *camera* rate and starves the 50 Hz motor control loop — was wrong. On current `main` (and on 1.9.0) the local IPC feed is already capped at 10 FPS (`media_server.py`'s `IPC_FPS`), and `FaceTracker` consumes that feed via `unixfdsrc`, so detection is already feed-bounded to ~10 Hz. The earlier python timer therefore just duplicated an existing cap. Thanks to @alozowski for catching that.

What still holds: once the IPC rate becomes configurable (#1263), `FaceTracker` will need its own independent rate limit so a faster feed doesn't drag detection — YuNet inference on the shared Python process/GIL — up with it.

## Change

Reframed as hardening for configurable IPC rates. Instead of the python timer, add a tracker-local `videorate` (`drop-only`) + framerate capsfilter right after `source`/`queue`, before the colour-convert chain — so frames above the target rate are dropped *before* the (possibly hardware) convert and the appsink pull, rather than pulled and detected on.

- Target rate is a module constant `_DETECTION_FPS = 10`, with a comment pointing at #1263 so it's a one-line change to read the IPC-rate config once that lands.
- `drop-only` so a feed slower than the target passes straight through without `videorate` duplicating frames to hit the rate.
- No-op against today's 10 FPS IPC feed; load-bearing only once the feed can run faster.

## Verification

`ruff check` / `ruff format`, `mypy`, and `tests/unit_tests/test_face_tracking.py` (9 passed) pass locally. Not yet exercised against a live camera feed on hardware — the element placement is standard, but a run on a wireless would be the real end-to-end check.

### DIFF
--- a/src/reachy_mini/vision/face_tracking.py
+++ b/src/reachy_mini/vision/face_tracking.py
@@ -27,6 +27,14 @@ logger = logging.getLogger(__name__)
 # Detector input width; smaller trades recall for CPU.
 _TRACKING_WIDTH = 320
 
+# FaceTracker's own detection-rate cap, enforced with a videorate element in the
+# GStreamer graph so surplus frames are dropped before colour-convert and the pull.
+# The local IPC feed already caps at 10 FPS today, so this is currently redundant;
+# it becomes load-bearing once the IPC rate is configurable
+# (see https://github.com/pollen-robotics/reachy_mini/issues/1263), at which point
+# this constant should read that config instead of being hardcoded.
+_DETECTION_FPS = 10
+
 
 @dataclass
 class FaceObservation:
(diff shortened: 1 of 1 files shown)


pierre-rouanet:
cc. @alozowski

alozowski:
Looking at the PR, but also cc @RemiFabre as a control loop expert

RemiFabre:
I'll look into this

alozowski:
Thank you @johannwest7012  for the detailed investigation and measurements!

I think the key point is where the pacing is applied. On current `main`, the local IPC feed is capped at 10 FPS, and FaceTracker consumes that feed. I confirmed approximately 10 Hz detection without this patch on a wireless, so the python timer duplicates the existing pacing

There is also an open issue to make the IPC rate configurable: [#1263](https://github.com/pollen-robotics/reachy_mini/issues/1263). When implemented, `FaceTracker` will need an independent limit. In that case, I'd prefer to restore a tracker-local gstreamer `videorate` before frame conversion, so unused frames are discarded earlier (cc @FabienDanieau)

Would you be open to adapting the PR in that direction and reframing it as hardening for configurable IPC rates?

alozowski:
TL;DR the python timer is redundant while IPC is fixed at 10 FPS, so I suggest to adapt this PR to cap `FaceTracker` at 10 Hz with gstreamer before conversion, merge it as hardening, and then safely make the shared IPC rate configurable in #1263

johannwest7012:
@alozowski Yeah can do! Happy to reframe as hardening for configurable IPC rates. I'll plan to drop the python timer and add a tracker-local videorate ! video/x-raw,framerate=10/1 right after the source/queue (before the convert chain), so surplus frames are dropped before colour-convert and the appsink pull, keeping FaceTracker's rate independent of the IPC feed.

Since [#1263](https://github.com/pollen-robotics/reachy_mini/issues/1263) is still open, I'll keep the target rate as a module-level constant (10 Hz) for now, with a comment pointing at [#1263](https://github.com/pollen-robotics/reachy_mini/issues/1263) so it's a one-line change to read the IPC-rate config once that lands. I'll push the updated PR.

PR #1292, opened 2026-07-23
1291 dont  broadcast when there are no clients

### LINKED ISSUE #1291: Don't _broadcast when there are no clients
Minor performance gain

RemiFabre:
Small fix to avoid useless CPU work when the client disconnects and the number of clients is 0.
This represents ~150 call_soon_threadsafe() calls per second saved, but after measuring this is very modest (like 1% of a core).


pierre-rouanet (approved):
Thanks @RemiFabre ! LGTM

PR #1286, opened 2026-07-21
feat(daemon): first wake-up wizard support (on latest main)

tfrere:
## Summary

Daemon-side support for the mobile **first wake-up wizard**, plus the surrounding robot-identity / session-ownership plumbing it relies on. Merged up to date with `main` (1.10 line): coexists with the JSON-RPC relay and visual head-tracking already in `main`.

## What's in here

**First wake-up wizard**
- `get_first_wake_up` / `set_first_wake_up` data-channel commands + a persisted, robot-wide flag (`utils/first_wake_up.py`) so the wizard only ever runs once.
- Live **pose stream** pushed over the unreliable/unordered `pose` data channel (`build_state_json` / `build_state_dict` as the single source of truth), so the wizard's 3D view is smooth and jitter-immune instead of polling `get_state`. Now also carries `face_target` so it stays consistent with main's head-tracking.

**Robot identity**
- Persisted, live **robot rename** (`utils/robot_name.py`): `get_robot_name` / `set_robot_name` commands; applied live to daemon status + central relay + mDNS, no restart, and wins over the `--robot-name` CLI default on next boot.
- Per-robot **BLE advertised name** + a **BLE IDENTIFY** command (plays `surprise.ogg`) to find a specific robot.
- `delete_hf_token` command to sign a robot out over the data channel.

**Session ownership**
- `RobotAppLock` (`daemon/robot_app_lock.py`): a single source of truth for the managed app slot (local Python app vs remote WebRTC session), with eviction/keep-remote semantics and a "became free" hook that resets the robot to a clean idle state so no app leaves it parked awake across sessions.
- `GET /api/daemon/robot-app-lock-status`.

**SDK (TS)**
- `getFirstWakeUp` / `setFirstWakeUp`, `getRobotName` / `setRobotName`, `deleteHfToken`, and live pose-stream consumption (with `face_target`).

**Design docs & tests**
- Design notes under `docs/design/`.
- Unit tests for each piece: first-wake-up flag, robot name, app lock, pose stream, media-server pose push, BLE identify proxy.

## Merge status
Up to date with `main`; the merge was resolved **additively** — both feature sets kept (our wake-up/identity commands **and** main's `set_head_tracking` / `get_tracked_face`). `build_state_dict` is the shared state builder for both the polled `get_state` and the pushed pose stream, and now includes `face_target`.

## Notes / test plan
- [ ] `pytest` unit suite
- [ ] SDK `tsc`
[...]

PR #1294, opened 2026-07-23
fix(daemon): reset robot to idle when the app slot becomes free

pierre-rouanet:
Extracted from #1286 as a standalone, self-contained correctness fix (part of splitting that PR into reviewable pieces — see the scope discussion there).

## What

When the shared `RobotAppLock` transitions to **FREE** — a remote WebRTC session ended/dropped, or a local Python app exited — the daemon now returns the robot to a clean idle state via `goto_sleep` (which ends with motors disabled, clearing `gravity_compensation`).

## Why

The daemon never reset motor state on session teardown, and clients aren't guaranteed to run a clean leave sequence: a crash, a buggy app, or a lost Wi-Fi link all skip it. Without this, whatever motor mode the last app left behind — notably `gravity_compensation` — survives into the next session, where the client's `ensureAwake()` sees `isAwake() == true`, skips the wake animation, and leaves the robot parked in a weird state.

Firing on **both graceful and abnormal teardown** is the whole point: those paths run no client-side cleanup.

## How

- `RobotAppLock` gains a `set_on_became_free_handler` hook, fired once per FREE transition (from `release_local`/`release_remote`), outside the mutex.
- The daemon wires `_on_robot_slot_free` → `backend.request_idle_reset()`, scheduled threadsafely on the backend loop.
- The reset is **cancelled** when a fresh command grabs the robot again (`_handle_webrtc_message`), and **skipped** when the robot is already limp (motors disabled).
- Unwired in `stop()` before the relay is torn down, so releasing the remote hold can't race the explicit shutdown `goto_sleep` with a second one.

## Tests

`tests/unit_tests/test_robot_app_lock.py` covers the became-free handler: fires on local/remote release, sees FREE state, not fired on no-op release or on eviction (REMOTE→LOCAL never passes through FREE), exception-swallowing, and clearing.

## Notes

- Purely additive to `main`; no behavior change for the existing lock callers.
- `--no-media` daemons have no backend loop and no remote sessions, so `request_idle_reset()` is a no-op there.

### DIFF
--- a/src/reachy_mini/apps/manager.py
+++ b/src/reachy_mini/apps/manager.py
@@ -153,11 +153,17 @@ class AppManager:
             if keep_remote:
                 self.daemon.robot_app_lock.acquire_local_keeping_remote(app_name)
             elif evict_remote:
-                await self.daemon.robot_app_lock.acquire_local_evicting_remote(
-                    app_name
-                )
+                await self.daemon.robot_app_lock.acquire_local_evicting_remote(app_name)
             elif not self.daemon.robot_app_lock.try_acquire_local(app_name):
                 raise RuntimeError("The robot app slot is already in use")
+
+            # We now own the app slot. Cancel any idle reset the daemon
+            # scheduled when the slot last became free: this local app runs its
+            # own wake/motion sequence and must not fight a stale goto_sleep.
+            # Threadsafe; no-op if the backend loop isn't up.
+            backend = getattr(self.daemon, "backend", None)
+            if backend is not None:
+                backend.cancel_idle_reset()
 
         try:
             # Get module name and Python path for subprocess execution.
@@ -263,9 +269,7 @@ class AppManager:
                 if self.current_app is not None:
                     if returncode == 0:
                         self.current_app.status.state = AppState.DONE
-                        self.logger.getChild("runner").info(
-                            f"App {app_name} finished"
-                        )
+                        self.logger.getChild("runner").info(f"App {app_name} finished")
                     else:
                         self.current_app.status.state = AppState.ERROR
                         error_msg = "\n".join(stderr_lines[-10:])  # Last 10 lines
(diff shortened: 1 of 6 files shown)


pierre-rouanet:
### Behavior note for reviewers: the idle reset fires *immediately* on release

Tracing the path, there's no grace period or debounce — the moment the lock frees, if the robot is awake it starts `goto_sleep` on the next backend-loop tick:

```
release_remote()/release_local() → _fire_became_free()      (sync, inside the lock)
  → Daemon._on_robot_slot_free()
    → backend.request_idle_reset()
      → loop.call_soon_threadsafe(_maybe_start_idle_reset)   (next loop tick)
        → goto_sleep()                                        (if awake)
```

`_maybe_start_idle_reset` only skips when `not ready`, `is_shutting_down`, or motors are already `Disabled`. Otherwise it goes to sleep right away. Two things I'd like a decision on before merge:

**1. No grace period for transient drops.** A brief Wi-Fi blip or a fast session hand-off releases the lock and immediately starts a `goto_sleep`. If the peer reconnects and sends a command, `_handle_webrtc_message` calls `_cancel_idle_reset()` — but by then the trajectory may already be mid-flight, and cancelling it (`CancelledError`) stops the move wherever it is, potentially leaving the robot in an **intermediate pose** rather than cleanly asleep or awake. On a flaky link this could read as "sleep-then-yank-back" thrash. A short cancellable debounce (schedule the reset ~1–2 s out) would avoid it. Or even quickly switching from an app to another should not trigger the sleep/wakeup.

**2. Only WebRTC commands cancel the reset — not a local app acquiring the slot.** `_cancel_idle_reset()` is wired only into `_handle_webrtc_message`. If the next owner is a **local Python app** (acquires via `try_acquire_local` / `acquire_local_*`), nothing on that path cancels an in-flight idle-reset `goto_sleep`, so the app's own wake sequence can contend with it. Note the docstring in `_maybe_start_idle_reset` claims "a fast reconnect (or a local app grabbing the robot) can cancel it" — but I don't see the local-acquire path actually calling `_cancel_idle_reset`, so either the wiring is missing or the comment overstates it.

What do you say @tfrere @cdussieux ?

PR #1296, opened 2026-07-23
feat(daemon): live pose stream over an unreliable data channel

pierre-rouanet:
Extracted from #1286 as a standalone feature (part of splitting that PR into reviewable pieces). Depends on nothing else in #1286; the mobile first-wake-up wizard is the main consumer, but the pose stream benefits any live 3D mirror / head-tracking view.

## What

A push-based pose stream: the daemon *pushes* the robot's present state to subscribed WebRTC peers over a dedicated **unreliable/unordered `pose` data channel** at ~30 Hz, instead of the client polling `get_state` over the reliable-ordered control channel.

## Why

Polling `get_state` for a live mirror is at the mercy of Wi-Fi round-trip latency and head-of-line blocking on the control channel, so the 3D view jitters. Pushing on a separate lossy channel is immune to both: a dropped frame is simply superseded by the next one 33 ms later rather than stalling the stream.

## How

- **Protocol** (`protocol.py`): `subscribe_pose` / `unsubscribe_pose` data-channel commands.
- **Backend** (`abstract.py`): `build_state_dict` / `build_state_json` become the single source of truth for both the polled `get_state` reply and the pushed frames; each pushed frame carries a monotonic `seq` so the client can drop stale/out-of-order frames. `get_state` now delegates to `build_state_dict` (same envelope, incl. `face_target`, plus per-motor `head_joint_positions` / `antennas_joint_positions`).
- **Media server** (`media_server.py`): a second per-peer `pose` channel (`ordered=false, max-retransmits=0`) created alongside the control channel, per-peer subscriber tracking, and a ~30 Hz `_push_pose` timer that only builds+sends a frame while at least one peer is subscribed (idle sessions cost nothing).
- **JS SDK** (`reachy-mini.ts`): label-based routing so the `pose` channel never becomes `_dc` nor gates session readiness; a `seq` high-water-mark stale-frame guard; and refcounted `subscribePose()` / `unsubscribePose()` (multiple local consumers share one daemon-side subscription), re-asserted on reconnect.

Backward-compatible: `hasattr` guards mean an older media server without the pose hooks simply falls back to `get_state` polling.

## Tests

- `tests/unit_tests/test_pose_stream.py` — backend dispatch (peer-scoped subscribe/unsubscribe, delegated to the media server) and the `build_state_json` envelope + monotonic `seq`.
- `tests/unit_tests/test_media_server_pose.py` — subscriber bookkeeping, `_push_pose` gating/broadcast, channel-close cleanup, idempotent timer arming (GStreamer/GLib edges mocked, mirroring `test_media_server_watchdog`).
- Local run: pose tests 14 passed; `ruff`/`mypy` clean; SDK `tsc` passes.

## Note for reviewers

`GstMediaServer.close()` reads `self._pose_push_source_id` unconditionally, so a `GstMediaServer` whose `__init__` raises before that attribute is set will raise an unraisable `AttributeError` in `__del__` → `close()` (harmless warning, but new noise). Cheap fix would be `getattr(self, "_pose_push_source_id", None)`. Kept as-is to stay a faithful extraction — flagging so it's decided in review.

## Scope note

This branch intentionally excludes two things that were bundled into the same upstream commit but belong to the onboarding/wizard work: the onboarding-dataset preloading in `recorded_move.py` and the BLE cue sourcing in `bluetooth_service.py`. Those will land with the wizard PR.

### DIFF
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -72,7 +72,9 @@ from reachy_mini.io.protocol import (
     StartUpdateCmd,
     StopRecordingCmd,
     SubscribeLogsCmd,
+    SubscribePoseCmd,
     UnsubscribeLogsCmd,
+    UnsubscribePoseCmd,
     UploadAudioChunkCmd,
     UploadAudioFinishCmd,
     UploadAudioStartCmd,
@@ -349,6 +351,12 @@ class Backend:
         # sends `subscribe_logs`, cancelled on `unsubscribe_logs` or
         # peer disconnect (the latter is wired in setup_media_server).
         self._log_tasks: dict[str, "asyncio.Task[None]"] = {}
+        # Monotonic sequence number stamped on every pushed pose frame so the
+        # client can drop out-of-order/stale frames that arrive on the
+        # unordered `pose` data channel (see `build_state_json`). Bumped only
+        # from the media server's GLib thread (single writer), so a plain int
+        # is safe.
+        self._pose_seq = 0
         # Asyncio loop on which the log-streaming tasks run. Captured
         # in setup_media_server alongside the WebRTC handler loop, so
         # cross-thread cleanup (peer disconnect fires from gstreamer's
@@ -1277,6 +1285,59 @@ class Backend:
         }
 
     # ------------------------------------------------------------------
+    # State snapshot (shared by get_state replies and the pose push)
+    # ------------------------------------------------------------------
+
+    def build_state_dict(self) -> dict[str, Any]:
+        """Build the present-state snapshot sent to clients.
+
+        Single source of truth for both the polled ``get_state`` reply and
+        the pushed pose stream (see :meth:`build_state_json`). All getters
+        read cached motor values (``get_last_position``), so this is cheap
+        and safe to call from a thread other than the motor loop.
+        """
+        return {
+            "head_pose": self.get_present_head_pose().tolist()
+            if self.current_head_pose is not None
+            else None,
+            "antennas": self.get_present_antenna_joint_positions().tolist()
+            if self.current_antenna_joint_positions is not None
+            else None,
+            # Per-motor joint positions (7 head incl. body yaw at [0], 2
+            # antennas), so WebRTC clients can check the physical pose motor
+            # by motor without the LAN /ws/sdk stream.
+            "head_joint_positions": self.get_present_head_joint_positions().tolist()
+            if self.current_head_joint_positions is not None
+            else None,
+            "antennas_joint_positions": self.get_present_antenna_joint_positions().tolist()
+            if self.current_antenna_joint_positions is not None
+            else None,
+            "body_yaw": self.get_present_body_yaw(),
+            "motor_mode": self.get_motor_control_mode().value,
+            "is_recording": self.is_recording,
+            "is_move_running": self.is_move_running,
+            "face_target": self.get_tracked_face().model_dump(),
+        }
+
+    def build_state_json(self) -> Optional[str]:
+        """Serialize the present state as the client-facing envelope.
+
+        Returns the ``{"state": ...}`` payload, or ``None`` if it can't be
+        built yet. Wired into the media server as the pose-stream provider
+        (``set_pose_provider``) so the daemon can *push* pose over the
+        unreliable/unordered ``pose`` data channel at a steady rate instead
+        of the client round-tripping ``get_state`` over Wi-Fi. Returning
+        ``None`` (e.g. before the first kinematics update, or during
+        shutdown) simply skips that tick.
+        """
+        try:
+            state = self.build_state_dict()
+        except Exception:
+            return None
+        self._pose_seq += 1
+        return json.dumps({"state": state, "seq": self._pose_seq})
+
+    # ------------------------------------------------------------------
     # Transport-agnostic command processing
     # ------------------------------------------------------------------
 
@@ -1447,20 +1508,7 @@ class Backend:
             send_response({"status": "ok", "command": "set_automatic_body_yaw"})
 
         elif isinstance(cmd, GetStateCmd):
-            state = {
-                "head_pose": self.get_present_head_pose().tolist()
-                if self.current_head_pose is not None
-                else None,
-                "antennas": self.get_present_antenna_joint_positions().tolist()
-                if self.current_antenna_joint_positions is not None
-                else None,
-                "body_yaw": self.get_present_body_yaw(),
-                "motor_mode": self.get_motor_control_mode().value,
-                "is_recording": self.is_recording,
-                "is_move_running": self.is_move_running,
-                "face_target": self.get_tracked_face().model_dump(),
-            }
-            send_response({"state": state})
+            send_response({"state": self.build_state_dict()})
 
         elif isinstance(cmd, GetVersionCmd):
             from importlib.metadata import version
@@ -1599,6 +1647,13 @@ class Backend:
             self._start_log_subscription(peer_id, send_response)
         elif isinstance(cmd, UnsubscribeLogsCmd):
             self._cancel_log_subscription(peer_id)
+
+        elif isinstance(cmd, SubscribePoseCmd):
+            self._set_pose_subscription(peer_id, True)
+            send_response({"status": "ok", "command": "subscribe_pose"})
+        elif isinstance(cmd, UnsubscribePoseCmd):
+            self._set_pose_subscription(peer_id, False)
+            send_response({"status": "ok", "command": "unsubscribe_pose"})
 
         elif isinstance(cmd, RestartDaemonCmd):
             # Ack BEFORE triggering the restart: the WebRTC transport
@@ -2063,6 +2118,18 @@ class Backend:
         if task is not None and not task.done():
             task.cancel()
 
+    def _set_pose_subscription(self, peer_id: Optional[str], enabled: bool) -> None:
+        """Enable/disable the pushed pose stream for a peer.
+
+        Delegated to the media server, which owns the per-peer ``pose`` data
+        channels and the push timer. No-op without a peer id (non-multiplexed
+        transports) or on an older media server lacking the hook.
+        """
+        if peer_id is None or self._media_server is None:
+            return
+        if hasattr(self._media_server, "set_pose_subscription"):
+            self._media_server.set_pose_subscription(peer_id, enabled)
+
     def _on_peer_disconnect(self, peer_id: str) -> None:
         """Cancel any per-peer state when the WebRTC peer goes away.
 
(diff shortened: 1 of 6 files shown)


src/reachy_mini/media/media_server.py
```diff
@@ -1739,6 +1762,93 @@ def _setup_data_channel(self, peer_id: str, webrtcbin: Gst.Element) -> None:
...
+    def _push_pose(self) -> bool:
+        """Broadcast the latest pose frame to every open pose channel.
+
+        Runs on the GLib main loop. Returns ``True`` to stay scheduled;
+        it's a cheap no-op while nobody is subscribed or no provider is set,
+        so we keep it alive for the media server's lifetime rather than
+        churning the timer as peers come and go.
+        """
+        if not self._pose_subscribers or self._pose_provider is None:
+            return True
```
A-Mahla: Once the first peer connects this timer stays armed at 30 Hz forever, even with zero subscribers, so "idle sessions cost nothing" isn't quite true, idle still costs a 30 Hz wakeup. Could we return False here when there are no subscribers and re-arm on the next subscribe?

A-Mahla (approved):

tfrere:
Heads up on a compat issue I hit while looking at the timer comment, plus the two commits I pushed.

## Old JS SDKs lose robot control against a daemon with this PR

The pose channel was created for every peer, pre-offer. Any SDK released before the label routing added here does this:

```js
this._pc.ondatachannel = (e) => { this._dc = e.channel; ... }
```

Last channel wins. Control is created first, pose second, so `_dc` ends up on the pose channel, which has no `on-message-string` on the daemon side. Every command is then dropped.

The nasty part is that it looks fine from the app: `_sendCommand` only checks `readyState === 'open'`, which the pose channel satisfies, so it returns `true`. `wake_up` just times out and the robot never moves.

I reproduced it on a real robot rather than guessing about DCEP ordering, with a standalone webrtcbin (GStreamer 1.26.2) mirroring `_setup_data_channel` exactly, and Chrome running the `ondatachannel` handler from `main` verbatim:

```
ondatachannel order seen by client: [data, pose]
client._dc.label = pose
client._dc.readyState = open
_sendCommand({type:'wake_up'}) returned = true
robot received on CONTROL: []
robot received on POSE:    ["{\"type\":\"wake_up\"}"]
```

So GStreamer does hand the two channels over in creation order, and the old client sends into the void. This matters because the SDK ships on npm/CDN while the daemon self-updates, and the docs (`javascript-sdk.md`, `APP_CREATION_GUIDE.md`, `skills/create-js-app.md`) tell app authors to pin `@1.8.0`.

## What I pushed

**`8f888ee84`** opens the pose channel on demand instead. `_setup_data_channel` only stores the peer's webrtcbin now, and `set_pose_subscription` opens the channel on first subscribe, hopped onto the GLib main loop (which also serialises the check-then-create). A client that never sends `subscribe_pose` never sees a second channel, so the skew stops mattering in both directions rather than needing a release-ordering rule.

Extra channels ride DCEP in-band, so no renegotiation. I checked that end to end on the robot too: before subscribing the client sees only `data`, and the pose channel shows up right after `subscribe_pose`. The SDK side already copes, it routes by label whenever the channel appears, re-asserts from `_checkSessionReady`, and the 500 ms `requestState()` poll covers the opening gap.

**`b22d7ab9f`** answers @A-Mahla's timer comment. Arming is driven by `subscribe_pose` and `_push_pose` returns `False` once the last subscriber leaves, so an idle session costs no wakeup at all.

Two things came with it. Subscriptions run on the backend's asyncio thread while the timer ticks on the GLib main loop, and a subscribe interleaved with a disarm could read a stale source id, skip the re-arm and leave the stream dead until reconnect, so `_pose_lock` guards the subscriber set, provider and source id (same idea as `_peer_states_lock`). Frame building and sending stay outside it. And disarming depends on the subscriber set emptying, which depended on the channel's `on-close`, which a dead peer may never fire, so `_consumer_removed` drops the peer's pose state too.

I also took the `close()` cleanup from the "Note for reviewers". Guarding only `_pose_push_source_id` moves the unraisable `AttributeError` onto `_bus_sender`, since the realistic `__init__` failure (no camera resolution, the `RuntimeError` before `_build_pipeline`) raises before either exists, so I guarded the whole thing.

Happy to drop the lazy-channel commit if you would rather solve the skew by sequencing the SDK release, it is self-contained. `protocol.py` conflicts with main, but that predates my commits.

A-Mahla (commented):
Two residual nits, neither blocking.

src/reachy_mini/media/media_server.py
```diff
@@ -1739,6 +1786,159 @@ def _setup_data_channel(self, peer_id: str, webrtcbin: Gst.Element) -> None:
...
+        # Unreliable + unordered: pose is a "latest value wins" stream, so we
+        # never want a lost frame to be retransmitted (it'd already be stale)
+        # nor to head-of-line-block the frames behind it.
+        options = Gst.Structure.from_string("options,ordered=false,max-retransmits=0")[
+            0
+        ]
+        channel = webrtcbin.emit("create-data-channel", "pose", options)
+        if not channel:
+            self._logger.error(f"Failed to create pose channel for peer {peer_id}")
+            return
```
A-Mahla: nit: if create-data-channel returns null here (or the webrtcbin was already gone in `_open_pose_channel_if_needed`), the peer stays in `_pose_subscribers` with no channel, so the timer keeps building a state-JSON frame at 30 Hz and sending it to nobody for the rest of the session. Bounded, `_consumer_removed` cleans it up when the peer leaves, but the failure path could also discard the subscriber.

ts/lib/reachy-mini.ts
```diff
@@ -1860,6 +1920,11 @@ export class ReachyMini extends EventTarget implements ReachyMiniInstance {
         if (this._iceConnected && this._dcOpen && this._sessionResolve) {
             this._state = 'streaming';
             this.requestState();
+            // Re-assert a pose subscription that was requested before the data
+            // channel was open, or lost on reconnect (a fresh peer starts
+            // unsubscribed on the daemon). Sent raw so it doesn't touch the
+            // local refcount, which already reflects the live consumer count.
+            if (this._poseSubRefs > 0) this._sendCommand({ type: 'subscribe_pose' });
             this._stateRefreshInterval = setInterval(() => this.requestState(), 500);
```
A-Mahla: nit: this 500 ms get_state polling still runs alongside an active pose stream, and poll replies have no seq so they bypass the high-water-mark guard. A slow poll reply can momentarily rewind the mirror the pose stream is supposed to smooth. Cosmetic, but maybe skip the poll (or the state apply) while `_poseSubRefs > 0`.

src/reachy_mini/media/media_server.py
tfrere: Good catch, fixed in d6d063de6. Both failure paths now discard the subscription: the missing-webrtcbin branch in `_open_pose_channel_if_needed`, and `_setup_pose_channel` which returns a bool so the caller can react to `create-data-channel` handing back null. With the subscriber gone the timer also disarms on its next tick instead of idling armed for the rest of the session. Covered by two tests.

ts/lib/reachy-mini.ts
tfrere: Agreed, fixed in d6d063de6. The poll now stands down while pose frames are arriving.

One deviation worth flagging: I gated it on frame arrival (`_lastPoseFrameAt`, 750 ms window) rather than on `_poseSubRefs > 0`. Against an older daemon the refcount goes up but `subscribe_pose` is rejected and no channel ever opens, so keying on the refcount would leave that client with no state at all instead of falling back to polling, which is the compat path the docstring promises. The timestamp also brings the poll back by itself if the stream stalls, and it is reset on session setup so a leftover value can't hold off the first poll after a reconnect.

tfrere:
tested this end to end on a real Mini, not just unit tests. deployed the branch (merged with main) onto the robot and drove it from a plain page talking straight to the local signalling server on 8443, so real Chrome WebRTC, no SDK in the loop.

8/8:

- no pose channel shows up before `subscribe_pose` is sent, which is the whole old-client story
- channel opens mid-session on subscribe, `ordered=false maxRetransmits=0`
- 30.3 frames/s, seq strictly increasing, 1 -> 121
- control channel still answers `get_state` while the stream runs
- frames stop on unsubscribe
- and they come back on re-subscribe, which is the disarm/re-arm race

daemon logs line up: `Pose stream subscribed by peer` then `Pose channel opened for peer`, and no second `Pose channel opened` on re-subscribe since the existing one gets reused.

heads up, the branch had conflicts with main: the command list in the protocol docstring, and a constant that landed right next to `POSE_STREAM_FRESH_MS` in the SDK. both additive, i merged main in and resolved them as unions, pushed as 228eef6b. the tree of that merge is byte identical to what i deployed and tested, and CI is green on it.

all yours to merge whenever, squash i'd say.

one gotcha if anyone tries to repro this headless: aiortc can't get through the handshake with webrtcsink, session gets ended ~100 ms after the answer with nothing in the logs. tried without STUN and with a silent audio track like the SDK sends, same result. went with a browser instead.

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -1290,6 +1298,59 @@ def get_present_passive_joint_positions(self) -> Optional[Dict[str, float]]:
...
+        of the client round-tripping ``get_state`` over Wi-Fi. Returning
+        ``None`` (e.g. before the first kinematics update, or during
+        shutdown) simply skips that tick.
+        """
+        try:
+            state = self.build_state_dict()
+        except Exception:
+            return None
+        self._pose_seq += 1
+        return json.dumps({"state": state, "seq": self._pose_seq})
```
pierre-rouanet: We should rely on pydantic for json ser/deser as it can have an heavy impact on the CM4 CPU. You can (should) use already existing BaseModel in protocol.
The get_full_state method from routers/state.py already does pretty much what you need. Maybe it can be abstract.

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -1290,6 +1298,59 @@ def get_present_passive_joint_positions(self) -> Optional[Dict[str, float]]:
...
+            "antennas": self.get_present_antenna_joint_positions().tolist()
+            if self.current_antenna_joint_positions is not None
+            else None,
+            # Per-motor joint positions (7 head incl. body yaw at [0], 2
+            # antennas), so WebRTC clients can check the physical pose motor
+            # by motor without the LAN /ws/sdk stream.
+            "head_joint_positions": self.get_present_head_joint_positions().tolist()
+            if self.current_head_joint_positions is not None
+            else None,
+            "antennas_joint_positions": self.get_present_antenna_joint_positions().tolist()
```
pierre-rouanet: You have twice antennas :)

RemiFabre (approved):
LGTM on behavior (tested on my wireless), but +1 on Pierre's double Antenna field.
Note: I tested the pydantic JSON dump and wasn't able to measure a meaningful performance difference (still a good change imo)

RemiFabre:
Note on the CPU cost: I measured +12-22% of a core with this branch compared to main. Seems OK if the full end to end test worked properly

src/reachy_mini/daemon/backend/abstract.py
tfrere: done in d663458. added StateSnapshot / PoseFrame models in protocol.py; build_state_snapshot() is now the single builder and both views derive from it (model_dump for the get_state reply, model_dump_json for the pushed frames).

I didn't reuse get_full_state directly: its keys differ from what the released JS SDKs already parse (control_mode vs motor_mode, antennas_position vs antennas), so reusing it as-is would have broken existing clients. The new model keeps the current wire keys.

src/reachy_mini/daemon/backend/abstract.py
tfrere: good catch :) dropped in d663458. antennas already is the two motor values, so the per-motor twin was pure duplication in every 30 Hz frame. head_joint_positions stays since it's genuinely distinct from the head_pose matrix.

tfrere:
Thanks for measuring! Quick question: was that with a client actively subscribed to the pose stream, or idle? +12-22% of a core during active streaming is the expected cost of the 30 Hz push. If you saw it while idle (no subscriber) that would be a bug, since b22d7ab parks the timer when nobody is subscribed.

Also worth a re-measure on latest: d663458 moved frame serialization to pydantic's model_dump_json and dropped a duplicated field, so frames are a bit cheaper now.

PR #1289, opened 2026-07-22
Fix face tracking on platforms without v4l2convert

alozowski:
Use v4l2convert when available and fall back to portable GStreamer converters otherwise. This preserves Raspberry Pi acceleration and restores face tracking on macOS, Windows, and Linux systems without the V4L2 plugin.

Add a regression test for the fallback path.

Closes https://github.com/pollen-robotics/reachy_mini/issues/1276

PR #1303, opened 2026-07-27
docs(design): conversation service design breakdown

tfrere:
## Summary

Adds a shared **design-discussion base** for the on-robot conversation service,
under `docs/design/conversation-service/`. Each doc is scoped to the **public
interface** of its block (the `conversation.*` / `motion.*` / `personalities.*`
surfaces, the `config` schema, the model-facing tools) plus the few decisions that
frame it - not internal wiring.

The core proposal: turn the conversation from a store app (separate subprocess,
cold start, REST + SSE) into a **first-class in-process service of the daemon**,
kept warm and exposed through a stable `conversation.*` protocol over the existing
WebRTC relay, with the embodiment (`MovementManager`) extracted as a **standalone
`motion.*` capability** any client or JS-SDK app can drive.

## Contents

| Doc | Public interface |
|---|---|
| `README.md` | Overview, the in-process architecture decision, roadmap, open questions |
| `01-conversation-engine.md` | `conversation.*` (RPC + events), the controller FSM, audio I/O |
| `02-motion-service.md` | `motion.*` - standalone embodiment over WebRTC |
| `03-vision.md` | Camera fan-out + face/head tracking - documents the **existing** daemon capability we reuse |
| `04-tools.md` | Model-facing verbs + execution contract |
| `05-memory.md` | `save_memory` / `recall_memory`, logging + prompt injection |
| `06-personas-and-config.md` | `personalities.*` / `voices.*` + the `config` schema they resolve into |
| `07-lifecycle-and-supervision.md` | Warm supervision, robot lock, wake/sleep, auth, crash surfacing, inactivity |
| `08-testing.md` | Contract/conformance tests + the WebRTC harness |

## Notes

- Docs only, no code changes. Status is "draft / design proposal" (except vision,
  which documents the already-implemented daemon surface).
- Meant as a common base to discuss and refine the breakdown before implementation.

### DIFF
--- a/docs/design/conversation-service/01-conversation-engine.md
+++ b/docs/design/conversation-service/01-conversation-engine.md
@@ -1,0 +1,95 @@
+# 01 - Conversation engine
+
+Status: draft / design proposal.
+
+The engine is the in-process component that runs one speech-to-speech (S2S)
+session at a time. It owns the cloud realtime connection, its audio I/O, and its
+public `conversation.*` control surface. It is glue: it dispatches tool calls and
+drives the body through `motion.*`, but never touches motors directly.
+
+This doc covers the engine, the protocol it exposes, and its audio path - the
+three faces of the same block.
+
+## Public interface
+
+### `conversation.*` (JSON-RPC 2.0 over the WebRTC DataChannel)
+
+Rides the existing relay (#1266); no bespoke transport. Requests are dispatched
+by direct in-process calls to the controller.
+
+| method | params | blocking | resolves with |
+|---|---|---|---|
+| `conversation.start` | `config` | yes, until `running` or `start_timeout` | phase snapshot |
+| `conversation.stop` | `session_id?` | no | phase snapshot (`idle`) |
+| `conversation.restart` | `config`, `session_id?` | yes | phase snapshot (fresh `session_id`) |
+| `conversation.status` | - | no | phase snapshot |
+| `conversation.say` | `text`, `session_id?` | no (on accept) | result |
+| `conversation.interrupt` | `session_id?` | yes, until playback stops | result |
+
+### Events (one-way notifications, broadcast to every connected transport)
+
+- `conversation.phase` - `{ session_id, phase, reason, origin{by}, config }`. Source of truth, also returned by `status`.
+- `conversation.turn` - `{ state: listening|thinking|speaking, reason? }`.
+- `conversation.transcript` - `{ role, text, final }`.
+- `conversation.level` - `{ rms }` (~10 Hz).
+
+### Error `reason` codes (stable, UI branches on them)
+
+`robot_busy`, `missing_credentials`, `already_running`, `not_running`,
+`start_timeout`.
+
+### Consumes
+
+- `motion.*` ([`02-motion-service.md`](./02-motion-service.md)) for embodiment.
+- Tools ([`04-tools.md`](./04-tools.md)) on model tool-calls.
+- The daemon's mic/speaker hardware for audio I/O.
+
+## Lifecycle (controller FSM)
+
+```
+idle -> starting -> running -> stopping -> idle
+                 \-> error (transient) -> idle
+```
+
+- `start` resolves config, takes the robot lock, wakes the robot if asleep, opens
+  the S2S session, goes `running`.
+- `stop` tears down, sleeps the robot, releases the lock. Idempotent.
+- `restart` = stop + start (v1 may be non-atomic).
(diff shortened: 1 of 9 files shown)


tfrere:
## Why in-daemon: report and comparison

Some context and numbers to back the core proposal of these docs (conversation as an in-process daemon service instead of a store app). Sources inline. Checked against the app's current `main` (648092b4).

### 1. Cold start: the app model has a structural floor

@alozowski measured startup on the wireless in #1282 (7 samples per case, with the conversation app from reachy_mini_conversation_app#506):

- app started directly, robot already awake: settings + event stream ready went from **9.772 s to 7.641 s**
- robot woken and app auto-started by the daemon: **11.975 s to 10.089 s**

That is the *optimized* number. The remaining ~7.6-10 s is structural to the store-app model: spawn a subprocess in the shared apps venv, import the SDK there (warm import alone was ~3.9 s on the CM4 before the import diet in #1308), connect back to the daemon over localhost, start a per-app settings web server, then open the cloud session. And as #1308 itself notes, its gain does not reach power-on-to-talking for the conversation app, because the app actually uses the lazily imported bits. #1282 / #1308 / #1309 are all good work, but they are shaving a boundary the default experience does not need.

In the in-process design the daemon imports everything once at boot and keeps the runtime dormant (07-lifecycle). `conversation.start` only has to open the S2S session, a couple of seconds masked by the starting animation. The cold-start category disappears instead of getting shaved.

### 2. Installation: the default experience currently depends on the most fragile machinery we have

Today the flagship conversation experience is gated by app install: the apps venv, uv resolve/install from an HF Space, update checker, app/SDK version compatibility. That machinery is where a disproportionate share of open field-bug fixes sit right now: install hanging on "finalizing" (#1299), uv vs git-lfs (#799), app name mismatches across venvs (#778), orphaned app processes on daemon restart (#818).

Shipping the conversation in the daemon removes install, update and version-skew from the out-of-box path entirely: it versions with the daemon, is tested against the exact SDK it runs with, and updates through the existing daemon update flow. The store machinery stays, for third-party apps, where it belongs.

### 3. Quick comparison: current app vs the proposed services

To be fair to the app first: it already unified its control surface on JSON-RPC (no REST/SSE left beyond static files), it already rides the daemon's WebRTC relay (#1266) for remote clients, and it no longer does network discovery of MCP tools at startup (bundled specs + cached manifest). The remaining differences are the structural ones:

| | store app (today) | in-daemon services (these docs) |
|---|---|---|
| process model | subprocess per session in the apps venv, cold start | supervised asyncio task, warm code / dormant runtime |
| transport | JSON-RPC over WS, but served twice: on `:7860` (unauthenticated, LAN) and via the daemon relay | relay only; the unauthenticated `:7860` surface and the per-app web server disappear |
| concurrency | threads + asyncio patchwork (`run_coroutine_threadsafe`, cross-thread events, polling loops in `console.py`) | single event loop, supervisor owns failures, bounded queues |
| embodiment | `MovementManager` buried in the app; robot is inert without an app running | standalone `motion.*` capability any client or JS-SDK app can drive; "alive" is a robot property |
| config/state | mutable global config, `os.environ` as an internal bus, five instance files (`.env`, `startup_settings.json`, `profile_toolsets.json`, `installed_tool_spaces.json`, `memory.v1.json`) | stateless protocol, full `config` on every `start`, effective config echoed on `phase` |
| tools | dynamic Python file loading per profile + per-profile toolset overrides | fixed built-in registry in v1, MCP as a later extension behind the same `config.tools` shape |
| failure surfacing | app subprocess dies; daemon monitors exit and releases the lock | `conversation.error` + `phase: error`, settle to `idle`, retry is just another `start` |

The good parts of the app are ported, not rewritten: the serialized `response.create` sender, barge-in flush, `BackgroundToolManager`, `PersonalityOps` (already transport-agnostic).

### 4. Trade-offs we accept (and mitigations)

- **Release coupling**: conversation fixes ship with daemon releases instead of an independent store update. Mitigated by the fact that the app already has to track SDK releases closely anyway (version-skew bugs above are the current cost of pretending otherwise).
- **Blast radius**: an engine bug now lives in the daemon process. The supervisor + memory watchdog + bounded queues cover the Python failure modes; the only native dep in scope (GStreamer) is already hosted by the daemon today.

### 5. Precedent

Face tracking followed the same path: it moved from app-side into the daemon as an in-process worker, a niced detector thread (per-thread nice 19, YuNet on ONNX Runtime) introduced with the vision subpackage in #1177 and refined since (#1268, #1289). Same escape hatch here: anything that loads a heavy local model gets its own process, the conversation engine does not.

PR #1307, opened 2026-07-30
Pr/gpio script direct

RemiFabre:
Two important changes:

1)
The power-button monitor launched as `python -m reachy_mini...`, importing the entire SDK (onnxruntime included) in parallel with daemon boot (for a gpiozero-only script). It now execs the script file directly (same code
): **~1s boot time reduction, RAM usage (RSS) 150→25MB. **

2) 
 Nothing ever synced the gpio systemd unit after install because fleet units point at a stale Nov-2025 path, so the first fix would never reach robots. A compare-then-copy check (same pattern as the existing bluetooth/launcher self-heals) now runs every boot; restart only on diff.

Note: point 2) is not hypothetical (verified on a production robot): the unit was generated at image-build time by `install_service.sh`, with ExecStart pointing at the provisioning source checkout `/venvs/src/reachy_mini` — frozen at 3186327 (2025-11-19) — while every later update only replaces site-packages.

Measured consequence (see the reproduction in the test comment below): changes to the monitor *module* do still ship even with the stale unit, because the old launcher's `python -m` resolves `reachy_mini` from site-packages (verified with `importlib.util.find_spec` on-robot). What NEVER ships is the launcher and the unit file themselves — the service's entry path. Concretely: **commit 1 of this very PR does nothing on a fleet robot without commit 2** (we shipped the new launcher to site-packages in the stale state, restarted the service, and it kept executing the frozen copy). It also means every fleet unit silently depends on `/venvs/src` existing forever.


RemiFabre:
### How to test on a robot

*(Every step below was executed end-to-end on a wireless unit on 2026-07-30,
including a full reproduction of the stale fleet unit via the frozen
provisioning installer; expected outputs are real measurements.)*

**1. Baseline (robot on main).** SSH in and look at the service:

    PID=$(systemctl show -p MainPID --value gpio-shutdown-daemon)
    ps -o pid,rss,etime,args -p $PID --ppid $PID

Expected: TWO processes — a bash launcher (~3 MB) whose child is the real
monitor at ~125 MB RAM (RSS):

        PID   RSS     ETIME  COMMAND
       1198  3188     ...    /bin/bash .../gpio_shutdown/launcher.sh
       1206  124480   ...    python -m reachy_mini.daemon...shutdown_monitor

The `-m` invocation imports the whole SDK for a gpiozero-only script.

**1b (optional, 5 min) — reproduce the delivery failure commit 2 fixes.**
Show that in the pre-PR state, shipped launcher changes never reach the
service. First check whether your robot is already in the fleet state:

    systemctl cat gpio-shutdown-daemon | grep ExecStart

`/venvs/src/...` → you're in the fleet state natively, skip the next block.
A site-packages path → your unit was already fixed; recreate the original
unit exactly the way the image did, by running the FROZEN provisioning
installer:

    sudo bash /venvs/src/reachy_mini/src/reachy_mini/daemon/app/services/gpio_shutdown/install_service.sh
    sudo systemctl restart gpio-shutdown-daemon

Now "ship" this PR's launcher fix the way pip does — write it into
site-packages — and watch it change nothing:

    SP=/venvs/mini_daemon/lib/python3.12/site-packages/reachy_mini/daemon/app/services/gpio_shutdown
    sudo cp $SP/launcher.sh /tmp/launcher.bak
    curl -s https://raw.githubusercontent.com/pollen-robotics/reachy_mini/pr/gpio-script-direct/src/reachy_mini/daemon/app/services/gpio_shutdown/launcher.sh | sudo tee $SP/launcher.sh >/dev/null
    sudo systemctl restart gpio-shutdown-daemon && sleep 2
    PID=$(systemctl show -p MainPID --value gpio-shutdown-daemon)
    ps -o pid,rss,args -p $PID --ppid $PID

Result (measured): STILL two processes, still executing
`/venvs/src/.../launcher.sh` — the fix sitting in site-packages is never
run, and a reboot wouldn't help (the unit pins the path). That's the
fleet-wide no-op commit 2 prevents.

Honest precision, also measured: monitor *module* changes DO ship even in
this state — the old launcher's `python -m` resolves `reachy_mini` from
site-packages (`importlib.util.find_spec` → site-packages path), so past
monitor fixes like the Dec-2025 debounce did reach robots. The stale unit
pins only the launcher + unit file — i.e. exactly what commit 1 changes.

Clean up before continuing: `sudo cp /tmp/launcher.bak $SP/launcher.sh` —
then proceed to step 2, whose install heals the unit for real.

**2. Install this branch.** The install runs as a background job and
**finishes by restarting the daemon**, which wipes the job registry — so
"job not found" + the new ref installed IS the success signal:

    JOB=$(curl -s -X POST "http://localhost:8000/update/start-from-ref?git_ref=pr/gpio-script-direct" | sed 's/.*"job_id":"\([^"]*\)".*/\1/')
    until curl -s "http://localhost:8000/update/info?job_id=$JOB" | grep -q "not found" \
       && curl -s http://localhost:8000/update/install-source | grep -q '"git_ref":"pr/gpio-script-direct"'; do sleep 5; echo -n .; done; echo " installed"

~40 s if the robot has fetched this ref before, a few minutes otherwise.
Note: "installed" is quiet — nothing visibly changes on the robot until the
reboot. The daemon restarts in the background; the gpio service keeps its
OLD process until the unit is (re)started (the self-heal only restarts it
when the unit file actually differs).

If it's stuck >5 min, `curl -s "http://localhost:8000/update/info?job_id=$JOB"`
shows status + pip logs ("failed" = the install didn't apply).

**3. Self-heal proof — check BEFORE rebooting.** The startup checks run when
the daemon restarts at the end of the install, so on a robot with the stale
image-time unit the heal happens right there:

    journalctl -b | grep -i "gpio-shutdown service"

Expected on a fleet robot: `gpio-shutdown service has changed, updating...`
then `Successfully updated gpio-shutdown service`. Every subsequent daemon
start/boot logs `gpio-shutdown service is up to date`. (If your robot's unit
happens to be current already, you only get "up to date" — to see the heal
in action anyway: point ExecStart at a bogus path, `daemon-reload`, restart
the daemon, and watch it get repaired; verified working.)

**4. New process shape.** Reboot (`sudo reboot`), then same command as step 1:

    PID=$(systemctl show -p MainPID --value gpio-shutdown-daemon)
    ps -o pid,rss,etime,args -p $PID --ppid $PID
    systemctl cat gpio-shutdown-daemon | grep ExecStart

Expected: ONE process — the launcher now `exec`s python, so MainPID IS the
monitor, at ~25 MB RAM (measured 24.8 MB), and ExecStart points into
site-packages (the self-healed unit):

        PID   RSS    ETIME  COMMAND
       1197  24776   ...    python .../gpio_shutdown/shutdown_monitor.py

**5. Button behavior — prove the shutdown code executes** (do this last,
it powers the robot off). Background: the power button also cuts power at
the hardware level after a short grace window (capacitors), so "the robot
turned off" proves nothing — a dead monitor still ends with a power cut,
just a DIRTY one. The proof must come from the previous boot's journal,
and these images ship an rpi drop-in forcing volatile journals, so first
make the journal persistent:

    sudo mkdir -p /etc/systemd/journald.conf.d
    printf '[Journal]\nStorage=persistent\n' | sudo tee /etc/systemd/journald.conf.d/99-boot-test-persistent.conf
    sudo systemctl restart systemd-journald && sudo journalctl --flush

Now press and release the power button once (nothing should happen while
you hold it; shutdown starts ~200 ms after release — the 200 ms loop is a
debounce that filters the contact bounces of one physical press, one press
= one trigger). Power the robot back on, then read the evidence:

    journalctl -b -1 -u gpio-shutdown-daemon | tail -2
    journalctl -b -1 | tail -3

Pass criteria (measured on a real unit):
- The first output must contain `COMMAND=/usr/sbin/shutdown -h now` issued
  by the unit = the monitor's shutdown code EXECUTED. (Don't look for the
  script's "Shutdown button released..." print — python block-buffers
  stdout to journald and never flushes, so that line is lost on every
  robot; the sudo COMMAND line is the reliable marker.)
- `The system will power off now!` in the second = shutdown was accepted
  and started.
- Note: the journal will END right there, without the `Journal stopped`
  tail — the hardware cuts power ~1 s into the shutdown, before the OS
  finishes. That happens identically on main (the monitor code is
  unchanged); it's a pre-existing hardware/firmware characteristic, not a
  regression from this PR.

Revert the journal config afterwards (it's an SD-wear tradeoff, not meant
to stay on):

    sudo rm /etc/systemd/journald.conf.d/99-boot-test-persistent.conf
    sudo rm -rf /var/log/journal/*
    sudo systemctl restart systemd-journald

**6. Restore the robot** (same completion signal as step 2):

    curl -s -X POST "http://localhost:8000/update/start-from-ref?git_ref=main"
    # wait for install-source to show "git_ref":"main", then:
    sudo reboot

After this the service is back to the two-process/-m shape of step 1.

---

Unit tests (NOT on the robot — on your dev machine, from a checkout of
this branch; CI runs them too):

    git fetch origin pr/gpio-script-direct && git checkout pr/gpio-script-direct
    uv run pytest tests/unit_tests/test_startup_check_gpio_shutdown.py

(6 tests: copy-on-diff, install-if-missing, noop-if-same, tolerate
systemctl failure, packaged unit sanity.)

src/reachy_mini/daemon/app/services/gpio_shutdown/launcher.sh
```diff
@@ -1,3 +1,8 @@
 #!/bin/bash
 source /venvs/mini_daemon/bin/activate
-python -m reachy_mini.daemon.app.services.gpio_shutdown.shutdown_monitor
+
+# Run the monitor file directly instead of `python -m reachy_mini...`:
+# module execution imports the whole reachy_mini SDK (several seconds of CPU
+# and I/O during boot, tens of MB of RSS) while the script itself only needs
+# gpiozero and the stdlib.
+exec python "$(dirname "$(readlink -f "$0")")/shutdown_monitor.py"
```
RemiFabre, line 8: This is change number 1)

src/reachy_mini/utils/wireless_version/startup_check.py
```diff
@@ -252,6 +255,82 @@ def check_and_update_wireless_launcher() -> None:
...
+    Compares the packaged gpio-shutdown-daemon.service with the installed
+    version at /etc/systemd/system. Older images generated the unit at install
+    time with an ExecStart pointing at wherever the SDK lived back then, so
+    without this sync a stale unit keeps running old launcher code forever.
+    On a difference the packaged unit is copied over and the service is
+    restarted; a missing unit is installed and enabled.
+    """
+    source = (
+        Path(__file__).parent
+        / ".."
```
pierre-rouanet, line 270: mix of parent and ".." is a bit odd.

src/reachy_mini/utils/wireless_version/startup_check.py
```diff
@@ -252,6 +255,82 @@ def check_and_update_wireless_launcher() -> None:
...
+        / "app"
+        / "services"
+        / "gpio_shutdown"
+        / "gpio-shutdown-daemon.service"
+    )
+    source = source.resolve()
+    target = GPIO_SHUTDOWN_UNIT_TARGET
+
+    if not source.exists():
+        print(f"Source gpio-shutdown service file not found at {source}")
```
pierre-rouanet, line 282: Use log instead of print.

pierre-rouanet (approved):
Nice! Changes seem ok to me and make sense.
But I don't have access to a robot to really test it.

andimarafioti (changes requested):
- **[P2] Partially applied unit updates are never retried** — [[startup_check.py L285–329](https://github.com/pollen-robotics/reachy_mini/blob/be0a0e52cf95c475949f7b1edefb9877fe40b6a7/src/reachy_mini/utils/wireless_version/startup_check.py#L285-L329)](https://github.com/pollen-robotics/reachy_mini/blob/be0a0e52cf95c475949f7b1edefb9877fe40b6a7/src/reachy_mini/utils/wireless_version/startup_check.py#L285-L329). If the copy succeeds but `daemon-reload`, restart, or enable fails, the exception is swallowed while the new target file remains. Every later daemon startup then sees equal files and returns early, leaving PID 1 using the old cached unit—or a missing service disabled—until another intervention or reboot. Restore the old target on failure, keep a retry marker, or check systemd state such as `NeedDaemonReload`/enabled/active before returning. Add a test that fails the second or third subprocess call and invokes the check again.

RemiFabre:
> * **[P2] Partially applied unit updates are never retried** — [[startup_check.py L285–329](https://github.com/pollen-robotics/reachy_mini/blob/be0a0e52cf95c475949f7b1edefb9877fe40b6a7/src/reachy_mini/utils/wireless_version/startup_check.py#L285-L329)](https://github.com/pollen-robotics/reachy_mini/blob/be0a0e52cf95c475949f7b1edefb9877fe40b6a7/src/reachy_mini/utils/wireless_version/startup_check.py#L285-L329
>       ). If the copy succeeds but `daemon-reload`, restart, or enable fails, the exception is swallowed while the new target file remains. Every later daemon startup then sees equal files and returns early, leaving PID 1 using the old cached unit—or a missing service disabled—until another intervention or reboot. Restore the old target on failure, keep a retry marker, or check systemd state such as `NeedDaemonReload`/enabled/active before returning. Add a test that fails the second or third subprocess call and invokes the check again.

Good catch!

andimarafioti (approved):
Great work! Thank you for addressing my comment!

PR #1308, opened 2026-07-30
perf(sdk): import diet — lazy WSServer/fastapi, import 3.5 s → 1.8 s on the robot

RemiFabre:
## Issue

No linked issue (predates the one-PR-one-issue rule). The problem: `import reachy_mini` pays for the whole daemon backend on the user-visible path of every app start — `io` eagerly pulls `WSServer` → `daemon.backend.abstract` → `vision.face_tracking` → onnxruntime, and `apps.app` eagerly pulls fastapi. Neither is needed by a client that just drives the robot. Reproduce with `python -X importtime -c "import reachy_mini"` on the robot.

## Description

Two deferrals (PEP 562 / late import), public API unchanged:

1. `reachy_mini.io` exports `WSServer` lazily — `from reachy_mini.io import WSServer` still works, but the daemon-backend chain (incl. onnxruntime) stays out of client processes until actually used.
2. fastapi is imported when a settings app is built, not at `import reachy_mini`.

This branch now includes #1343 (merged in): the original third commit here (lazy scipy) is superseded by it — the numpy `utils.rotation` module is essentially free to import, so those four call sites are eager again, and `recorded_move.py`'s scipy pre-warm guard is gone with scipy itself. **Merge #1343 first**; this PR's diff shrinks to the two deferrals once it lands.

Trade-off: the deferred cost moves to first use — the daemon pays ~2.0 s at first `WSServer` access instead of every client app paying it at import.

## Testing

- Measured on the wireless robot (CM4, 2026-08-12), this branch vs #1343 alone vs `main`, warm = 3 runs, cold = `drop_caches` before each of 3 runs — table in the comments below. Headline: `import reachy_mini` **4.43 s → 1.80 s warm, 5.75 s → 2.60 s cold** combined with #1343 (−59% / −55%); this PR's share is −1.67 s warm / −1.97 s cold on top of #1343.
- `onnxruntime`, `fastapi` and `scipy` all confirmed absent from `sys.modules` after `import reachy_mini` on the robot.
- With scipy blocked from importing entirely (simulating a fresh runtime install): `import reachy_mini`, `create_head_pose` and `linear_pose_interpolation` all pass.
- Daemon restarted on this exact build on the robot: state `running`, WSServer lazy path exercised by the daemon itself.

## Tested on

- [x] Reachy Mini Wireless
- [ ] Reachy Mini Lite

## AI assistance

Assisted-by: Claude:claude-opus-5
Assisted-by: Claude:claude-fable-5


pierre-rouanet (commented):
Code looks fine but I'm really not sure it changes much to the metrics _time spent before the robot starts moving_ as those import are still required. They will simply be done a few steps later.

src/reachy_mini/utils/interpolation.py
```diff
@@ -66,6 +65,12 @@ def linear_pose_interpolation(
     Use `yaw_as_scalar` to interpolate yaw as a signed scalar Euler angle instead of along the SO(3) geodesic.
     This keeps the path through the front rather than taking the shortest 3D rotation through +-180° (the back).
     """
+    # Imported here so that `import reachy_mini` doesn't pay for scipy (~1s on
+    # the wireless robot). Callers that evaluate this in a control loop must
+    # import scipy at module scope to avoid a first-call stall (see
+    # recorded_move.py and daemon/backend/abstract.py).
+    from scipy.spatial.transform import Rotation as R
```
FabienDanieau: I'm not a big fan of the lazy imports. Besides any goto / fk is going to use it right? So basically this import is mandatory.
Now it seems that scipy is only used for the Rotation. Why don't we remove the scipy deps and make our own small lib instead?

RemiFabre:
> This does NOT improve the "power-on → talking" time (yet) because the conv app uses all the lazy imports (but this will change once we merge another change on the conv app side because it allows to move compute time away from the critical path). This is also a win for anything that doesn't use the lazily imported libs.

Yes this is said above:
> This does NOT improve the "power-on → talking" time (yet) because the conv app uses all the lazy imports (but this will change once we merge another change on the conv app side because it allows to move compute time away from the critical path). This is also a win for anything that doesn't use the lazily imported libs.

The change that makes this faster is this:
https://github.com/pollen-robotics/reachy_mini_conversation_app/pull/513/changes

src/reachy_mini/utils/interpolation.py
RemiFabre: Let me measure the cost of this specific import

RemiFabre: Ok so good call! Measured on the CM4, this single import is huge:
image

And we import a BUNCH of stuff we don't use:
image
Like KD-tree, sparse matrices, numpy.testing, and crackfortran, a Fortran source parser...

Of course my Claude was like "dw fam I recoded the 8 functions you need in these 150 lines (using numpy, already imported) and tested against scipy rotation and it works!":
image

Of course it also goes on to say that it's faster in most cases:
image

I was very suspicious of this, how can you beat a cython implementation that much? Turns out we pay mostly for fixed costs like object creation. The scipy implementation is much faster when batching calculations (which we don't need in our code base):
image

RemiFabre: So this is very cool. I'm down to do this, but in a different PR and with a lot of tests.
Ok @FabienDanieau @pierre-rouanet ?

pierre-rouanet: yes it makes sense. It also makes me sad. But I'm ok with this in this particular case.

RemiFabre:
## Measurements (2026-08-12, after merging #1343 in)

Wireless robot (CM4), `/venvs/mini_daemon` python, same venv across rows. Warm = 3 consecutive runs; cold = `sync; echo 3 > drop_caches` before each of 3 runs.

### `import reachy_mini`

| build | warm | cold |
|---|---|---|
| `main` (pre-#1343, ref numbers) | 4.43 s | 5.75 s |
| #1343 alone (`feat/drop-scipy-rotation`) | 3.47 s | 4.57 s |
| **this branch (#1343 + import diet, `1c04e67c8`)** | **1.80 s** (1.95 / 1.79 / 1.80) | **2.60 s** (2.62 / 2.59 / 2.60) |

This PR's marginal contribution on top of #1343: **−1.67 s warm / −1.97 s cold**.

### Where that comes from (importtime subtrees inside `import reachy_mini`, on #1343 alone)

| deferred subtree | warm | cold |
|---|---|---|
| fastapi (via `apps.app`) | 1.11 s | 1.48 s |
| `io.ws_server` chain (`backend.abstract` → `vision.face_tracking` → `face_detector` → onnxruntime) | ~1.07 s | ~1.2 s |

Note onnxruntime's *own* subtree is only ~0.11 s — the value of the WSServer deferral is the whole vision/backend chain above it.

### Deferred price

`import reachy_mini; from reachy_mini.io import WSServer` warm: 3.85 s, i.e. the process that really needs the backend (the daemon) pays ~2.0 s at first access. Client apps never do.

### Sanity

`[m for m in ("onnxruntime","fastapi","scipy") if m in sys.modules]` after `import reachy_mini` on the robot: `[]`.

RemiFabre:
So the minimal rotation PR is almost merged (we're aiming at the next RC, not this one). I merged the new rotation lib in this PR and measured the gain of the remaining lazy imports: −1.67 s warm / −1.97 s cold.
This gains are only visible when we start an app as soon as the daemon is ready (e.g. the conv app). I'd be pro merging this, but no strong opinions.

FabienDanieau (approved):
All good!

PR #1309, opened 2026-07-30
perf(wireless): skip unchanged startup checks with a boot stamp (-4.7 s)

RemiFabre:
**Stacked on #1307: merge that first; this PR then shows only the stamp commits.**

Every boot, the wireless robot re-runs maintenance on its two Python environments ("venvs": `/venvs/mini_daemon` for the daemon, `/venvs/apps_venv` for apps): walk every file under `/venvs` to fix ownership (~4 s), ask pip whether the backup venv under `/restore` is intact (~1.5 s), and compare the apps venv's SDK against the daemon's. The answers only change when something is installed, updated, or repaired. Yet every boot paid for them.

The fix: after a fully successful check run, write a small JSON file, the **stamp** (`/venvs/mini_daemon/.startup_check_stamp.json`), recording what the checks saw: for each installed `reachy_mini`, the signature of its RECORD file (pip's own package inventory, a text file listing every installed file, rewritten by pip/uv on every install, upgrade or uninstall), where the SDK came from (PyPI vs a git ref), and a hash of the check code itself. Next boot, if everything matches, skip the expensive checks. If ANYTHING is off (file missing, unreadable, mismatched, an editable install, an update, a factory reset), throw the stamp away and run the full checks. Cheap checks (bluetooth files, systemd-unit self-heals) still run every boot: they guard files outside `/venvs` that a venv-keyed stamp cannot see. Measured gain: **4.7 s less on every boot** (the biggest single win of the boot-time work after network ordering).

**Worst case if the stamp is wrongly valid?** The robot skips a repair it actually needed (for example a manually corrupted apps venv stays broken until the next install or update clears the stamp). The keying is designed so that no pip/uv operation can leave a wrongly-valid stamp (they all rewrite RECORD files), and `update_reachy_mini` deletes the stamp *before* touching any venv as a second layer. The known blind spot is manual mutation over SSH (sudo-created files, scp into site-packages): accepted and documented, bounded (any install/update restores the full checks), and the most critical paths are still ownership-probed cheaply on every fast boot.

**Lite robots:** out of scope, this improvement is wireless-only by construction. The only call site runs inside `if args.wireless_version:` in `daemon/app/main.py`, and both update entry points that clear the stamp are wireless-gated too. Proof details, a 10-scenario backfire review, and an adversarial campaign on a real robot are in the test comment below.

14 unit tests pin the contract: any doubt (missing/corrupt/stale stamp, failed check, editable install) falls back to the full checks, and a stamp is only written after every expensive check succeeded.


RemiFabre:
### How to test on a robot

*(Every step below was executed end-to-end on a wireless unit on 2026-07-31; expected outputs are real.)*

**1. Install this branch.** The install runs as a background job and **finishes by restarting the daemon**, which wipes the job registry, so "job not found" + the new ref installed IS the success signal:

    JOB=$(curl -s -X POST "http://localhost:8000/update/start-from-ref?git_ref=pr/startup-check-stamp" | sed 's/.*"job_id":"\([^"]*\)".*/\1/')
    until curl -s "http://localhost:8000/update/info?job_id=$JOB" | grep -q "not found" \
       && curl -s http://localhost:8000/update/install-source | grep -q '"git_ref":"pr/startup-check-stamp"'; do sleep 5; echo -n .; done; echo " installed"

The daemon restart at the end of the install already runs the startup checks. Look at them: first run has no stamp, so the FULL checks run, then the stamp is written:

    journalctl -u reachy-mini-daemon | grep -i stamp

Expected:

    Startup stamp not usable ([Errno 2] No such file or directory: '/venvs/mini_daemon/.startup_check_stamp.json'); running full startup checks
    Startup stamp written to /venvs/mini_daemon/.startup_check_stamp.json

**2. See the fast path.** Reboot (`sudo reboot`), then:

    journalctl -b | grep -i stamp

Expected:

    Startup stamp valid, skipping expensive startup checks

The stamp itself is human-readable: `cat /venvs/mini_daemon/.startup_check_stamp.json | python3 -m json.tool`. You'll see the RECORD signatures (path:mtime_ns:size) for the daemon, apps and restore venvs, the install source, and the hash of the check code.

**3. See any install invalidate it.** Run any SDK install (e.g. re-run step 1, or `/update/start`), reboot: the journal shows the full checks again, then a fresh stamp. App installs from the dashboard do NOT invalidate it (they don't touch what the expensive checks repair; see the campaign table below for why that's correct and verified).

**4. See corruption fall back safely.** `echo garbage | sudo tee /venvs/mini_daemon/.startup_check_stamp.json`, reboot: journal shows `Startup stamp not usable (...); running full startup checks`, robot boots healthy, stamp rewritten.

**5. Restore.** `curl -X POST "http://localhost:8000/update/start-from-ref?git_ref=main"` (same completion signal as step 1).

**Dev machine:** `uv run pytest tests/unit_tests/test_startup_check_stamp.py -q`: 14 tests pin the contract (any doubt means full checks; stamp only written after full success; update clears the stamp before touching any venv; stale stamps say why in the logs).

---

### Backfire review: "when could the stamp be VALID while the venvs are actually broken or stale?"

Code read end-to-end on commit d11d599c (+ test commit 205ea064). Verdicts:

| # | Scenario | Verdict | Evidence |
|---|----------|---------|----------|
| 1 | Interrupted install (pip/uv killed mid-update) | Fail-safe, doubly covered | `update_reachy_mini` deletes the stamp before its first pip command (`update.py:22-31`; ordering pinned by `test_update_clears_stamp_before_first_install`). Independently, any completed dist-info rewrite changes RECORD (path, mtime_ns, size) so a half-finished install can't reproduce the recorded signature. Robot-verified: pkill -9 mid-install scenario D3. |
| 2 | Check code changes via update, stamp still valid | Covered | `checks_hash` = sha256 of `startup_check.py` itself. Helper modules it imports (`update_available`, `utils`) can only change through a (re)install (editable installs are refused in `_compute_stamp_signature`), and every install rewrites RECORD → stamp invalid. Manual scp of a single file into site-packages is the documented blind spot (see #3). |
| 3 | Ownership drift outside installs (sudo-created root files under /venvs) | Accepted blind spot, bounded | Stamp stays valid → full chown scan skipped. Mitigations: `_quick_ownership_ok` probes /venvs, both venv roots, daemon python, and the stamp itself on every fast boot; any install/update clears the stamp and restores the full scan. Deep root-owned files persist until then. Robot-verified in scenario D5. |
| 4 | App installs/uninstalls mutate apps_venv daily | Safe by keying | App packages don't touch what the expensive checks repair. The one case that matters, dependency resolution up/down-grading `reachy_mini` inside apps_venv, rewrites its RECORD, which is part of the stamp (`apps_record`), so the next boot runs the full checks incl. SDK sync. Robot-verified in scenario D1. |
| 5 | SOFTWARE_RESET / factory restore | Verified on robot | `/bluetooth/commands/SOFTWARE_RESET.sh` starts with `rm -rf /venvs/` then copies `/restore/venvs/` (no stamp inside) → stamp destroyed, next boot full checks. |
| 6 | Downgrade to a pre-stamp SDK, upgrade back | Safe | Both installs rewrite RECORD and both run `update_reachy_mini`, which pre-clears the stamp. Format enforcement: validity is whole-dict equality against a freshly computed signature that always contains `"format": STAMP_FORMAT`, so any schema drift mismatches. |
| 7 | Disk full / read-only fs when writing the stamp | Fail-safe | `_write_startup_stamp` failures are caught in `run_wireless_startup_checks` (never fatal), followed by `clear_startup_stamp` (itself try/except OSError). Boot continues; next boot has no valid stamp → full checks. A partial `.tmp` file can be left on ENOSPC: never read, overwritten by the next successful write. |
| 8 | Corrupt stamp (truncated JSON, garbage, wrong types) | Fail-safe, tested | JSON parse errors → exception → full checks (`test_corrupt_stamp_falls_back_to_full_checks`). Valid-JSON-wrong-shape fails dict equality → full checks. Stamp path squatted by a directory → full checks, no crash (`test_stamp_path_being_a_directory_never_crashes_boot`). Robot-verified in scenario D4. |
| 9 | Clock weirdness (RTC-less Pi, time jumps) | No wall-clock dependence | Validity is equality of recorded stat values (mtime_ns, size, absolute path), with no expiry, no "newer than" comparison. A clock jump alone can't make a mismatched stamp match; it would additionally require a reinstall that reproduces the exact nanosecond mtime AND a skipped pre-clear, which no supported flow can produce. |

Extra surface found while reading (beyond the planned list):

| # | Scenario | Verdict | Evidence |
|---|----------|---------|----------|
| 10 | App install racing the boot-time checks | Impossible via API | `run_wireless_startup_checks` runs synchronously in `main()` before `run_app()` starts the HTTP server, so no API exists yet to install anything. Only manual SSH pip during the ~11 s check window could race, which is the manual-mutation blind spot of #3. |

---

### Adversarial campaign on a real robot (2026-07-31)

Goal: behave like a slightly unlucky, slightly creative user and try to end up with a broken or stale robot. After every scenario: reboot, check which path ran (journal), verify health (daemon running, app install + start/stop works, `current-app-status` flows).

| # | Scenario | What was done | Observed | Verdict |
|---|----------|---------------|----------|---------|
| D0 | Baseline | Installed branch via `start-from-ref`; watched the post-install daemon restart, then rebooted | Post-install restart: `Startup stamp not usable (No such file...); running full startup checks` then `Startup stamp written`. Reboot: `Startup stamp valid, skipping expensive startup checks`. Daemon running, backend ready | PASS |
| D1 | 20x app install/uninstall cycles (the smartphone-app flow) | 20 remove+install cycles of `wake_me_up` via `/api/apps/remove` + `/api/apps/install`, then reboot | Stamp byte-identical after all 20 cycles (same md5); the apps venv's reachy_mini RECORD mtime untouched (apps install beside it, never rewrite it); reboot fast-paths; app then starts (`running`) and stops (`null`) cleanly. Not invalidating is correct: app installs change nothing the expensive checks repair, and if an app install ever up/down-grades reachy_mini itself, that rewrites RECORD and invalidates the stamp | PASS |
| D2 | SDK downgrade to pre-stamp `main`, reboot, upgrade back, plus a stale-stamp replant | `start-from-ref?git_ref=main`, reboot, `start-from-ref?git_ref=pr/startup-check-stamp`; then hand-edited the fresh stamp's recorded RECORD mtime (minus 1000 ns) to fake a stamp from another install epoch, rebooted | Downgrade: stamp deleted before install (file absent after), `main` boots with the old always-full checks, healthy. Upgrade back: full checks + new stamp. Replanted stale stamp: rejected, full checks ran, fresh stamp written, healthy. Follow-up reboot fast-paths | PASS |
| D3 | pip killed mid-install (software half of a power cut) | `start-from-ref` (force-reinstall), then `sudo pkill -9 -f "pip install"` 8 s in (confirmed the daemon-venv `uv pip install` was alive and was killed) | Stamp already deleted at job start, so no boot could ever trust pre-kill state. The update flow logged the failure and moved on (pre-existing behavior: phase errors are streamed, not raised); the post-job daemon restart ran FULL checks, verified both venvs consistent (uv leaves old-or-new state, not shreds), wrote a fresh stamp, daemon healthy. A retried update then succeeded normally with full checks + fresh stamp after | PASS |
| D4 | Corrupt the stamp by hand | 100 random bytes over the stamp, reboot | `Startup stamp not usable ('utf-8' codec can't decode byte 0xa9...); running full startup checks`, healthy boot, stamp rewritten | PASS |
| D5 | Ownership drift: root-owned file under /venvs (sudo touch) | `sudo touch /venvs/mini_daemon/rooty_file`, reboot; then cleared the stamp (what any install/update does) and restarted | Reboot: fast path taken, file stays root-owned, robot fully healthy. This is the documented blind spot, observed as predicted. Bound demonstrated: after the stamp clear, full checks printed `Found file not owned by pollen: /venvs/mini_daemon/rooty_file` + `Fixing ownership`, and the file became pollen-owned | PASS (blind spot bounded as documented) |
| D6 | Disk full around the stamp write | Skipped on this shared test unit: filling the root filesystem risks corrupting unrelated state (journald, app data) on a robot used daily. The path is unit-covered: a write failure is caught, the stamp is cleared, the boot continues, and the next boot runs the full checks | n/a | SKIPPED (justified) |
| D7 | App install racing the boot-time checks | Removed the stamp, rebooted, probed the API every ~3 s from SSH during boot while journald recorded precise timings | API returned connection-refused for the entire check window; journal ordering: `running full startup checks` 11:28:50 → `stamp written` 11:28:56 → uvicorn `Application startup complete` 11:28:58. The HTTP server only starts after the checks return, so no dashboard/app-store operation can race them, by construction rather than by luck | PASS (race impossible via API) |

Campaign ran at commit 205ea064; commit 0c2cdadb (adds the `Startup stamp mismatch` journal line found missing during D2) was then installed and verified on-robot: planted planted stale stamp produced `Startup stamp mismatch (install state changed); running full startup checks`, then a fresh stamp, then a fast path on the next reboot.

RemiFabre:
Pinging @cdussieux because this could be of interest to you

src/reachy_mini/utils/wireless_version/update.py
```diff
@@ -19,6 +19,17 @@ async def update_reachy_mini(
...
+    # Invalidate the startup-check stamp before touching any venv, so a crash
+    # mid-update always triggers the full startup checks on the next boot.
+    # (The stamp's dist-info signature would catch this anyway; this makes it
+    # double-covered.)
+    try:
+        from .startup_check import clear_startup_stamp
+
+        clear_startup_stamp()
+    except Exception:
+        logger.warning("Could not clear startup-check stamp before update")
```
pierre-rouanet, line 31: Not sure about this only be a log. If clear_statup_stamp fails we probably have a broken setup anyway. What do you think?

src/reachy_mini/utils/wireless_version/startup_check.py
```diff
@@ -31,17 +54,17 @@ def check_and_fix_venvs_ownership(
         pollen_uid = pwd.getpwnam(USER).pw_uid
     except KeyError:
         print(f"User '{USER}' does not exist on this system")
```
pierre-rouanet, line 56: Not on your diff, but we really should log instead of print in all this file. Clearly some info that we want to be able to access later.

src/reachy_mini/utils/wireless_version/startup_check.py
```diff
@@ -252,12 +279,91 @@ def check_and_update_wireless_launcher() -> None:
...
+            print("gpio-shutdown service has changed, updating...")
+        except Exception as e:
+            print(f"Error comparing gpio-shutdown service files: {e}")
+            return
+    else:
+        print(f"gpio-shutdown service not installed at {target}, installing...")
+
+    try:
+        print(f"Copying {source} to {target}")
+        subprocess.run(
```
pierre-rouanet, line 324: Using shutil instead?

src/reachy_mini/utils/wireless_version/update.py
RemiFabre: What do you mean by "only be a log"? You mean if this happens, the message should be more visible and contain more info? e.g. like this:
logger.error(
            "Could not clear the startup-check stamp at %s. Continuing: the "
            "install signature will invalidate it on the next boot anyway. "
            "A stamp that cannot be deleted points at a filesystem or "
            "ownership problem worth investigating.",
            STAMP_PATH,
            exc_info=True,
        )

I agree that if clear_startup_stamp fails (pretty much what it does is trying to delete a file) then we have some serious problems like a read-only remount, a corrupted SD card or ownership changes. But not sure what we can do on those scenarios.

Note: the way all of this is designed is that if anything goes wrong, the stamp is cleared and the full startup checks are done.

src/reachy_mini/utils/wireless_version/startup_check.py
RemiFabre: Agreed, I'll do a separate PR for this to keep things clean

src/reachy_mini/utils/wireless_version/startup_check.py
RemiFabre: I don't think we can because we need to reach /etc/systemd/system/gpio-shutdown-daemon.service which is owned by root no?

Also the "sudo cp" pattern is used in other places too (check_and_update_bluetooth_service and check_and_update_wireless_launcher). So if we rework this we should do it everywhere

src/reachy_mini/utils/wireless_version/update.py
pierre-rouanet: No I meant not sure we should catch the error, but maybe it should really raise.

src/reachy_mini/utils/wireless_version/startup_check.py
pierre-rouanet: ah yes ok, my bad.

src/reachy_mini/utils/wireless_version/update.py
RemiFabre: Hum, I think if we let it crash here we lose on a (tiny) chance that the daemon restart fixes the problem (or somehow manages to run with a wacky version). I think for a dev robot I would make it crash, but for a robot managed by non technical users I'd rather increase the chance of it booting.

No strong opinions tho and this seems like a very unlikely scenario anyways

FabienDanieau (approved):
Oki, tested on a wireless

PR #1304, opened 2026-07-29
Host/SDK/daemon integration branch - to be split (do not review)

tfrere:
Work-in-progress integration branch for the host / SDK / daemon work. **The split is done** - review happens on the PRs below, not here. This PR stays open only as the integration reference until every piece lands, then it will be closed.

## Split status

Each item below renders with its live state chip (green = open, purple = merged, red = closed) - no manual bookkeeping, this list is always current.

**Daemon** - independent, land in any order:

- [x] https://github.com/pollen-robotics/reachy_mini/pull/1310
- [ ] https://github.com/pollen-robotics/reachy_mini/pull/1311
- [ ] https://github.com/pollen-robotics/reachy_mini/pull/1312
- [x] https://github.com/pollen-robotics/reachy_mini/pull/1313

**TypeScript - SDK (`ts/lib`)**:

- [ ] https://github.com/pollen-robotics/reachy_mini/pull/1315

**TypeScript - Host (`ts/host`)** - stacked, in order:

- [ ] https://github.com/pollen-robotics/reachy_mini/pull/1316
- [ ] https://github.com/pollen-robotics/reachy_mini/pull/1321

(#1321 is a follow-up from the #1316 review: the SDK staleness check pulled out of #1316 and rebuilt on the right version source - stacked on #1316.)

**Already landed / superseded:**

- [x] https://github.com/pollen-robotics/reachy_mini/pull/1296
- #1314 was the former single JS block, closed and superseded by #1315 + #1316.

## Notes

Each split branch was rebuilt from `main` by cherry-picks and verified content-identical to this branch on its owned paths (the mixed commits were split by path). Unit tests were re-run per block; #1315 additionally builds and passes the suite with the CURRENT host code, proving its additive-API claim.

Merge ordering: the 4 daemon PRs and #1315 are independent and land in any order; #1316 lands after #1315 (its base), and #1321 after #1316. The TS side is fail-open against daemons that lack the matching commands (unknown commands are silently dropped and the SDK resolves `null` on a shared 4 s timeout). Behavioural pairings, noted in each PR:

- **Sleep-on-leave** (#1316) pairs with the **handoff grace** (#1311): without it the daemon may replay its own goto_sleep after a handoff.
- **Update gate** (#1316) pairs with **exit-code propagation** (#1312): it's what makes the gate's failure path honest instead of looping on a silently-failed install.
- **`getImu()`** (#1315) pairs with #1313; resolves `null` against older daemons.
[...]

PR #1310, opened 2026-08-03
feat(daemon): stream Direction-of-Arrival in the robot state

tfrere:

The ReSpeaker DoA reading is added to the state snapshot, so it rides the pushed pose stream and `get_state` replies instead of a REST round-trip.

- `get_DoA()` is a blocking USB control transfer, so it never runs on the ~30 Hz pose-push path: a poller caches it at ~10 Hz and `build_state_snapshot()` reads the cache under a lock.
- The poller is demand-driven: it spawns lazily on the first cache demand, exits after 5 s without consumers, and never runs without a device - zero USB traffic while nobody wants DoA.
- REST reads (`/state/doa`, `/state/full?with_doa=true`) go through the shared cache + USB lock, so a one-shot request still gets a real reading when the poller is cold without ever interleaving with the poller's USB conversation.
- Typed as `DoaSnapshot` on the pydantic `StateSnapshot` (`doa: null` when no ReSpeaker is present or no reading has landed yet).
- `Backend.close()` stops the poller and releases the ReSpeaker USB handle.

Wire compat: `doa` is a new optional field on the state payload; released SDKs ignore it. The JS-side parsing lands in the JS block PR (fail-open: shows n/a against older daemons).

Tests: `test_backend_doa.py` (11 tests, incl. close() teardown) + `test_pose_stream.py` still green (17 passed locally). Validated on hardware (wireless robot, real ReSpeaker).

### DIFF
--- a/src/reachy_mini/daemon/app/routers/state.py
+++ b/src/reachy_mini/daemon/app/routers/state.py
@@ -65,10 +65,11 @@ async def get_doa(
 
     Returns the angle in radians (0=left, π/2=front, π=right) and speech detection status.
     Returns None if the audio device is not available.
+
+    Reads through the backend's DoA cache (`read_doa`) so it can never
+    interleave with the background poller's USB conversation.
     """
-    if not backend.doa:
-        return None
-    result = backend.doa.get_DoA()
+    result = backend.read_doa()
     if result is None:
         return None
     return DoAInfo(angle=result[0], speech_detected=result[1])
@@ -121,8 +122,8 @@ async def get_full_state(
             result["passive_joints"] = list(joints.values())
         else:
             result["passive_joints"] = None
-    if with_doa and backend.doa:
-        doa_result = backend.doa.get_DoA()
+    if with_doa:
+        doa_result = backend.read_doa()
         if doa_result:
             result["doa"] = DoAInfo(angle=doa_result[0], speech_detected=doa_result[1])
 
--- a/src/reachy_mini/daemon/app/routers/state.py
+++ b/src/reachy_mini/daemon/app/routers/state.py
@@ -64,15 +64,10 @@ async def get_doa(
     """Get the Direction of Arrival from the microphone array.
 
     Returns the angle in radians (0=left, π/2=front, π=right) and speech detection status.
-    Returns None if the audio device is not available.
-
-    Reads through the backend's DoA cache (`read_doa`) so it can never
-    interleave with the background poller's USB conversation.
+    Returns None if the audio device is not available or no reading has
+    landed yet (the first one lands within ~100 ms of the first request).
     """
-    result = backend.read_doa()
-    if result is None:
-        return None
-    return DoAInfo(angle=result[0], speech_detected=result[1])
+    return backend.read_doa()
 
 
 @router.get("/full")
@@ -123,9 +118,7 @@ async def get_full_state(
         else:
             result["passive_joints"] = None
     if with_doa:
-        doa_result = backend.read_doa()
-        if doa_result:
-            result["doa"] = DoAInfo(angle=doa_result[0], speech_detected=doa_result[1])
+        result["doa"] = backend.read_doa()
 
     result["timestamp"] = datetime.now(timezone.utc)
     return FullState.model_validate(result)
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -33,6 +33,7 @@ from reachy_mini.io.protocol import (
     CancelMoveCmd,
     ClearIncomingAudioCmd,
     DeleteHfTokenCmd,
+    DoaSnapshot,
     FaceTarget,
     GetHardwareIdCmd,
     GetMicrophoneVolumeCmd,
@@ -147,19 +148,29 @@ class Backend:
         self.use_audio = use_audio
 
         self.doa = AudioDoA() if use_audio else None
-        # Latest Direction-of-Arrival reading, refreshed by a lightweight
+        # Latest Direction-of-Arrival reading, refreshed by a demand-driven
         # background poller. `get_DoA()` is a blocking USB control transfer, so
         # we must NOT call it from `build_state_dict()` - that feeds the ~30 Hz
         # pose push (`_push_pose`) on the media server's GLib loop, which has to
-        # stay cheap and non-blocking. Instead this thread samples DoA at
-        # ~10 Hz (matching the ReSpeaker's useful rate) and caches it here; the
-        # state snapshot just reads the cache under a lock.
+        # stay cheap and non-blocking. Instead a thread samples DoA at ~10 Hz
+        # and caches it here; the state snapshot just reads the cache under
+        # `_doa_lock`. The thread is spawned lazily on first demand (a pushed
+        # frame or a REST read) and exits after `DOA_IDLE_STOP_S` without any,
+        # so an idle daemon doesn't poll USB forever (see `_note_doa_demand`).
         self._doa_lock = threading.Lock()
+        # Serializes every `get_DoA()` call: `ReSpeaker.read()` is a
+        # multi-step USB conversation on a non-thread-safe pyusb device, so
+        # the poller and an on-demand REST read must never interleave.
+        self._doa_usb_lock = threading.Lock()
         self._last_doa: tuple[float, bool] | None = None
+        # Monotonic time of the last completed read (poller or direct).
+        # 0.0 = never; `_doa_snapshot` treats entries older than
+        # `DOA_CACHE_FRESH_S` as absent.
+        self._last_doa_ts: float = 0.0
+        # Monotonic time of the last cache consumer; keeps the poller alive.
+        self._doa_demand_ts: float = 0.0
         self._doa_stop = threading.Event()
         self._doa_thread: threading.Thread | None = None
-        if self.doa is not None:
-            self._start_doa_poller()
 
         self.should_stop = threading.Event()
         self.ready = threading.Event()
@@ -1331,43 +1342,105 @@ class Backend:
     # ~10 Hz: fast enough to track a speaker turning, slow enough to keep the
     # blocking ReSpeaker USB reads off the ~30 Hz pose-push loop.
     DOA_POLL_INTERVAL_S = 0.1
+    # Cache entries older than this are treated as absent: covers both a
+    # poller that just stopped (idle) and a burst of failed USB reads.
+    DOA_CACHE_FRESH_S = 0.5
+    # The poller exits after this long without any cache consumer, so DoA
+    # costs zero USB traffic while no session or REST client wants it.
+    DOA_IDLE_STOP_S = 5.0
 
-    def _start_doa_poller(self) -> None:
-        """Spawn the background thread that caches the latest DoA reading.
+    def _doa_available(self) -> bool:
+        """Return True when a ReSpeaker was actually found at startup."""
+        return self.doa is not None and self.doa.available
 
-        Runs until :meth:`close` sets ``_doa_stop``. Every failure path is
-        swallowed so a flaky USB read can never take the thread (or the state
-        stream that reads its cache) down.
+    def _note_doa_demand(self) -> None:
+        """Record cache demand and make sure the poller is running.
+
+        Called from every consumer path (`_doa_snapshot`, `read_doa`).
+        Spawning is guarded by ``_doa_lock`` so the ~30 Hz push loop and
+        concurrent REST requests can't start two threads.
         """
+        now = time.monotonic()
+        with self._doa_lock:
+            self._doa_demand_ts = now
+            if self._doa_thread is not None and self._doa_thread.is_alive():
+                return
+            self._doa_thread = threading.Thread(
+                target=self._doa_poll_loop, name="doa-poller", daemon=True
+            )
+            self._doa_thread.start()
 
-        def _loop() -> None:
-            # `wait()` returns True once stopped, False on timeout - so this
-            # both paces the poll and exits promptly on shutdown.
-            while not self._doa_stop.wait(self.DOA_POLL_INTERVAL_S):
-                try:
+    def _doa_poll_loop(self) -> None:
+        """Cache the latest DoA reading until shutdown or idle timeout.
+
+        Every failure path is swallowed so a flaky USB read can never take
+        the thread (or the state stream that reads its cache) down.
+        """
+        # `wait()` returns True once stopped, False on timeout - so this
+        # both paces the poll and exits promptly on shutdown.
+        while not self._doa_stop.wait(self.DOA_POLL_INTERVAL_S):
+            with self._doa_lock:
+                idle_for = time.monotonic() - self._doa_demand_ts
+            if idle_for > self.DOA_IDLE_STOP_S:
+                return
+            try:
+                with self._doa_usb_lock:
                     reading = self.doa.get_DoA() if self.doa is not None else None
-                except Exception:  # noqa: BLE001 - never kill the poller
-                    reading = None
-                with self._doa_lock:
-                    self._last_doa = reading
+            except Exception:  # noqa: BLE001 - never kill the poller
+                reading = None
+            with self._doa_lock:
+                self._last_doa = reading
+                self._last_doa_ts = time.monotonic()
 
-        self._doa_thread = threading.Thread(
-            target=_loop, name="doa-poller", daemon=True
-        )
-        self._doa_thread.start()
+    def _cached_doa(self) -> tuple[float, bool] | None:
+        """Return the cached reading if fresh, else None."""
+        with self._doa_lock:
+            if time.monotonic() - self._last_doa_ts > self.DOA_CACHE_FRESH_S:
+                return None
+            return self._last_doa
 
-    def _doa_snapshot(self) -> dict[str, Any] | None:
-        """Return the last cached DoA as ``{"angle", "speech_detected"}``.
+    def _doa_snapshot(self) -> DoaSnapshot | None:
+        """Return the last cached DoA reading.
 
-        ``None`` when no ReSpeaker is present or no reading has landed yet.
+        Never blocks: this feeds the ~30 Hz pose push. Registers demand so
+        the poller (re)starts, and returns ``None`` when no ReSpeaker is
+        present, the cache is stale, or the first poll hasn't landed yet.
         Matches the REST ``DoAInfo`` field names so both surfaces agree.
         """
-        with self._doa_lock:
-            reading = self._last_doa
+        if not self._doa_available():
+            return None
+        self._note_doa_demand()
+        reading = self._cached_doa()
         if reading is None:
             return None
         angle, speech_detected = reading
-        return {"angle": angle, "speech_detected": speech_detected}
+        return DoaSnapshot(angle=angle, speech_detected=speech_detected)
+
+    def read_doa(self) -> tuple[float, bool] | None:
+        """Return a current DoA reading, blocking briefly if needed.
+
+        The REST surface (`/state/doa`, `/state/full?with_doa=true`) goes
+        through here instead of calling ``doa.get_DoA()`` directly, so a
+        one-shot request still gets a real reading when the poller is cold
+        (matching the historical behaviour) without ever interleaving with
+        the poller's own USB conversation (``_doa_usb_lock``).
+        """
+        if not self._doa_available():
+            return None
+        self._note_doa_demand()
+        cached = self._cached_doa()
+        if cached is not None:
+            return cached
+        try:
+            with self._doa_usb_lock:
+                reading = self.doa.get_DoA() if self.doa is not None else None
+        except Exception:  # noqa: BLE001 - a failed read reads as "no DoA"
+            self.logger.debug("Direct DoA read failed", exc_info=True)
+            reading = None
+        with self._doa_lock:
+            self._last_doa = reading
+            self._last_doa_ts = time.monotonic()
+        return reading
 
     # ------------------------------------------------------------------
     # State snapshot (shared by get_state replies and the pose push)
@@ -1402,7 +1475,7 @@ class Backend:
             face_target=self.get_tracked_face(),
             # Sound Direction of Arrival (ReSpeaker mic array), or None when
             # unavailable. Read from the ~10 Hz cache so this stays cheap on
-            # the pose-push path (see _start_doa_poller).
+            # the pose-push path (see _note_doa_demand / _doa_poll_loop).
             doa=self._doa_snapshot(),
         )
 
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -147,6 +147,19 @@ class Backend:
         self.use_audio = use_audio
 
         self.doa = AudioDoA() if use_audio else None
+        # Latest Direction-of-Arrival reading, refreshed by a lightweight
+        # background poller. `get_DoA()` is a blocking USB control transfer, so
+        # we must NOT call it from `build_state_dict()` - that feeds the ~30 Hz
+        # pose push (`_push_pose`) on the media server's GLib loop, which has to
+        # stay cheap and non-blocking. Instead this thread samples DoA at
+        # ~10 Hz (matching the ReSpeaker's useful rate) and caches it here; the
+        # state snapshot just reads the cache under a lock.
+        self._doa_lock = threading.Lock()
+        self._last_doa: tuple[float, bool] | None = None
+        self._doa_stop = threading.Event()
+        self._doa_thread: threading.Thread | None = None
+        if self.doa is not None:
+            self._start_doa_poller()
 
         self.should_stop = threading.Event()
         self.ready = threading.Event()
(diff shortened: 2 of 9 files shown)


src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -431,6 +455,17 @@ def close(self) -> None:
...
+        self._doa_stop.set()
+        if self._doa_thread is not None:
+            self._doa_thread.join(timeout=1.0)
+            self._doa_thread = None
+        if self.doa is not None:
+            try:
+                self.doa.close()
+            except Exception:  # noqa: BLE001 - best-effort teardown
+                self.logger.debug("DoA close failed", exc_info=True)
+            self.doa = None
```
FabienDanieau: `close()` can dispose the USB device while the poller is still inside `ctrl_transfer`.

`_doa_stop.set()` can't interrupt a transfer already in flight, and `join(timeout=1.0)` gives up after 1 s. A `ReSpeaker.read()` in the retry handshake takes **≥1 s by construction** (`audio_control_utils.py`: up to 100 attempts, 10 ms sleep each), and a single wedged transfer can block up to 100 s (`TIMEOUT = 100000`).

Past the join timeout we fall through to `self.doa.close()` → `usb.util.dispose_resources(self.dev)` on a device another thread is actively transferring on — a libusb use-after-free: segfault, or a spurious `USBError` out of the poller.

Reachable in practice: `wrapped_run()` calls `close()` from the backend thread on any `run()` exception, with the poller live.

Taking the USB lock makes the teardown provably safe:

```suggestion
        # Stop the DoA poller and release the ReSpeaker USB handle.
        self._doa_stop.set()
        if self._doa_thread is not None:
            self._doa_thread.join(timeout=1.0)
            self._doa_thread = None
        # `join(timeout=...)` can return with a wedged USB read still in flight,
        # so only the lock proves the poller isn't mid-`ctrl_transfer` here.
        with self._doa_usb_lock:
            if self.doa is not None:
                try:
                    self.doa.close()
                except Exception:  # noqa: BLE001 - best-effort teardown
                    self.logger.debug("DoA close failed", exc_info=True)
                self.doa = None
```

Also worth a test with a fake whose `get_DoA()` sleeps ~0.3 s: the current fakes all return instantly, so the blocking tail — the reason the poller thread exists — is never exercised.

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -147,6 +148,29 @@ def __init__(
...
+        # stay cheap and non-blocking. Instead a thread samples DoA at ~10 Hz
+        # and caches it here; the state snapshot just reads the cache under
+        # `_doa_lock`. The thread is spawned lazily on first demand (a pushed
+        # frame or a REST read) and exits after `DOA_IDLE_STOP_S` without any,
+        # so an idle daemon doesn't poll USB forever (see `_note_doa_demand`).
+        self._doa_lock = threading.Lock()
+        # Serializes every `get_DoA()` call: `ReSpeaker.read()` is a
+        # multi-step USB conversation on a non-thread-safe pyusb device, so
+        # the poller and an on-demand REST read must never interleave.
+        self._doa_usb_lock = threading.Lock()
```
FabienDanieau: This lock only serializes *this* handle — three other paths talk to the same physical ReSpeaker.

- The `ApplyAudioConfigCmd` / `ReadAudioParameterCmd` branch further down this file opens its own `init_respeaker_usb()` handle, runs a multi-transfer conversation (`apply_audio_config(verify=True)` is many read+write round-trips), then `dispose_resources` — none of it under this lock.
- `audio_base.py:125` gives every `MediaManager` / `GStreamerAudio` its own `AudioDoA`, read through `MediaManager.get_DoA()`. That's every SDK app process — conversation app included — running on the robot next to the daemon. A `threading.Lock` doesn't span processes.

Before this PR the daemon touched the ReSpeaker only when someone explicitly asked, so the interleaving window was small and user-initiated. At a continuous 10 Hz it becomes the default case: `_doa_snapshot()` sits on the `build_state_snapshot()` path, so *any* connected client reading state keeps the poller alive whether or not it cares about `doa`.

The asymmetry is the part that worries me. The poller swallows the resulting `ValueError` (`doa` just flickers to `null`), but `AudioDoA.get_DoA()` has no try/except — so `read()` raising `Unknown status code` / `Read attempt exceeds 100 times` propagates straight into app code. The daemon's background polling can now break an app's `get_DoA()`, or the audio-settings panel's `read_audio_parameter`.

Options, laziest first:

1. Route every in-daemon ReSpeaker access through `self.doa` + this lock. Kills the transient handle in the `ApplyAudioConfigCmd` branch and is a strict improvement regardless of the rest.
2. Poll only while something actually consumes DoA, rather than while any client reads state — that shrinks the collision window back to roughly what it was before.
3. If cross-process contention turns out to be real on hardware, arbitration has to move into the device layer (single owner + an accessor), which is a much bigger change — worth knowing now whether it's needed.

Have you tried the conversation app (which does its own `get_DoA()`) against a daemon with the poller running? That's the combination this would show up in.

src/reachy_mini/daemon/app/routers/state.py
```diff
@@ -65,10 +65,11 @@ async def get_doa(
...
     Returns the angle in radians (0=left, π/2=front, π=right) and speech detection status.
     Returns None if the audio device is not available.
+
+    Reads through the backend's DoA cache (`read_doa`) so it can never
+    interleave with the background poller's USB conversation.
     """
-    if not backend.doa:
-        return None
-    result = backend.doa.get_DoA()
+    result = backend.read_doa()
```
FabienDanieau: `get_doa` is `async def`, so this blocking USB read runs *on the event loop*.

On a cold cache `read_doa()` falls through to `get_DoA()` under `_doa_usb_lock`, so this can wait out the poller's in-flight read plus its own — seconds in the retry handshake, up to 100 s on a wedged transfer (`TIMEOUT = 100000`). That doesn't slow one request, it freezes the daemon's whole asyncio loop: every HTTP route, WebRTC signalling, the lot.

The shape is pre-existing (the old code called `get_DoA()` from the same `async def`), but the PR makes it more likely by adding lock contention on top of it.

Cheapest fix is to drop `async` from the two signatures — FastAPI then runs them in the threadpool, where blocking is fine:

- L61 `async def get_doa(` → `def get_doa(`
- L79 `async def get_full_state(` → `def get_full_state(` (it calls `read_doa()` too)

Nothing else changes; neither body awaits anything.

Alternatively, serve the cache only and never block here — see my note on `read_doa`.

<sub>(edited: the original had a `suggestion` block that was a no-op — the signature lines aren't in this PR's diff, so the change can't be anchored as an applyable suggestion.)</sub>

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -1300,6 +1335,113 @@ def get_present_passive_joint_positions(self) -> Optional[Dict[str, float]]:
...
+
+    # ~10 Hz: fast enough to track a speaker turning, slow enough to keep the
+    # blocking ReSpeaker USB reads off the ~30 Hz pose-push loop.
+    DOA_POLL_INTERVAL_S = 0.1
+    # Cache entries older than this are treated as absent: covers both a
+    # poller that just stopped (idle) and a burst of failed USB reads.
+    DOA_CACHE_FRESH_S = 0.5
+    # The poller exits after this long without any cache consumer, so DoA
+    # costs zero USB traffic while no session or REST client wants it.
+    DOA_IDLE_STOP_S = 5.0
```
FabienDanieau: This doesn't gate on what the comment says it gates on.

`_doa_snapshot()` is called unconditionally from `build_state_snapshot()`, which backs *both* the ~30 Hz pose push and every `get_state` reply. So the demand heartbeat is refreshed by any client reading state at all — desktop app, mobile app, an SDK `get_state()` loop — whether or not anything ever looks at `doa`. In practice `DOA_IDLE_STOP_S` means "no client connected", not "no session or REST client wants it".

That matters because of the multi-handle contention I flagged on `_doa_usb_lock`: the 10 Hz USB stream ends up on by default in normal use, which is the state the gating was meant to avoid.

Two honest ways out:

- **Gate on real consumption** — a DoA subscription flag, or `with_doa`-style opt-in — so the heartbeat means what the comment claims.
- **Drop the gating.** If "zero USB traffic while idle" isn't a measured requirement, starting the poller in `__init__` when `_doa_available()` (what the first revision of this PR did) removes `_note_doa_demand`, `_doa_demand_ts`, `DOA_IDLE_STOP_S`, the idle check in the loop, the restart path, and 2 tests — ~40 lines. It also makes the poller always warm, which in turn makes `read_doa`'s direct-read fallback dead code.

I'd rather have the first if the contention is real and the second if it isn't. What it currently is — gating that fires only when nobody's connected — is the cost of both with the benefit of neither.

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -1300,6 +1335,113 @@ def get_present_passive_joint_positions(self) -> Optional[Dict[str, float]]:
...
+            if idle_for > self.DOA_IDLE_STOP_S:
+                return
+            try:
+                with self._doa_usb_lock:
+                    reading = self.doa.get_DoA() if self.doa is not None else None
+            except Exception:  # noqa: BLE001 - never kill the poller
+                reading = None
+            with self._doa_lock:
+                self._last_doa = reading
+                self._last_doa_ts = time.monotonic()
```
FabienDanieau: Two things in this loop, both fixable in one edit.

**The first reading is always one interval late.** `wait()` runs *before* the first read, so `doa` is `null` for the first ~100 ms after demand — about 3 pose frames.

**A retrying read expires its own cache.** `DOA_CACHE_FRESH_S = 0.5`, but `ReSpeaker.read()`'s retry handshake sleeps 10 ms per attempt for up to 100 attempts, so a struggling-but-working device takes 0.6–1 s per read. The cache then goes stale *while a valid reading is in flight* and `doa` flips to `null` and back — exactly when a consumer doing "turn toward the sound" least wants gaps.

Reading first and pacing afterwards fixes the latency; clearing the cache on idle exit removes the need for a freshness window at all (the only stale case is a poller that stopped and restarted — nothing else can leave a stale entry, since a failed read already writes `None`):

```suggestion
        while True:
            with self._doa_lock:
                idle_for = time.monotonic() - self._doa_demand_ts
            if idle_for > self.DOA_IDLE_STOP_S:
                with self._doa_lock:
                    self._last_doa = None  # never serve a pre-idle reading
                return
            try:
                with self._doa_usb_lock:
                    reading = self.doa.get_DoA() if self.doa is not None else None
            except Exception:  # noqa: BLE001 - never kill the poller
                reading = None
            with self._doa_lock:
                self._last_doa = reading
            # `wait()` returns True once stopped: paces the poll, exits promptly.
            if self._doa_stop.wait(self.DOA_POLL_INTERVAL_S):
                return
```

That lets `DOA_CACHE_FRESH_S`, `_last_doa_ts` and the freshness check in `_cached_doa()` go — three fields/constants and ~12 lines, plus `test_snapshot_ignores_stale_cache`.

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -1300,6 +1335,113 @@ def get_present_passive_joint_positions(self) -> Optional[Dict[str, float]]:
...
+        try:
+            with self._doa_usb_lock:
+                reading = self.doa.get_DoA() if self.doa is not None else None
+        except Exception:  # noqa: BLE001 - a failed read reads as "no DoA"
+            self.logger.debug("Direct DoA read failed", exc_info=True)
+            reading = None
+        with self._doa_lock:
+            self._last_doa = reading
+            self._last_doa_ts = time.monotonic()
+        return reading
```
FabienDanieau: This fallback is the one that can park the asyncio loop for 100 s (both callers are `async def`), and it buys very little: the poller lands a reading within one `DOA_POLL_INTERVAL_S` of the same `_note_doa_demand()` call that just ran two lines up. So the window where this branch does anything is ~100 ms after the *first* demand ever — and if the poller stops being demand-gated it never opens at all.

```suggestion
        return self._cached_doa()
```

13 lines → 1, and it takes `test_read_doa_cold_cache_reads_directly`, `test_read_doa_serves_fresh_cache_without_usb` and `test_read_doa_swallows_usb_errors` with it (~30 test lines).

Honest trade-off: this *is* a behaviour change — a `/state/doa` request arriving before the first poll now gets `null` instead of blocking for a reading, which is the "matching the historical behaviour" property you called out in the description. My read is that a `null` for 100 ms is a much better failure mode than a 100 s event-loop stall, but it's your call. If you want to keep the blocking read, then #3 (dropping `async` from the endpoints) becomes mandatory rather than nice-to-have.

src/reachy_mini/io/protocol.py
```diff
@@ -97,6 +97,13 @@ class FaceTarget(BaseModel):
     ts: float | None = None
 
 
+class DoaSnapshot(BaseModel):
+    """Sound Direction of Arrival reading (ReSpeaker mic array)."""
+
+    angle: float
+    speech_detected: bool
```
FabienDanieau, line 107: This is field-for-field identical to `DoAInfo` in `daemon/app/models.py`, and `models.py` already imports from `reachy_mini.io.protocol` — so the duplicate isn't buying any layering.

Keep this one, and in `models.py` replace the class with an alias:

```python
from reachy_mini.io.protocol import DoaSnapshot, MotorControlMode

DoAInfo = DoaSnapshot  # kept for the REST surface's name
```

Then have `read_doa()` return `DoaSnapshot | None` instead of `tuple[float, bool] | None` and both router sites collapse:

```python
return backend.read_doa()
...
result["doa"] = backend.read_doa()
```

That drops the tuple→model unpacking in two places, the `DoAInfo` definition, and the "Matches the REST `DoAInfo` field names so both surfaces agree" caveat in `_doa_snapshot`'s docstring — the two models can't drift if there's one model. Public schema is unchanged, so `openapi.json` stays as regenerated.

tests/unit_tests/test_backend_doa.py
```diff
@@ -0,0 +1,211 @@
...
+        self.available = available
+        self.raise_on_read = raise_on_read
+        self.calls = 0
+
+    def get_DoA(self) -> tuple[float, bool] | None:  # noqa: N802 - mirrors AudioDoA
+        """Return the configured reading, counting every call."""
+        self.calls += 1
+        if self.raise_on_read:
+            raise RuntimeError("usb read failed")
+        return self.reading
```
FabienDanieau: Every fake read returns instantly, so the blocking tail — the entire reason this poller exists on its own thread — is never exercised. The lifecycle is well covered; the failure modes aren't.

One knob covers most of the gap:

```suggestion
    def get_DoA(self) -> tuple[float, bool] | None:  # noqa: N802 - mirrors AudioDoA
        """Return the configured reading, counting every call."""
        self.calls += 1
        time.sleep(self.read_delay)
        if self.raise_on_read:
            raise RuntimeError("usb read failed")
        return self.reading
```

(plus `read_delay: float = 0.0` in `__init__`). With `read_delay=0.3` you can then cover:

- **`close()` while a read is in flight** — the use-after-free I flagged on the teardown block: assert the device isn't disposed until the transfer returns.
- **cache expiry mid-read** — with `read_delay` above `DOA_CACHE_FRESH_S`, `_doa_snapshot()` returns `None` on a device that's working fine.
- **`read_doa()` contending with the poller** on `_doa_usb_lock`.

tests/unit_tests/test_backend_doa.py
```diff
@@ -0,0 +1,211 @@
...
+
+def test_build_state_dict_carries_doa_and_no_dead_payload() -> None:
+    """The pushed frame has `doa` and no duplicated antenna payload."""
+    backend = _make_backend(doa=None)
+    state = backend.build_state_dict()
+    assert "doa" in state
+    # Per-motor head values are a real feature of the pose stream...
+    assert "head_joint_positions" in state
+    # ...but the antenna twin duplicated `antennas` verbatim and was dropped.
+    assert "antennas_joint_positions" not in state
```
FabienDanieau: These two assertions are about the antenna-twin removal from another PR, not about DoA — the test name promises `no_dead_payload` but this PR doesn't touch that payload. Line 207 (`assert "doa" in state`) is the part that belongs here.

```suggestion
```

FabienDanieau:
## Summary

The core design is right: `get_DoA()` has no bounded latency (`TIMEOUT = 100000` ms, plus a retry loop of up to 100 attempts at 10 ms), so it cannot run on the GLib loop that carries video, audio and the pose push. A cache filled by a dedicated thread is the correct answer, and the two-lock split (cache lock never held across a USB transfer) is the right call — a single lock would park the 30 Hz push behind a control transfer.

Wire compat looks fine: `doa` is additive, pydantic ignores unknown fields so released SDKs are unaffected, `FullState.doa`'s type is unchanged so the regenerated `openapi.json` differs only by a description — and `build_state_json` already wraps `build_state_snapshot()` in `try/except`, so a raising `_doa_snapshot` can't kill the pose timer.

**Blocking:** the teardown race and the multi-handle contention (two comments on `abstract.py`). Everything else is discussion.

## Complexity

Separately from correctness: the machinery around the cache is carrying more than it earns. The first revision of this PR did the same job in ~45 lines; head is 146 + 211 test lines. Ranked by how much they remove:

| What | Where | Cut |
|---|---|---|
| Demand gating + idle stop + restart | `_note_doa_demand`, `_doa_demand_ts`, `DOA_IDLE_STOP_S`, the idle check | ~40 lines + 2 tests |
| `read_doa`'s direct-read fallback | `read_doa` | 13 lines + 3 tests |
| Cache freshness window | `DOA_CACHE_FRESH_S`, `_last_doa_ts`, the check in `_cached_doa` | ~12 lines + 1 test |
| Duplicate model | `DoaSnapshot` vs `DoAInfo` | ~13 lines, across 3 files |
| `_doa_available()` | 3-line wrapper, 2 callers | 3 lines |
| Internals in the public endpoint docstring | `state.py` + the `openapi.json` churn | 4 lines |

They compound: an always-on poller makes the fallback dead code, and clearing the cache on idle exit makes the freshness window unnecessary. Details are in the line comments.

One more with no single line to hang it on: **the rationale is stated five times.** "`get_DoA()` blocks on USB, don't call it from the 30 Hz push path" appears in the `__init__` comment (9 lines), the section header, three docstrings, and again at the `doa=` call site in `build_state_snapshot`. Once — in `_doa_poll_loop`'s docstring, where the thread is — and the rest can point at it. ~30 lines of the diff.

**net: ~−155 lines**, with no behaviour lost except `/state/doa` returning `null` instead of blocking during the ~100 ms before the first poll lands (see the `read_doa` comment — that one is a real trade-off, not a free cut).

src/reachy_mini/daemon/backend/abstract.py
tfrere: fixed in 96ddafd: close() now disposes the handle under _doa_usb_lock. also added the slow-read test you asked for, with a 1.5s fake read (past the join timeout) - checked it actually catches the bug by reverting the lock locally, it fails with a close-mid-read.

src/reachy_mini/daemon/app/routers/state.py
tfrere: went with your alternative in 96ddafd: read_doa() is cache-only now and never touches usb, so nothing blocks in these handlers anymore. kept them async since there's nothing left to move off the loop (and ws_full_state awaits get_full_state directly).

src/reachy_mini/daemon/backend/abstract.py
tfrere: applied as suggested in 96ddafd: read first, pace after, cache cleared on idle exit. the freshness window (DOA_CACHE_FRESH_S, _last_doa_ts) and its test are gone.

src/reachy_mini/daemon/backend/abstract.py
tfrere: took the trade-off in 96ddafd: cache only, /state/doa returns null for the ~100ms before the first poll lands instead of blocking. the three direct-read tests went with it.

src/reachy_mini/io/protocol.py
tfrere: done in 96ddafd, DoAInfo = DoaSnapshot alias in models.py and read_doa() returns the model directly. one heads-up: the openapi component gets renamed DoAInfo -> DoaSnapshot (payload unchanged), and the docs autodoc now points at the protocol model.

tests/unit_tests/test_backend_doa.py
tfrere: added in 96ddafd, plus the fake records a close() landing while a read is in flight, which is exactly the use-after-free case. the teardown test uses read_delay=1.5 and fails without the close() fix. cache-expiry-mid-read is moot now that the freshness window is gone.

tests/unit_tests/test_backend_doa.py
tfrere: dropped in 96ddafd, the test only checks the doa key now.

src/reachy_mini/daemon/backend/abstract.py
tfrere: took option 1 in 96ddafd: the audio-config branch now reuses the doa handle under _doa_usb_lock, held across the whole conversation. the transient handle only remains when no device was probed, i.e. no poller either. so the daemon side is down to one serialized handle (the media server's AudioDoA is still a second in-process handle, but the daemon never reads doa through it). cross-process is the open part: haven't run the conversation app against a warm poller yet, will do that on hardware and report here before merge.

src/reachy_mini/daemon/backend/abstract.py
tfrere: you're right that the comment was dishonest, rewrote it in 96ddafd to say what it actually gates: demand = any client reading state, since doa rides the snapshot by design. i kept the gating rather than dropping it because the benefit isn't zero: an idle daemon with no client connected (the normal state of an always-on robot) does zero usb traffic. the fallback-dead-code half of the cost is gone anyway since read_doa is cache-only now. if the ~30 remaining lines still feel too expensive once the conversation-app test says contention is a non-issue, i'm fine dropping it then.

src/reachy_mini/daemon/backend/abstract.py
FabienDanieau: Confirmed on hardware (`96ddafdb`, wireless robot, real ReSpeaker, nothing subscribed to the state stream): `/api/state/doa` on a 10 s cadence returns `null` every single time.

At 0.5 s intervals the first request is `null` and the rest are served. After 8 s idle → `null`; 0.5 s later → a reading. Daemon thread count tells the story: 41 idle → 42 while polling → 41 again 8 s after the last request. The poller exits on the idle timeout and clears `_last_doa`, then the next request respawns it and reads the cache it just cleared — and since nothing else refreshes demand, a client slower than `DOA_IDLE_STOP_S` repeats that forever.

This is the case the direct-read fallback used to cover and I argued it away, so the consequence is mine. `DOA_IDLE_STOP_S = 60.0` is the one-line fix: any plausible REST cadence stays warm, a genuinely idle daemon still goes quiet a minute later.

Invisible in validation because `examples/sound_doa.py` polls at ~2 Hz, and any state-stream subscriber refreshes demand at 30 Hz.

src/reachy_mini/io/protocol.py
```diff
@@ -119,6 +129,7 @@ class StateSnapshot(BaseModel):
     is_recording: bool
     is_move_running: bool
     face_target: FaceTarget
+    doa: Optional[DoaSnapshot] = None
```
FabienDanieau, line 132: No Python-side reader for this field: `WSClient._dispatch` caches `joint_positions`, IMU, status and recorded data, but there's no `get_state` or `subscribe_pose` consumer and no `doa` accessor in the SDK — the readers are the JS apps.

So `examples/sound_doa.py` still polls `/api/state/doa` over HTTP, the round-trip this PR replaces, and it lands on the wrong side of the idle-stop issue: a `subscribe_pose` consumer would keep the poller warm by construction.

Suggest a follow-up PR rather than growing this one — `subscribe_pose` + cache the snapshot in `_dispatch`, a `doa` accessor, then port the example. Is that already planned in the JS block PR, or separately?

FabienDanieau (approved):
Approving. The teardown race and the multi-handle contention are both properly fixed, and `test_close_waits_for_inflight_read` with a blocking fake is a better test than what I suggested. The cache/fallback/duplicate-model cleanups all landed too.

Two non-blocking notes inline. The one I'd actually want before this ships is `DOA_IDLE_STOP_S = 60.0`: a REST client polling slower than 5 s currently gets `null` forever, confirmed on hardware (details on the `read_doa` thread). One line.

Also pushed `528d4bbc` to this branch: `examples/sound_doa.py` opened a fresh TCP connection per poll and was timing out over Wi-Fi under its own WebRTC audio stream — it now reuses a `requests.Session` with a 3 s timeout.

src/reachy_mini/daemon/backend/abstract.py
tfrere: done in 5648741: DOA_IDLE_STOP_S = 60.0, and the comment now spells out the cadence constraint (idle exit clears the cache, so anything slower than the timeout would loop on the null it just re-primed). thanks for catching it on hardware, and for the sound_doa.py session fix.

src/reachy_mini/io/protocol.py
tfrere: separately - the js block pr (#1315) only touches ts/. i'll open a python sdk follow-up with exactly what you describe: subscribe_pose + snapshot caching in _dispatch, a doa accessor, then port examples/sound_doa.py onto it so the example stops doing the http round-trip this pr replaces.

PR #1311, opened 2026-08-03
feat(daemon): idle-reset handoff grace + wake stand-down

tfrere:

A client ending its session deliberately is often making room for a successor (e.g. the mobile app dropping its session so an app's iframe can take the slot), and a Space cold-start outlives the 1.5 s idle-reset debounce - the robot fell asleep mid-handoff.

- An `expect_handoff` flag travels from the relay's `endSession` paths through `RobotAppLock.release_remote` into `Backend.request_idle_reset`, which waits 15 s (tuned down from 30 after hardware testing) instead of the debounce.
- The grace is cancelled immediately when a remote session acquires the slot (`RobotAppLock.set_on_remote_acquired_handler` wired to `backend.cancel_idle_reset()` in the daemon), so a successful handoff never leaves a stale sleep timer behind.
- Companion fix: `wake_up` stand-down - see the behavior change section below, it is observable outside the handoff flow.

## Behavior change: `wake_up` stands down on an already-awake robot

This is a deliberate, observable change for the daemon-side `wake_up` callers: REST `/move/play_wake_up`, the WebRTC data-channel command, and the daemon's own boot/antenna wake paths.

> **Not covered: the Python SDK.** `ReachyMini.wake_up()` reimplements the emote client-side (goto + sound + roll) and never reaches `Backend.wake_up`, so Python apps still replay it on an awake robot. Fixing that needs a wake-up task request in the protocol to keep the method blocking - follow-up PR.

- **Before**: `wake_up` always replayed the full emote + "toudoum", even when the robot was already awake and standing at the init pose.
- **After**: when `Backend.is_awake_at_init_pose()` holds - motors enabled and init is the commanded target the controller is actively holding - `wake_up` is a silent no-op. The wake-completed callback still fires, so "start app after wake" semantics are preserved: the robot IS awake.
- `wake_up(force=True)` bypasses the stand-down for callers that need the emote unconditionally.

Rationale: clients routinely send `wake_up` on session start, and now that a handoff keeps the robot awake between sessions, replaying the animation on an already-awake robot reads as a glitch, not a wake. (Unlike `goto_sleep`'s stand-down, which still plays its cue, this one is deliberately silent - the sound on session start is the whole point.)

The "already awake" decision is a single predicate, `Backend.is_awake_at_init_pose()`, shared by `wake_up` and the antenna-touch startup path. **It compares the commanded targets, not the measured pose**: the first hardware pass (review below) showed a measured-pose comparison against the ideal init never fires - per-unit calibration residual ate the whole threshold, silently, on every wake. "Standing at init" now means "init is what the controller is holding": a head sagging a few mm under its own weight while commanded to init IS at init; a head an app parked elsewhere isn't. Both per-unit magic thresholds are gone; the remaining `atol` bounds the goto interpolation tail (a `play_move` loop never evaluates at exactly `t=duration`), a property of the trajectory, not of the robot.

Two knock-ons on the antenna path:

- It used to enable the motors before calling `wake_up()`, which could satisfy the stand-down it had just decided to bypass (`enable_motors` pins the targets to the measured pose); it now decides first and wakes with `force=True`.
- "Motors enabled but robot parked off init" (e.g. crashed app) used to get the cue-only branch and never re-home; it now runs the full wake.

## Known limitation

Any client death on the `endSession` path - tab kill, ICE failure, network drop - is indistinguishable from a deliberate hang-up, so those crashes get the same 15 s grace. The worst case is bounded: a client that "promised" a handoff and never came back leaves the robot awake for 15 s, then the normal idle reset runs. (Tightening this would need an explicit intent/reason field on the relayed `endSession`.)

Under `--sim`, the mujoco backend hardcodes `get_motor_control_mode() -> Enabled`, so the stand-down rests on the target check there. A fresh boot has no commanded target yet (the predicate's `None` guard), so `wake_up_on_start` still plays the full wake in simulation.

Pre-existing bug found during hardware review, not fixed here: `play_move`'s loop never evaluates a move at exactly `t=duration`, so every goto leaves the robot holding a target one eased tick short of the commanded endpoint. Core motion code, every move affected - follow-up PR.

## Behavioural pairing

Fail-open, any merge order: the JS block's sleep-on-leave pairs with this - without the grace the daemon may replay its own `goto_sleep` after a handoff.

## Tests

`test_robot_app_lock.py`, `test_backend_idle_reset.py`, `test_central_signaling_relay.py`, `test_backend_wake_up.py`, `test_autostart_app.py` (148 passed locally), including seam tests proving a pending handoff-grace reset task is cancelled by the successor's acquire. Handoff flows validated on hardware (mobile-to-app and web-host-to-app); the target-based stand-down was verified on hardware in review: 0.1 s no-op task, 0.001 rad peak roll, vs a 1.0 s / 0.216 rad replay before.

### DIFF
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -1154,12 +1154,31 @@ class Backend:
     )
 
     async def wake_up(self) -> None:
-        """Wake up the robot - go to the initial head position and play the wake up emote and sound."""
+        """Wake up the robot - go to the initial head position and play the wake up emote and sound.
+
+        Skips the emote (and its "toudoum") when the robot is already awake and
+        standing at the init pose: clients routinely send ``wake_up`` on
+        session start, and now that a handoff keeps the robot awake between
+        sessions (see ``request_idle_reset``'s handoff grace), replaying the
+        animation on an already-awake robot reads as a glitch, not a wake.
+        Mirrors ``goto_sleep``, which likewise stands down near its target
+        pose. The wake-completed callback still fires so the "start app after
+        wake" hook keeps its semantics: the robot IS awake.
+        """
         await asyncio.sleep(0.1)
 
         _, _, magic_distance = distance_between_poses(
             self.get_current_head_pose(), self.INIT_HEAD_POSE
         )
+
+        # Same empirical "close enough" threshold as goto_sleep's.
+        if (
+            self.get_motor_control_mode() == MotorControlMode.Enabled
+            and magic_distance < 10
+        ):
+            if self._on_wake_up_callback is not None:
+                self._on_wake_up_callback()
+            return
 
         await self.goto_target(
             self.INIT_HEAD_POSE,
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -1170,11 +1170,25 @@ class Backend:
         _, _, magic_distance = distance_between_poses(
             self.get_current_head_pose(), self.INIT_HEAD_POSE
         )
+        # The goto below is also what re-homes the antennas, so the
+        # stand-down must check them too: head at init but antennas left
+        # askew by the previous app would otherwise stay askew all session.
+        antenna_offset = float(
+            np.max(
+                np.abs(
+                    self.get_present_antenna_joint_positions()
+                    - self.INIT_ANTENNAS_JOINT_POSITIONS
+                )
+            )
+        )
 
-        # Same empirical "close enough" threshold as goto_sleep's.
+        # Head: same empirical "close enough" threshold as goto_sleep's.
+        # Antennas: ~17° of slack absorbs sag/jitter while still catching a
+        # genuinely misplaced antenna (sleep parks them ~165° away).
         if (
             self.get_motor_control_mode() == MotorControlMode.Enabled
             and magic_distance < 10
+            and antenna_offset < 0.3
         ):
             if self._on_wake_up_callback is not None:
                 self._on_wake_up_callback()
@@ -2713,11 +2727,7 @@ class Backend:
             if self.get_motor_control_mode() == MotorControlMode.Disabled:
                 return
             self._cancel_idle_reset()
-            self._idle_reset_task = asyncio.create_task(
-                self._async_idle_reset(
-                    self.IDLE_RESET_DEBOUNCE_S if grace_s is None else grace_s
-                )
-            )
+            self._idle_reset_task = asyncio.create_task(self._async_idle_reset(grace_s))
         except Exception:
             self.logger.warning("Idle reset scheduling failed", exc_info=True)
 
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -2610,7 +2610,18 @@ class Backend:
     # an intermediate pose.
     IDLE_RESET_DEBOUNCE_S: float = 1.5
 
-    def request_idle_reset(self) -> None:
+    # Longer grace used when the previous owner released the slot on purpose so
+    # another session could take over (see ``RobotAppLock.release_remote``).
+    # The canonical case is the mobile app dropping its own session to hand the
+    # robot to an app's iframe: the daemon only sees an empty slot, but a new
+    # session *is* coming - it just has to cold-start a Space (page load, auth,
+    # bundle, WebRTC handshake), which takes far longer than the debounce
+    # above. Sleeping in that window is exactly what the user doesn't want:
+    # they tapped an app, not "go to sleep". Long enough to cover a slow Space,
+    # short enough that a hand-off that never lands still parks the robot.
+    IDLE_RESET_HANDOFF_GRACE_S: float = 30.0
+
+    def request_idle_reset(self, *, expect_handoff: bool = False) -> None:
         """Return the robot to a clean idle state when no managed app owns it.
 
         Called by the daemon when the shared ``RobotAppLock`` transitions to
@@ -2627,11 +2638,22 @@ class Backend:
         one that also handles data-channel commands) and returns immediately.
         No-op when that loop isn't up yet (e.g. ``--no-media`` daemons, which
         have no remote sessions anyway).
+
+        Args:
+            expect_handoff: True when the slot was released on purpose for a
+                successor session, which earns the longer
+                ``IDLE_RESET_HANDOFF_GRACE_S`` instead of the default debounce.
+
         """
         loop = getattr(self, "_log_loop", None)
         if loop is None:
             return
-        loop.call_soon_threadsafe(self._maybe_start_idle_reset)
+        grace_s = (
+            self.IDLE_RESET_HANDOFF_GRACE_S
+            if expect_handoff
+            else self.IDLE_RESET_DEBOUNCE_S
+        )
+        loop.call_soon_threadsafe(self._maybe_start_idle_reset, grace_s)
 
     def cancel_idle_reset(self) -> None:
         """Cancel a pending/in-flight idle reset from any thread.
@@ -2654,7 +2676,7 @@ class Backend:
             task.cancel()
         self._idle_reset_task = None
 
-    def _maybe_start_idle_reset(self) -> None:
+    def _maybe_start_idle_reset(self, grace_s: float) -> None:
         """Kick off a goto_sleep if the robot is still awake. Runs on the loop."""
         try:
             if not self.ready.is_set() or self.is_shutting_down:
@@ -2665,17 +2687,21 @@ class Backend:
             if self.get_motor_control_mode() == MotorControlMode.Disabled:
                 return
             self._cancel_idle_reset()
-            self._idle_reset_task = asyncio.create_task(self._async_idle_reset())
+            self._idle_reset_task = asyncio.create_task(
+                self._async_idle_reset(
+                    self.IDLE_RESET_DEBOUNCE_S if grace_s is None else grace_s
+                )
+            )
         except Exception:
             self.logger.warning("Idle reset scheduling failed", exc_info=True)
 
-    async def _async_idle_reset(self) -> None:
+    async def _async_idle_reset(self, grace_s: float) -> None:
         """Graceful return to the sleep pose, which ends with motors disabled.
 
-        Starts with a short debounce (``IDLE_RESET_DEBOUNCE_S``): a fast
-        reconnect or a local app grabbing the robot cancels the task during that
-        window, so no motion happens at all on transient drops. Only if the slot
-        stays free past the grace period do we actually goto_sleep.
+        Starts with a debounce (``grace_s``): a fast reconnect or a local app
+        grabbing the robot cancels the task during that window, so no motion
+        happens at all on transient drops. Only if the slot stays free past the
+        grace period do we actually goto_sleep.
 
         Mirrors the reference leave behaviour clients already run by hand
         (gotoSleep -> motors off). ``goto_sleep()`` finishes with
@@ -2684,7 +2710,7 @@ class Backend:
         correctly triggers a fresh wake.
         """
         try:
-            await asyncio.sleep(self.IDLE_RESET_DEBOUNCE_S)
+            await asyncio.sleep(grace_s)
             # Conditions may have changed during the grace period (shutdown
             # started, or the robot was already put to sleep by whoever briefly
             # held the slot). Re-check before committing to the trajectory.
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -2638,7 +2638,14 @@ class Backend:
     # above. Sleeping in that window is exactly what the user doesn't want:
     # they tapped an app, not "go to sleep". Long enough to cover a slow Space,
     # short enough that a hand-off that never lands still parks the robot.
-    IDLE_RESET_HANDOFF_GRACE_S: float = 30.0
+    #
+    # Sized down from an initial 30 s after hardware testing: real handoffs
+    # re-acquire the slot within 1-4 s, and abrupt client deaths (tab kill)
+    # ride the same local endSession path as deliberate hang-ups - the relay
+    # cannot tell them apart - so this grace is also how long a crashed
+    # client keeps the robot needlessly awake. A Space colder than this just
+    # gets a legitimate wake animation when it finally connects.
+    IDLE_RESET_HANDOFF_GRACE_S: float = 15.0
 
     def request_idle_reset(self, *, expect_handoff: bool = False) -> None:
         """Return the robot to a clean idle state when no managed app owns it.
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -1153,26 +1153,34 @@ class Backend:
         ]
     )
 
-    # Init-pose proximity in magic-mm (mm + deg), same empirical threshold as
-    # goto_sleep's stand-down (see SLEEP_POSE_MAGIC_DISTANCE, apps/manager.py).
-    INIT_POSE_MAGIC_DISTANCE: float = 10.0
-    # ~17° of antenna slack: absorbs sag/jitter while still catching a
-    # genuinely misplaced antenna (sleep parks them ~165° away).
-    INIT_ANTENNAS_TOLERANCE_RAD: float = 0.3
+    # Bounds the goto interpolation tail, not the hardware: play_move never
+    # evaluates a move at exactly t=duration, so the last written target sits
+    # a (velocity-eased) tick short of the commanded endpoint.
+    INIT_TARGET_ATOL: float = 1e-2
 
     def is_awake_at_init_pose(self) -> bool:
-        """Motors on and head + antennas already home: nothing to wake."""
-        _, _, head_offset = distance_between_poses(
-            self.get_current_head_pose(), self.INIT_HEAD_POSE
-        )
+        """Motors on and init is the pose the controller is actively holding.
+
+        Compares the commanded targets, not the measured pose: a head sagging
+        a few mm under its own weight while commanded to init IS standing at
+        init, and per-unit calibration residual must not flip the answer.
+        """
+        if self.target_head_pose is None or self.target_antenna_joint_positions is None:
+            return False
         return (
             self.get_motor_control_mode() == MotorControlMode.Enabled
-            and head_offset < self.INIT_POSE_MAGIC_DISTANCE
             and bool(
                 np.allclose(
-                    self.get_present_antenna_joint_positions(),
+                    self.target_head_pose,
+                    self.INIT_HEAD_POSE,
+                    atol=self.INIT_TARGET_ATOL,
+                )
+            )
+            and bool(
+                np.allclose(
+                    self.target_antenna_joint_positions,
                     self.INIT_ANTENNAS_JOINT_POSITIONS,
-                    atol=self.INIT_ANTENNAS_TOLERANCE_RAD,
+                    atol=self.INIT_TARGET_ATOL,
                 )
             )
         )
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -1153,47 +1153,50 @@ class Backend:
         ]
     )
 
-    async def wake_up(self) -> None:
-        """Wake up the robot - go to the initial head position and play the wake up emote and sound.
+    # Init-pose proximity in magic-mm (mm + deg), same empirical threshold as
+    # goto_sleep's stand-down (see SLEEP_POSE_MAGIC_DISTANCE, apps/manager.py).
+    INIT_POSE_MAGIC_DISTANCE: float = 10.0
+    # ~17° of antenna slack: absorbs sag/jitter while still catching a
+    # genuinely misplaced antenna (sleep parks them ~165° away).
+    INIT_ANTENNAS_TOLERANCE_RAD: float = 0.3
 
-        Skips the emote (and its "toudoum") when the robot is already awake and
-        standing at the init pose: clients routinely send ``wake_up`` on
-        session start, and now that a handoff keeps the robot awake between
-        sessions (see ``request_idle_reset``'s handoff grace), replaying the
-        animation on an already-awake robot reads as a glitch, not a wake.
-        Mirrors ``goto_sleep``, which likewise stands down near its target
-        pose. The wake-completed callback still fires so the "start app after
-        wake" hook keeps its semantics: the robot IS awake.
-        """
-        await asyncio.sleep(0.1)
-
-        _, _, magic_distance = distance_between_poses(
+    def is_awake_at_init_pose(self) -> bool:
+        """Motors on and head + antennas already home: nothing to wake."""
+        _, _, head_offset = distance_between_poses(
             self.get_current_head_pose(), self.INIT_HEAD_POSE
         )
-        # The goto below is also what re-homes the antennas, so the
-        # stand-down must check them too: head at init but antennas left
-        # askew by the previous app would otherwise stay askew all session.
-        antenna_offset = float(
-            np.max(
-                np.abs(
-                    self.get_present_antenna_joint_positions()
-                    - self.INIT_ANTENNAS_JOINT_POSITIONS
+        return (
+            self.get_motor_control_mode() == MotorControlMode.Enabled
+            and head_offset < self.INIT_POSE_MAGIC_DISTANCE
+            and bool(
+                np.allclose(
+                    self.get_present_antenna_joint_positions(),
+                    self.INIT_ANTENNAS_JOINT_POSITIONS,
+                    atol=self.INIT_ANTENNAS_TOLERANCE_RAD,
                 )
             )
         )
 
-        # Head: same empirical "close enough" threshold as goto_sleep's.
-        # Antennas: ~17° of slack absorbs sag/jitter while still catching a
-        # genuinely misplaced antenna (sleep parks them ~165° away).
-        if (
-            self.get_motor_control_mode() == MotorControlMode.Enabled
-            and magic_distance < 10
-            and antenna_offset < 0.3
-        ):
+    async def wake_up(self, *, force: bool = False) -> None:
+        """Wake up the robot - go to the initial head position and play the wake up emote and sound.
+
+        Stands down (silently - no motion, no sound) when the robot is already
+        awake at the init pose, so a boot-time ``wake_up`` after a session
+        handoff doesn't replay the emote. The wake-completed callback still
+        fires either way: the robot IS awake. ``force=True`` bypasses the
+        stand-down, e.g. for a deliberate replay or a caller that just
+        enabled the motors itself and knows a real wake is due.
+        """
+        await asyncio.sleep(0.1)
+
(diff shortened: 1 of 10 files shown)


tfrere:
Heads up for review: this PR intentionally changes one observable behavior outside the handoff flow. Flagging it here so it doesn't have to be discovered in the diff.

## `wake_up` now stands down on an already-awake robot

Before: `wake_up` always replayed the full emote + toudoum, even on a robot already awake and standing at init.
After: if motors are enabled and head AND antennas are already at the init pose (~17 deg antenna slack), it's a silent no-op. The wake-completed callback still fires, so the "start app after wake" hook keeps its semantics: the robot IS awake.

Why: clients send `wake_up` on every session start, and with the handoff grace the robot now stays awake between sessions, so the emote was replaying on an already-awake robot and reads as a glitch. It also mirrors `goto_sleep`, which has always stood down near its target pose: both commands become idempotent goal-seekers rather than emotes.

In practice:

- robot woken twice in a row plays the animation once
- fresh session right after a handoff: no toudoum, the robot just stays put
- a genuinely asleep or misplaced robot (motors off, head off init, or antennas askew) wakes exactly as before

If you'd rather keep the old animation reachable (deliberate greeting use case), an optional `force` flag bypassing the stand-down is a ~10 line additive change, happy to add it.

The stand-down cases are covered in `test_backend_wake_up.py` (already-at-init, disabled-at-init, antennas-askew), and both handoff flows were validated on hardware.

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -1154,12 +1154,45 @@ def clear_incoming_audio(self) -> None:
...
+        """
         await asyncio.sleep(0.1)
 
         _, _, magic_distance = distance_between_poses(
             self.get_current_head_pose(), self.INIT_HEAD_POSE
         )
+        # The goto below is also what re-homes the antennas, so the
+        # stand-down must check them too: head at init but antennas left
+        # askew by the previous app would otherwise stay askew all session.
+        antenna_offset = float(
```
FabienDanieau: This is the third place in the codebase that decides "is the robot already awake", and all three answer differently (table in the review summary). Suggest extracting the predicate:

```python
INIT_POSE_MAGIC_DISTANCE = 10.0        # mirrors SLEEP_POSE_MAGIC_DISTANCE, apps/manager.py:26
INIT_ANTENNAS_TOLERANCE_RAD = 0.3      # ~17 deg: absorbs sag, catches a parked antenna (~165 deg off)

def is_awake_at_init_pose(self) -> bool:
    """Motors on and head+antennas already home: nothing to wake."""
    _, _, head_offset = distance_between_poses(
        self.get_current_head_pose(), self.INIT_HEAD_POSE
    )
    return (
        self.get_motor_control_mode() == MotorControlMode.Enabled
        and head_offset < self.INIT_POSE_MAGIC_DISTANCE
        and np.allclose(
            self.get_present_antenna_joint_positions(),
            self.INIT_ANTENNAS_JOINT_POSITIONS,
            atol=self.INIT_ANTENNAS_TOLERANCE_RAD,
        )
    )
```

That folds these 8 lines into one `np.allclose`, and gives the two magic numbers names next to the existing `SLEEP_POSE_MAGIC_DISTANCE`. Here it becomes just the predicate + callback + `return`.

Then `startup_app.wake_or_start_startup_app_if_idle` (`startup_app.py:196-212`) branches on the same predicate instead of `motor_mode == Disabled`, which lets `play_awake_startup_cue` go away (3-line wrapper, one caller):

```python
if backend.is_awake_at_init_pose():
    backend.play_sound("wake_up.wav")          # ack the touch
else:
    backend.set_motor_control_mode(MotorControlMode.Enabled)
    await backend.wake_up()
    await asyncio.sleep(0)
    if not _startup_app_slot_is_free(app_manager, daemon):
        return True
```

One trap if you do it: evaluate the predicate **before** `set_motor_control_mode(Enabled)` there, or enabling the motors makes it true and an antenna touch produces nothing at all.

Knock-on worth a line in the description: today "motors enabled but head/antennas askew" (crashed app) takes the cue-only branch and never re-homes; with the predicate it runs the full wake. Better, but observable.

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -1154,12 +1154,45 @@ def clear_incoming_audio(self) -> None:
...
     async def wake_up(self) -> None:
-        """Wake up the robot - go to the initial head position and play the wake up emote and sound."""
+        """Wake up the robot - go to the initial head position and play the wake up emote and sound.
+
+        Skips the emote (and its "toudoum") when the robot is already awake and
+        standing at the init pose: clients routinely send ``wake_up`` on
+        session start, and now that a handoff keeps the robot awake between
+        sessions (see ``request_idle_reset``'s handoff grace), replaying the
+        animation on an already-awake robot reads as a glitch, not a wake.
+        Mirrors ``goto_sleep``, which likewise stands down near its target
```
FabienDanieau: Not quite accurate — `goto_sleep`'s stand-down still plays `go_sleep.wav` (line ~1230) and only skips the *motion*, so the house convention for a stand-down is cue-yes/motion-no. Yours is fully silent.

I think silent is the right call here — the toudoum on session start is the whole point of the PR — so I'd drop or qualify this sentence rather than change the behaviour to match.

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -2610,7 +2643,25 @@ def _send_webrtc_response(self, peer_id: str, response: dict[str, Any]) -> None:
...
+    # they tapped an app, not "go to sleep". Long enough to cover a slow Space,
+    # short enough that a hand-off that never lands still parks the robot.
+    #
+    # Sized down from an initial 30 s after hardware testing: real handoffs
+    # re-acquire the slot within 1-4 s, and abrupt client deaths (tab kill)
+    # ride the same local endSession path as deliberate hang-ups - the relay
+    # cannot tell them apart - so this grace is also how long a crashed
+    # client keeps the robot needlessly awake. A Space colder than this just
+    # gets a legitimate wake animation when it finally connects.
+    IDLE_RESET_HANDOFF_GRACE_S: float = 15.0
```
FabienDanieau, line 2810: 20 lines of comment for one float, and it's the same content as the two commit messages (which is the right place for it). Two lines covers it: what the grace covers (Space cold start) and what it bounds (how long a dead client keeps the robot awake).

Same for the 9 rationale lines added to the `wake_up` docstring above.

src/reachy_mini/media/central_signaling_relay.py
```diff
@@ -705,7 +705,11 @@ async def _run_loop(self) -> None:
                             f"Connection failed after {self._connection_attempts} attempts: {e}",
                         )
 
-                if self._running and not had_exception and self._state == RelayState.ERROR:
+                if (
+                    self._running
+                    and not had_exception
```
FabienDanieau, line 710: Unrelated reflow — worth dropping to keep the diff to the feature.

src/reachy_mini/media/central_signaling_relay.py
```diff
@@ -1604,13 +1611,14 @@ async def _process_local_message(self, msg: dict[str, Any]) -> None:
...
-                # If no sessions remain, release the robot lock.
+                # If no sessions remain, release the robot lock. Deliberate
+                # hang-up, so same hand-off treatment as the central-side path.
                 if (
                     self._robot_app_lock is not None
                     and not self._central_to_local_session
                     and not self._pending_central_sessions
                 ):
-                    self._robot_app_lock.release_remote()
+                    self._robot_app_lock.release_remote(expect_handoff=True)
```
FabienDanieau, line 1621: This path fires for anything GStreamer surfaces as `endSession`, not only a deliberate hang-up: ICE failure and network drop ride it too, alongside the tab kill you already call out. Same bounded worst case, so no code change needed — but the "Known limitation" section should say *any client death*, not just a tab kill.

If it's ever worth tightening, the only real signal is an explicit intent/`reason` field on the relayed `endSession`, which would let genuine drops keep the 1.5 s debounce.

src/reachy_mini/daemon/daemon.py
```diff
@@ -250,7 +250,7 @@ async def _stop_central_signaling_relay(self) -> None:
         except Exception as e:
             self.logger.debug(f"Error stopping central signaling relay: {e}")
 
-    def _on_robot_slot_free(self) -> None:
+    def _on_robot_slot_free(self, expect_handoff: bool = False) -> None:
```
FabienDanieau: The default is unreachable — the lock always passes the flag. Fine to drop it and keep the signature honest.

tests/unit_tests/test_backend_idle_reset.py
```diff
@@ -70,12 +73,34 @@ async def test_idle_reset_skips_when_already_disabled() -> None:
...
+
+    Regression guard for the mobile app dropping its session so an app iframe
+    can take the slot: the short debounce expires long before a Space has
+    finished cold-starting, and the robot would fall asleep in that gap.
+    """
+    scheduled: list[tuple] = []
+    fake = _fake_backend()
+    fake._maybe_start_idle_reset = lambda *args: None
+    fake._log_loop = SimpleNamespace(
+        call_soon_threadsafe=lambda fn, *args: scheduled.append(args)
```
FabienDanieau: The lambda captures `args` but discards `fn`, so this passes even if the wrong callable gets posted. Capturing `(fn, args)` pins it to `_maybe_start_idle_reset`.

Bigger gap while you're here: the daemon seam is untested. `set_on_remote_acquired_handler` → `_on_robot_slot_acquired` → cancel is covered on the lock side (4 tests, nice) and on the backend side, but nothing checks that an actually-*pending* handoff-grace reset gets cancelled by a remote acquire — which is the gap PATCH 4 exists to close.

FabienDanieau (changes requested):
Mechanism looks right, and I like that the grace is cancelled on acquire rather than waiting for the successor's first data-channel command. I checked the concurrency and it holds up: `_fire_remote_acquired()` is correctly outside the mutex and after the `return False` branch, release-then-acquire both run on the relay loop so the cancel is always posted *after* the schedule, and `_maybe_start_idle_reset` cancels-and-reschedules so the newest grace wins. Test coverage is genuinely good.

I ran the branch locally (worktree on `feat/daemon-handoff-grace`, 4 commits on top of `66c3492d5`):

- **Your four test files: 108 passed** — claim confirmed.
- Full `tests/unit_tests`: 768 passed, 25 failed, 21 skipped. **All 25 failures are pre-existing** — I ran the same eight files on `main` and got the identical 25 (no running daemon, no ReSpeaker board, WebRTC/GStreamer backend, `placo` not installed). Nothing here is on you.
- `ruff check` clean; the eight files you touched are `ruff format`-clean (the 21 files it flags are the repo's existing drift). `mypy` reports one error, `tools/camera_calibration/calibrate.py:202`, identical on `main`.
- Setup note for anyone repeating this: `uv sync` alone can't collect `test_router_update.py`, `test_wifi_connect_retry.py` or `test_wireless_version.py` — `semver` lives in the `wireless-version` extra, so a full collection needs `uv sync --extra wireless-version`.

Two things before merge, one blocking.

## Blocking: the Python SDK doesn't go through the code you changed

`ReachyMini.wake_up()` (`src/reachy_mini/reachy_mini.py:724`) is a separate client-side reimplementation of the emote — `goto_target` → `play_sound("wake_up.wav")` → roll → back. It never calls `Backend.wake_up()`. So "a deliberate, observable change for **all** `wake_up` callers (Python SDK, REST `/move/play_wake_up`, WebRTC command)" isn't accurate: every Python app still replays the emote on an already-awake robot, which is exactly the glitch this PR exists to remove, on the surface where most apps live.

I don't think the fix belongs here, though. `WakeUpCmd` already exists (`io/protocol.py:215`) and is in `AnyCommand`, so the SDK method could collapse to `self.client.send_command(WakeUpCmd())` and inherit the stand-down for free — except `AnyTaskRequest = GotoTaskRequest | PlayMoveTaskRequest` (`protocol.py:985`) has no wake-up task, so there's no completion signal and the method would stop blocking (~2.5 s → instant). That breaks user scripts. So: follow-up PR (add a wake-up task request, or duplicate the predicate client-side the way `ts/animation/distance.ts` already duplicates the distance math), and correct the claim in this description meanwhile.

## Discuss: there are now three "already awake" policies, with three different answers

| Where | Test | On awake |
|---|---|---|
| `Backend.wake_up` (this PR) | motors Enabled + head + antennas at init | nothing at all |
| `startup_app.wake_or_start_startup_app_if_idle:200` | `motor_mode == Disabled` | sound only (`play_awake_startup_cue`) |
| `Backend.goto_sleep:1201-1230` | near the *sleep* pose | **plays its cue**, skips the motion |

Details in the inline comment on the stand-down block — short version is one `Backend.is_awake_at_init_pose()` used by both wake paths.

## Non-blocking, worth knowing

`backend/mujoco/backend.py:337-339` hardcodes `get_motor_control_mode() -> Enabled` regardless of `set_motor_control_mode`, so the motor-mode half of your guard is a no-op under `--sim` and the stand-down rests entirely on the pose check. Worth one smoke run: if the scene's initial qpos sits near init, `wake_up_on_start` goes silent in simulation. (`mockup_sim` is fine — it boots at `SLEEP_HEAD_JOINT_POSITIONS`.)

And one design limit, not asking you to fix it: the handoff needs strict release-before-acquire. `try_acquire_remote` returns `False` while the state is still `REMOTE_SESSION` (`robot_app_lock.py:286-291`), so a successor whose `startSession` lands before the predecessor's `endSession` gets `robot_busy` and the grace can't help it. That ordering holds for the Space cold-start flow this targets — just don't count on the grace for an overlapping handoff.

tfrere:
thanks for the deep review - all addressed in the last two commits + description update:

- SDK claim: you're right, `ReachyMini.wake_up()` never reaches the backend. description now scopes the change to daemon-side callers and calls out the SDK gap; follow-up PR for the wake-up task request so the method can stay blocking.
- extracted `is_awake_at_init_pose()` with named thresholds and routed the antenna-touch path through it. while doing it I found the trap you warned about already existed there: the path enabled the motors *before* calling `wake_up()`, so a robot resting at init would satisfy the stand-down it had just decided to bypass. it now evaluates the predicate first and wakes with the new `force=True`. the enabled-but-askew case runs a full wake now, noted in the description.
- dropped the goto_sleep mirror claim from the docstring (correct, its stand-down still plays the cue - this one is silent on purpose).
- grace comment down to 2 lines, docstring rationale dropped, unreachable default gone.
- the `_run_loop` reflow stays: turns out the pre-commit ruff-format hook rewrites that line whenever the file is touched, reverting it just fights the hook.
- test lambda now captures the callable too, and added the seam tests: a real `RobotAppLock` wired the way the daemon wires it, with a genuinely pending grace reset cancelled by `try_acquire_remote` (plus the positive control where no successor shows up and the robot sleeps).
- also moved both handler registrations before the relay start, closing the small boot window where a session could acquire unwired.
- mujoco: confirmed `get_motor_control_mode` hardcodes Enabled, so under --sim the stand-down rests on the pose check alone - noted in the description. mockup_sim boots at the sleep pose so wake_up_on_start is unaffected there.

146 passed locally across the five test files.

src/reachy_mini/daemon/backend/abstract.py
tfrere: done in 0fec7de: extracted is_awake_at_init_pose() with the named thresholds and routed the antenna-touch path through it. also caught a live version of the trap you flagged: that path enabled the motors before calling wake_up, so a robot resting at init would satisfy the stand-down it had just decided to bypass. it now evaluates the predicate first and wakes with force=True.

src/reachy_mini/daemon/backend/abstract.py
tfrere: right, dropped the sentence - this stand-down is silent on purpose, no cue.

src/reachy_mini/daemon/backend/abstract.py
tfrere: trimmed to 2 lines in 5e852d3, same for the wake_up docstring rationale - the long version lives in the commit messages.

src/reachy_mini/media/central_signaling_relay.py
tfrere: tried to revert it and the pre-commit ruff-format hook immediately rewrote the line (it is 91 chars unwrapped), so it stays - reverting just fights the hook.

src/reachy_mini/media/central_signaling_relay.py
tfrere: description updated: known limitation now says any client death on the endSession path (tab kill, ICE failure, network drop), and mentions the intent/reason field as the way to tighten it later.

src/reachy_mini/daemon/daemon.py
tfrere: dropped in 5e852d3, the signature is now honest: expect_handoff is required.

tests/unit_tests/test_backend_idle_reset.py
tfrere: both done in 5e852d3: the lambda captures (fn, args) and pins the schedule to _maybe_start_idle_reset, and the daemon seam is covered - a real RobotAppLock wired the way the daemon wires it, with a genuinely pending handoff-grace reset cancelled by try_acquire_remote, plus the positive control where no successor arrives and goto_sleep fires.

FabienDanieau:
## Hardware follow-up: the wake stand-down never fires, and I think I have the fix

Tested the branch at `5e852d3c2` on a wireless unit (editable checkout, daemon on :8000). The follow-up commits all check out — 146 passed, `ruff`/`mypy` clean, and the antenna-touch `force=True` path works end to end (booted asleep, touch → motors enabled → full wake → startup app running, 2.4 s). But the headline behaviour doesn't happen on real hardware.

### Symptom

`wake_up` on a robot that is unambiguously awake and standing at init replays the full emote, every time. Three consecutive REST calls, identical:

| | measured |
|---|---|
| move task duration | 1.0 s |
| peak head roll | 0.216 rad (12°) |

### Root cause 1: the predicate compares a measured pose against an ideal one

`INIT_HEAD_POSE` is the identity matrix, so `is_awake_at_init_pose()` spends the unit's entire calibration residual before the threshold is even consulted. On this robot, holding the init pose it had just been commanded to, motors on, dead still:

```
head translation :   5.95–6.05 mm
head rotation    :   7.99–8.50 deg
head MAGIC       :  14.0–14.5     threshold INIT_POSE_MAGIC_DISTANCE = 10.0   → FAIL
antenna offset   :   0.008 rad                                                → PASS

is_awake_at_init_pose() → False
```

Stable across 10 s of sampling and two independent setups. Not a measurement artefact: `get_current_head_pose()` is literally `return self.get_present_head_pose()` (`abstract.py:1023`), so the predicate reads exactly the value I sampled.

The failure is silent — no log, no error, hardcoded threshold, no per-unit knob. Any robot whose calibration residual exceeds 10 magic-mm quietly keeps replaying the animation.

### Root cause 2 (separate, pre-existing bug): gotos never write their exact endpoint

Found while fixing the above. `play_move`'s loop:

```python
while time.time() - t0 < move.duration:
    t = time.time() - t0
    head, antennas, body_yaw = move.evaluate(t)
    if head is not None:
        self.set_target_head_pose(head)
```

The condition is `elapsed < duration` and evaluation only ever happens at `t < duration`, so the endpoint is never written. **Every `goto_target` leaves the robot holding a target one tick short of what was commanded**, permanently, since the controller then holds that penultimate target.

Small in magnitude (a smooth easing's last tick covers ~0.1% of the travel, velocity → 0), but it means "the commanded target equals X" is never exactly true after a goto. Worth fixing independently of this PR.

The clean fix is to evaluate at `move.duration` after the loop — but `RecordedMove.evaluate` raises on `t >= timestamps[-1]` (`recorded_move.py:113`), so it needs an isinstance/try guard. I didn't touch it: core motion code, every move affected, your call.

### The fix, verified on hardware

Compare the *commanded target* instead of the measured pose:

```python
def is_awake_at_init_pose(self) -> bool:
    # Motors on and init is the pose we are actively holding.
    if self.target_head_pose is None or self.target_antenna_joint_positions is None:
        return False
    return (
        self.get_motor_control_mode() == MotorControlMode.Enabled
        and bool(np.allclose(self.target_head_pose, self.INIT_HEAD_POSE, atol=1e-2))
        and bool(
            np.allclose(
                self.target_antenna_joint_positions,
                self.INIT_ANTENNAS_JOINT_POSITIONS,
                atol=1e-2,
            )
        )
    )
```

Result on the same robot, same sequence, three consecutive runs:

| | before | after |
|---|---|---|
| move task duration | 1.0 s | **0.1 s** |
| peak head roll | 0.216 rad | **0.001 rad** |
| peak antenna deviation | 0.008 rad | **0.000 rad** |

Full wake from sleep still works (2.3 s task, 0.48 rad pitch swing, antennas 165° from sleep to init). The residual 0.1 s task is just `wake_up`'s opening `await asyncio.sleep(0.1)` before it stands down; `create_move_task` registers a task regardless of whether the body of `wake_up` moves anything.

Why this is better than retuning the threshold:

- **No calibration sensitivity.** Both `INIT_POSE_MAGIC_DISTANCE` and `INIT_ANTENNAS_TOLERANCE_RAD` disappear. Nothing to tune per unit.
- **The remaining `atol=1e-2` bounds the interpolation tail, not the hardware** — a property of the trajectory, not of the robot. Still ~30× tighter than the 0.3 rad the antenna check allowed, and immune to the 8.5° bias that broke the original.
- **Better semantics.** "Standing at init" should mean "init is what it's holding". A head sagging 6 mm under its own weight while commanded to init *is* standing at init; a head an app parked elsewhere isn't, however the numbers land.
- **Probably deletes `force=True`.** Enabling the motors re-pins the Cartesian target to the measured pose (`robot/backend.py:344`), so the predicate goes False and the wake runs — which is exactly the trap `force` was added to work around. Worth checking; if it holds, the flag and its single call site go.

Cost to report: the six `test_backend_wake_up.py` cases need their fake to carry `target_head_pose` / `target_antenna_joint_positions` instead of only measured poses. I mapped target ← the same values and all six keep their original intent (49 passed across the wake, autostart and idle-reset files, `ruff`/`mypy` clean). `_far_pose()`'s docstring referencing "the 10 threshold" goes stale.

### Caveats, stated plainly

- **One robot.** A better-calibrated unit might come in under 10 — which is itself the argument against an absolute threshold measured from an ideal.
- **I could not verify the sound by ear**, only that no motion occurs and `play_sound` isn't on the stand-down path.
- **The 15 s handoff grace itself is still unverified by me** — it needs a real remote session from the mobile app. Everything above concerns the companion `wake_up` change only.
- Reproducing gotcha: a REST `wake_up` fires the one-shot startup-app hook, so on a normally-configured robot it launches your startup app, which then drives the head and contaminates the measurement. Clear `startup-app` first.

tfrere:
great catch, thanks for going all the way to hardware on this. took your predicate as-is in 4f5c429 - target comparison, both magic thresholds gone, the atol is now INIT_TARGET_ATOL with the interpolation-tail rationale in a comment.

two deliberate deviations from your notes:

- kept force=True on the antenna path. enable_motors pins the targets to the *measured* pose, and on a well-calibrated unit resting at init that pin can land within atol - which re-creates exactly the swallowed-touch trap force was added for. your unit happens to fail the predicate on its rotation residual, but that's the calibration-dependence we just removed, so I'd rather not lean on it.
- the play_move endpoint bug stays out, agreed it's core motion code and a follow-up - noted in the description next to the SDK wake-task one.

nice side effect of the None guard: under --sim a fresh boot has no commanded target yet, so wake_up_on_start keeps its full wake there - your smoke-run concern is covered structurally now.

tests: mapped target <- same values as you did so your six keep their intent, plus two new ones (no-target-yet boot, half-atol interpolation tail). 148 passed locally, ruff/mypy clean, _far_pose docstring de-staled. description updated with the target-based predicate, the sim note and your hardware numbers.

FabienDanieau (approved):
Reviewed `4f5c42970`, and the blocking hardware issue is resolved. Verified on the same wireless unit as the earlier follow-up.

## The stand-down fix works on hardware

Re-ran the failing sequence against the committed code (not my patched version — the two are semantically identical, but the numbers below are from the commit):

| | before | `4f5c42970` |
|---|---|---|
| move task duration | 1.0 s | **0.1 s** |
| peak head roll | 0.216 rad (12°) | **0.000 rad** |
| peak antenna deviation | 0.008 rad | **0.000–0.002 rad** |

Three consecutive stand-downs, identical. Full wake from sleep unchanged: 2.3 s task, 0.47 rad pitch swing, antennas 165° sleep→init. The residual 0.1 s task is just `wake_up`'s opening `await asyncio.sleep(0.1)` before it returns; `create_move_task` registers a task regardless of whether anything moves.

**You were right to keep `force=True`, and the reasoning is better than what I proposed.** I suggested the flag might become deletable; it can't. `enable_motors` pins the targets to the measured pose, so on a *well-calibrated* robot resting at init that pinned target lands inside `atol`, the predicate flips True, and the touch's wake gets swallowed — exactly the trap the flag exists for. It only looked deletable because this unit's 8.5° residual puts the pinned target far outside `atol`.

The `--sim` gap is closed properly too: `test_wake_up_runs_when_no_target_commanded_yet` covers mujoco hardcoding motor mode to `Enabled`, where the `None` guard keeps `wake_up_on_start` audible.

Also confirmed by the reviewer operating the robot directly (observed, not instrumented): **the 15 s handoff grace behaves as designed** — the robot stays awake after a deliberate disconnect instead of dropping off in ~1.5 s, and survives a quick reconnect. And **antenna touch works**. That was the last untested part of this PR.

148 passed, `ruff check` clean, touched files format-clean, `mypy` unchanged from `main`, no stale references to the removed constants.

## One non-blocking point: the boot wake didn't get the same treatment

`daemon.py:474-475`:

```python
self.backend.set_motor_control_mode(MotorControlMode.Enabled)
await self.backend.wake_up()          # no force=
```

That's the identical enable-then-wake sequence the antenna path got `force=True` for, and by the argument in your own commit message it has the same exposure: on a well-calibrated robot resting at init, enabling the motors pins the target within `atol`, the predicate returns True, and the boot wake is silently swallowed.

Invisible on the unit I tested (residual too large) and unreachable on the wireless unit (`--no-wake-up-on-start` skips the path entirely), so it only bites a well-calibrated robot on a config that wakes at start — the desktop app and the default. Either add `force=True` or a line saying why boot differs.

## Two pre-existing things this PR runs into, for the follow-up

**The `play_move` endpoint bug** you've already deferred — agreed it's out of scope. Worth noting in that PR that `INIT_TARGET_ATOL` can shrink once it lands: the tail it absorbs scales with `dt/duration`, so it's ~3.4e-3 rad for the wake's long antenna goto but grows sharply for short ones (a 0.05 s goto loses ~5.8% of its travel). Fail-safe in that direction — a spurious full wake, never a missed one — but it means the constant is still empirically sized, which is the thing this commit set out to eliminate.

**`_start_central_signaling_relay()` fails silently.** It returns early on `if not self._media_server` with no log at any level, so when media init fails (missing `webrtcsink` in my case), "remote access is off" is indistinguishable from "remote access was never requested". Cost me a long detour before I worked out why no client could reach the robot. Not introduced here, but this PR's headline feature is unreachable without the relay — one `logger.warning` in that early return would pay for itself.

Nit: `test_wake_up_stands_down_despite_calibration_residual` is misnamed — its own docstring says hardware residual is out of the picture and the mechanism is the interpolation tail. `..._despite_interpolation_tail` matches what it tests.

Approving. Nice work on the turnaround, particularly catching the enable-before-predicate bug in the antenna path that my review only flagged as a hazard to avoid.

PR #1312, opened 2026-08-03
fix(update): fail the update job on a non-zero install exit code

tfrere:

`call_logger_wrapper` streamed the install output but never checked the exit code, so a failed `pip install` sailed through: the daemon restarted on the OLD version while every route reported success.

- Now raises on a non-zero exit, with a per-step failure policy: daemon venv install is fatal before restart, `apps_venv` is non-fatal, restart is fatal.
- The restart keeps the exact whitelisted argv `sudo systemctl restart reachy-mini-daemon` and tolerates `(0, -15, 143)`. It has to stay verbatim: `/etc/sudoers.d/010-pollen-reachy` matches the argv literally, so any extra flag (`--no-block` included) makes `sudo` demand a password and the restart never runs. And since restarting our own cgroup SIGTERMs this `systemctl` child, `-15`/`143` is the normal outcome of a *successful* restart, not a failure; anything else (unknown unit, sudo refusal) still raises. `call_logger_wrapper` gained an `ok_returncodes` param (default `(0,)`), so every other call site stays strict. Verified on a real robot from inside the daemon's cgroup.
- Also fixes the `step1 && step2 || step3` precedence bug in the git-ref install chain. Bonus: this same chain is built by `check_and_sync_apps_venv_sdk` (`startup_check.py`), which gates on `returncode` - it used to log "Successfully synced" after quietly installing the latest PyPI wheel instead of the requested ref.

Failure archaeology (kept out of the code comments on purpose):

- daemon venv install failure aborts BEFORE the restart. On the PyPI path nothing was applied, so a retry works. On the git-ref path step1 (`--force-reinstall --no-deps`) can succeed while the dep sync fails, leaving the venv on the new version with the daemon on the old one; `/update/start-from-ref` has no availability guard so a retry still works there.
- apps_venv failure is non-fatal by design: the daemon venv is already on the new version at that point, so skipping the restart would leave it installed-but-not-running with retries refused as "no update available"; `check_and_sync_apps_venv_sdk` re-syncs on the next boot. Catches `Exception`, not just `RuntimeError`, so an `OSError` out of `create_subprocess_shell` doesn't skip the restart either.

Behavioural pairing (fail-open, any merge order): the JS block's daemon-update gate pairs with this - it's what makes the gate's failure path honest instead of looping on a silently-failed install.

Tests: `test_wireless_version.py` (33 passed locally, requires the `wireless-version` extra).

### DIFF
--- a/src/reachy_mini/utils/wireless_version/update.py
+++ b/src/reachy_mini/utils/wireless_version/update.py
@@ -57,10 +57,14 @@ async def update_reachy_mini(
     else:
         logger.info("apps_venv not found, skipping")
 
-    # Restart daemon to apply updates. --no-block: a blocking restart SIGTERMs
-    # our own cgroup (the systemctl child included), so it can exit 143 on a
-    # SUCCESSFUL update; --no-block returns once the job is queued, so a
-    # non-zero exit is a real failure.
+    # Restart daemon to apply updates. Two constraints, both load-bearing:
+    # - the argv must match /etc/sudoers.d/010-pollen-reachy exactly (any extra
+    #   flag, --no-block included, makes sudo demand a password and fail);
+    # - systemd SIGTERMs our own cgroup, killing this systemctl child, so
+    #   143/-15 means "the restart we asked for is happening", not a failure.
+    # Everything else (unknown unit, sudo refusal) still raises.
     await call_logger_wrapper(
-        "sudo systemctl --no-block restart reachy-mini-daemon", logger
+        "sudo systemctl restart reachy-mini-daemon",
+        logger,
+        ok_returncodes=(0, -15, 143),
     )
--- a/src/reachy_mini/utils/wireless_version/update.py
+++ b/src/reachy_mini/utils/wireless_version/update.py
@@ -18,13 +18,22 @@ async def update_reachy_mini(
         pre_release: If True, install pre-release from PyPI (ignored if git_ref set).
         git_ref: If set, install from this GitHub tag/branch instead of PyPI.
 
+    Raises:
+        RuntimeError: If the daemon venv install (or the final restart
+            command) fails. The daemon is then NOT restarted, nothing has
+            been applied, and the job is reported as failed - a retry stays
+            possible. An apps_venv failure is deliberately non-fatal (see
+            below).
+
     """
-    # Update daemon venv
+    # Update daemon venv. A failure here aborts the whole update BEFORE the
+    # restart: the running daemon is untouched and the failed install left
+    # the venv on the previous version, so a retry works.
     logger.info("Updating daemon venv...")
     cmd, extra_env = build_install_command(
         extras="wireless-version",
-        git_ref=git_ref, 
-        pre_release=pre_release, 
+        git_ref=git_ref,
+        pre_release=pre_release,
         upgrade=True,
     )
     await call_logger_wrapper(cmd, logger, env=extra_env or None)
@@ -33,15 +42,26 @@ async def update_reachy_mini(
     apps_venv_python = Path("/venvs/apps_venv/bin/python")
     if apps_venv_python.exists():
         logger.info("Updating apps_venv SDK...")
-        cmd, extra_env = build_install_command( 
+        cmd, extra_env = build_install_command(
             extras="",
-            git_ref=git_ref, 
+            git_ref=git_ref,
             pre_release=pre_release,
-            python=apps_venv_python, 
+            python=apps_venv_python,
             upgrade=True,
         )
-        await call_logger_wrapper(cmd, logger, env=extra_env or None)
-        logger.info("Apps venv SDK updated successfully")
+        try:
+            await call_logger_wrapper(cmd, logger, env=extra_env or None)
+            logger.info("Apps venv SDK updated successfully")
+        except RuntimeError as e:
+            # Non-fatal by design. The daemon venv is already on the new
+            # version at this point: aborting without restarting would leave
+            # it installed-but-not-running, and a retry would be refused as
+            # "no update available" (the availability check reads installed
+            # metadata). Restart anyway; `check_and_sync_apps_venv_sdk`
+            # re-syncs apps_venv to the daemon's version on every boot.
+            logger.error(
+                f"apps_venv SDK update failed (will re-sync on next boot): {e}"
+            )
     else:
         logger.info("apps_venv not found, skipping")
 
--- a/src/reachy_mini/utils/wireless_version/update.py
+++ b/src/reachy_mini/utils/wireless_version/update.py
@@ -19,16 +19,12 @@ async def update_reachy_mini(
         git_ref: If set, install from this GitHub tag/branch instead of PyPI.
 
     Raises:
-        RuntimeError: If the daemon venv install (or the final restart
-            command) fails. The daemon is then NOT restarted, nothing has
-            been applied, and the job is reported as failed - a retry stays
-            possible. An apps_venv failure is deliberately non-fatal (see
-            below).
+        RuntimeError: If the daemon venv install or the restart command
+            fails. An apps_venv failure is deliberately non-fatal.
 
     """
-    # Update daemon venv. A failure here aborts the whole update BEFORE the
-    # restart: the running daemon is untouched and the failed install left
-    # the venv on the previous version, so a retry works.
+    # Update daemon venv. Fatal: abort before the restart. On the PyPI path
+    # the venv is left on the previous version, so a retry works.
     logger.info("Updating daemon venv...")
     cmd, extra_env = build_install_command(
         extras="wireless-version",
@@ -52,18 +48,19 @@ async def update_reachy_mini(
         try:
             await call_logger_wrapper(cmd, logger, env=extra_env or None)
             logger.info("Apps venv SDK updated successfully")
-        except RuntimeError as e:
-            # Non-fatal by design. The daemon venv is already on the new
-            # version at this point: aborting without restarting would leave
-            # it installed-but-not-running, and a retry would be refused as
-            # "no update available" (the availability check reads installed
-            # metadata). Restart anyway; `check_and_sync_apps_venv_sdk`
-            # re-syncs apps_venv to the daemon's version on every boot.
+        except Exception as e:
+            # Non-fatal: the daemon venv is already on the new version, and
+            # check_and_sync_apps_venv_sdk re-syncs apps_venv on the next boot.
             logger.error(
                 f"apps_venv SDK update failed (will re-sync on next boot): {e}"
             )
     else:
         logger.info("apps_venv not found, skipping")
 
-    # Restart daemon to apply updates
-    await call_logger_wrapper("sudo systemctl restart reachy-mini-daemon", logger)
+    # Restart daemon to apply updates. --no-block: a blocking restart SIGTERMs
+    # our own cgroup (the systemctl child included), so it can exit 143 on a
+    # SUCCESSFUL update; --no-block returns once the job is queued, so a
+    # non-zero exit is a real failure.
+    await call_logger_wrapper(
+        "sudo systemctl --no-block restart reachy-mini-daemon", logger
+    )
--- a/src/reachy_mini/utils/wireless_version/utils.py
+++ b/src/reachy_mini/utils/wireless_version/utils.py
@@ -90,10 +90,7 @@ def build_install_command(
         if not _check_uv_available():
             step3_args += ["--upgrade-strategy", "only-if-needed"]
         step3 = shlex.join(step3_args)
-        # Group steps 2-3 explicitly: with the bare `step1 && step2 || step3`
-        # shell precedence, a FAILED step1 would run step3 (a plain PyPI
-        # upgrade), silently replacing the requested git ref with the latest
-        # wheel and masking the failure behind step3's exit code.
+        # Grouped: a failed step1 must not fall through to step3.
         cmd = f"{step1} && ( {step2} || {step3} )"
         logger.info(f"Git ref install: {cmd}")
         extra_env: dict[str, str] = {"GIT_LFS_SKIP_SMUDGE": "1"}
@@ -130,10 +127,7 @@ async def call_logger_wrapper(
         env: Optional environment variables dict. If None, inherits current environment.
 
     Raises:
-        RuntimeError: If the command exits with a non-zero code, so callers
-            (the update flow) can surface a real failure instead of carrying
-            on - e.g. restarting the daemon after a failed install and
-            reporting the job as done.
+        RuntimeError: If the command exits with a non-zero code.
 
     """
     logger.info(f"Running: {command}")
@@ -164,6 +158,4 @@ async def call_logger_wrapper(
     await asyncio.gather(*tasks)
     returncode = await process.wait()
     if returncode != 0:
-        message = f"Command failed with exit code {returncode}: {command}"
-        logger.error(message)
-        raise RuntimeError(message)
+        raise RuntimeError(f"Command failed with exit code {returncode}: {command}")
--- a/src/reachy_mini/utils/wireless_version/utils.py
+++ b/src/reachy_mini/utils/wireless_version/utils.py
@@ -118,6 +118,7 @@ async def call_logger_wrapper(
     command: str,
     logger: logging.Logger,
     env: dict[str, str] | None = None,
+    ok_returncodes: tuple[int, ...] = (0,),
 ) -> None:
     """Run a shell command asynchronously, streaming stdout and stderr to logger in real time.
 
@@ -125,9 +126,11 @@ async def call_logger_wrapper(
         command: Shell command string.
         logger: logger object with .info and .error methods
         env: Optional environment variables dict. If None, inherits current environment.
+        ok_returncodes: Exit codes treated as success. Only the daemon restart
+            needs more than ``(0,)`` - see ``update_reachy_mini``.
 
     Raises:
-        RuntimeError: If the command exits with a non-zero code.
+        RuntimeError: If the command exits with a code outside *ok_returncodes*.
 
     """
     logger.info(f"Running: {command}")
@@ -157,5 +160,5 @@ async def call_logger_wrapper(
 
     await asyncio.gather(*tasks)
     returncode = await process.wait()
-    if returncode != 0:
+    if returncode not in ok_returncodes:
         raise RuntimeError(f"Command failed with exit code {returncode}: {command}")
--- a/src/reachy_mini/utils/wireless_version/utils.py
+++ b/src/reachy_mini/utils/wireless_version/utils.py
@@ -79,7 +79,9 @@ def build_install_command(
         git_url = f"git+https://github.com/{GITHUB_REPO}.git@{git_ref}"
         git_package = f"reachy-mini[{extras}] @ {git_url}"
         # Step 1: force reinstall the package without the dependencies
-        step1 = shlex.join(base + [git_package, "--force-reinstall", "--no-deps", "--no-cache-dir"])
+        step1 = shlex.join(
+            base + [git_package, "--force-reinstall", "--no-deps", "--no-cache-dir"]
+        )
         # Step 2: check if dependencies need to be updated
         check_base = [arg if arg != "install" else "check" for arg in base]
         step2 = shlex.join(check_base)
@@ -88,12 +90,18 @@ def build_install_command(
         if not _check_uv_available():
             step3_args += ["--upgrade-strategy", "only-if-needed"]
         step3 = shlex.join(step3_args)
-        cmd = f"{step1} && {step2} || {step3}"
+        # Group steps 2-3 explicitly: with the bare `step1 && step2 || step3`
+        # shell precedence, a FAILED step1 would run step3 (a plain PyPI
+        # upgrade), silently replacing the requested git ref with the latest
+        # wheel and masking the failure behind step3's exit code.
+        cmd = f"{step1} && ( {step2} || {step3} )"
         logger.info(f"Git ref install: {cmd}")
         extra_env: dict[str, str] = {"GIT_LFS_SKIP_SMUDGE": "1"}
         return cmd, extra_env
 
-    logger.info(f"Installing from PyPI: {version if version else 'latest pre-release' if pre_release else 'latest stable'}")
+    logger.info(
+        f"Installing from PyPI: {version if version else 'latest pre-release' if pre_release else 'latest stable'}"
+    )
     package = f"reachy-mini[{extras}]"
     if version:
         package = f"{package}=={version}"
@@ -120,6 +128,12 @@ async def call_logger_wrapper(
         command: Shell command string.
         logger: logger object with .info and .error methods
         env: Optional environment variables dict. If None, inherits current environment.
+
+    Raises:
+        RuntimeError: If the command exits with a non-zero code, so callers
+            (the update flow) can surface a real failure instead of carrying
+            on - e.g. restarting the daemon after a failed install and
+            reporting the job as done.
 
     """
     logger.info(f"Running: {command}")
@@ -148,4 +162,8 @@ async def call_logger_wrapper(
         tasks.append(asyncio.create_task(stream_output(process.stderr, logger.error)))
 
     await asyncio.gather(*tasks)
-    await process.wait()
+    returncode = await process.wait()
+    if returncode != 0:
+        message = f"Command failed with exit code {returncode}: {command}"
+        logger.error(message)
+        raise RuntimeError(message)
--- a/tests/unit_tests/test_wireless_version.py
+++ b/tests/unit_tests/test_wireless_version.py
@@ -122,6 +122,29 @@ async def test_call_logger_wrapper_nonzero_raises():
         await utils.call_logger_wrapper("exit 7", logging.getLogger("test"))
 
 
+@pytest.mark.asyncio
+async def test_call_logger_wrapper_tolerated_returncode():
+    """A code listed in ok_returncodes does not raise (the restart's 143/-15)."""
+    # 143 = exit status relayed by the shell.
+    await utils.call_logger_wrapper("exit 143", logging.getLogger("test"), ok_returncodes=(0, -15, 143))
+    # -15 = process killed by SIGTERM (asyncio reports the negated signal), the
+    # real self-inflicted signal seen inside the daemon's cgroup.
+    await utils.call_logger_wrapper("kill -TERM $$", logging.getLogger("test"), ok_returncodes=(0, -15, 143))
+    # A code outside the list still raises.
+    with pytest.raises(RuntimeError, match="exit code 5"):
+        await utils.call_logger_wrapper("exit 5", logging.getLogger("test"), ok_returncodes=(0, -15, 143))
+
+
+def test_restart_argv_matches_sudoers_whitelist():
+    """The restart argv is whitelisted verbatim in /etc/sudoers.d/010-pollen-reachy.
+
+    Any extra flag (e.g. --no-block) stops sudo matching the NOPASSWD rule, so the
+    restart fails with "a password is required" and the update never applies.
+    """
+    src = Path(update.__file__).read_text()
+    assert '"sudo systemctl restart reachy-mini-daemon"' in src
+
+
 # --- _semver_version ---
 
 
@@ -317,7 +340,7 @@ async def test_update_with_apps_venv(monkeypatch):
     await update.update_reachy_mini(logging.getLogger("test"))
 
     assert calls.call_count == 3
-    assert calls.await_args_list[-1].args[0] == "sudo systemctl --no-block restart reachy-mini-daemon"
+    assert calls.await_args_list[-1].args[0] == "sudo systemctl restart reachy-mini-daemon"
 
 
 @pytest.mark.asyncio
@@ -330,7 +353,7 @@ async def test_update_without_apps_venv(monkeypatch):
     await update.update_reachy_mini(logging.getLogger("test"), pre_release=True)
 
     assert calls.call_count == 2
-    assert calls.await_args_list[-1].args[0] == "sudo systemctl --no-block restart reachy-mini-daemon"
+    assert calls.await_args_list[-1].args[0] == "sudo systemctl restart reachy-mini-daemon"
 
 
 @pytest.mark.asyncio
@@ -352,7 +375,7 @@ async def test_update_apps_venv_failure_still_restarts(monkeypatch):
     """An apps_venv failure is non-fatal: the restart still applies the update."""
     commands: list[str] = []
 
-    async def fake_call(cmd, logger, env=None):
+    async def fake_call(cmd, logger, env=None, ok_returncodes=(0,)):
         commands.append(cmd)
         if len(commands) == 2:  # second call = apps_venv install
             raise RuntimeError("apps_venv install failed")
@@ -363,7 +386,7 @@ async def test_update_apps_venv_failure_still_restarts(monkeypatch):
     await update.update_reachy_mini(logging.getLogger("test"))
 
     assert len(commands) == 3
-    assert commands[-1] == "sudo systemctl --no-block restart reachy-mini-daemon"
+    assert commands[-1] == "sudo systemctl restart reachy-mini-daemon"
 
 
 @pytest.mark.asyncio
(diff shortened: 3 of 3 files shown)


src/reachy_mini/utils/wireless_version/utils.py
```diff
@@ -148,4 +162,8 @@ async def stream_output(
         tasks.append(asyncio.create_task(stream_output(process.stderr, logger.error)))
 
     await asyncio.gather(*tasks)
-    await process.wait()
+    returncode = await process.wait()
+    if returncode != 0:
+        message = f"Command failed with exit code {returncode}: {command}"
+        logger.error(message)
+        raise RuntimeError(message)
```
FabienDanieau: **Blocking.** This is what makes `update.py:69`'s `sudo systemctl restart reachy-mini-daemon` fatal, and that command can exit non-zero on the *success* path: systemd SIGTERMs the whole cgroup, `sudo systemctl` included, so 143 is a normal outcome here. Since uvicorn's SIGTERM handler makes shutdown graceful rather than instant, the update thread has time to raise and broadcast `update_progress status="failed"` for an update that actually worked.

Two ways out: `systemctl --no-block restart` (returns before the kill, so the exit code means something again), or tolerate `143`/`-15` for this one command. `test_update_restart_failure_propagates` injects a synthetic `RuntimeError`, so it can't catch this.

_(ponytail)_ Minor, while here: all three call sites already log the exception (`update.py:63`, `bg_job_register.py:84`, `daemon.py:768`), so the local `logger.error` is duplicate output — `raise RuntimeError(f"Command failed with exit code {returncode}: {command}")` is the whole thing.

src/reachy_mini/utils/wireless_version/utils.py
```diff
@@ -88,12 +90,18 @@ def build_install_command(
         if not _check_uv_available():
             step3_args += ["--upgrade-strategy", "only-if-needed"]
         step3 = shlex.join(step3_args)
-        cmd = f"{step1} && {step2} || {step3}"
+        # Group steps 2-3 explicitly: with the bare `step1 && step2 || step3`
+        # shell precedence, a FAILED step1 would run step3 (a plain PyPI
+        # upgrade), silently replacing the requested git ref with the latest
+        # wheel and masking the failure behind step3's exit code.
+        cmd = f"{step1} && ( {step2} || {step3} )"
```
FabienDanieau: Correct fix. Note it also repairs `check_and_sync_apps_venv_sdk` (`startup_check.py:332`), which builds this same chain and checks `returncode` — worth a line in the PR body.

_(ponytail)_ 4 lines of comment for one pair of parens. `# grouped: a failed step1 must not fall through to step3` carries it.

src/reachy_mini/utils/wireless_version/utils.py
```diff
@@ -88,12 +90,18 @@ def build_install_command(
...
         logger.info(f"Git ref install: {cmd}")
         extra_env: dict[str, str] = {"GIT_LFS_SKIP_SMUDGE": "1"}
         return cmd, extra_env
 
-    logger.info(f"Installing from PyPI: {version if version else 'latest pre-release' if pre_release else 'latest stable'}")
+    logger.info(
+        f"Installing from PyPI: {version if version else 'latest pre-release' if pre_release else 'latest stable'}"
+    )
```
FabienDanieau, line 101: _(ponytail)_ Unrelated reformat of an untouched line. This repo isn't format-clean, so reformatting lines the fix doesn't need just adds diff to read.

src/reachy_mini/utils/wireless_version/utils.py
```diff
@@ -121,6 +129,12 @@ async def call_logger_wrapper(
         logger: logger object with .info and .error methods
         env: Optional environment variables dict. If None, inherits current environment.
 
+    Raises:
+        RuntimeError: If the command exits with a non-zero code, so callers
+            (the update flow) can surface a real failure instead of carrying
+            on - e.g. restarting the daemon after a failed install and
+            reporting the job as done.
```
FabienDanieau: _(ponytail)_ 5 lines explaining *why callers care*, in the callee's docstring. `RuntimeError: If the command exits with a non-zero code.` — the motivation is already in the PR description, where it's discoverable.

src/reachy_mini/utils/wireless_version/update.py
```diff
@@ -18,13 +18,22 @@ async def update_reachy_mini(
...
+            command) fails. The daemon is then NOT restarted, nothing has
+            been applied, and the job is reported as failed - a retry stays
+            possible. An apps_venv failure is deliberately non-fatal (see
+            below).
+
     """
-    # Update daemon venv
+    # Update daemon venv. A failure here aborts the whole update BEFORE the
+    # restart: the running daemon is untouched and the failed install left
+    # the venv on the previous version, so a retry works.
```
FabienDanieau: "the failed install left the venv on the previous version, so a retry works" doesn't hold for the git-ref path: `step1` (`--force-reinstall --no-deps`) can succeed while `step2 || step3` fails, leaving the venv on the **new** version with the daemon still on the old one.

`/update/start-from-ref` has no availability guard so a retry there is fine, but `/update/start` would then refuse with "No update available" — exactly the installed-but-not-running trap the `apps_venv` comment below argues against. Worth scoping the claim to the PyPI path.

src/reachy_mini/utils/wireless_version/update.py
```diff
@@ -18,13 +18,22 @@ async def update_reachy_mini(
         pre_release: If True, install pre-release from PyPI (ignored if git_ref set).
         git_ref: If set, install from this GitHub tag/branch instead of PyPI.
 
+    Raises:
+        RuntimeError: If the daemon venv install (or the final restart
+            command) fails. The daemon is then NOT restarted, nothing has
+            been applied, and the job is reported as failed - a retry stays
+            possible. An apps_venv failure is deliberately non-fatal (see
+            below).
```
FabienDanieau: _(ponytail)_ 6 lines restating the policy that the two inline comments below already state. Two lines is enough here.

src/reachy_mini/utils/wireless_version/update.py
```diff
@@ -33,15 +42,26 @@ async def update_reachy_mini(
...
+        except RuntimeError as e:
+            # Non-fatal by design. The daemon venv is already on the new
+            # version at this point: aborting without restarting would leave
+            # it installed-but-not-running, and a retry would be refused as
+            # "no update available" (the availability check reads installed
+            # metadata). Restart anyway; `check_and_sync_apps_venv_sdk`
+            # re-syncs apps_venv to the daemon's version on every boot.
+            logger.error(
+                f"apps_venv SDK update failed (will re-sync on next boot): {e}"
+            )
```
FabienDanieau: If this step is non-fatal by design, catch `Exception`. As written, anything that isn't `RuntimeError` — e.g. an `OSError` out of `create_subprocess_shell` — skips the restart, which is the precise outcome the comment argues against.

_(ponytail)_ 8 lines of comment on a 2-line `except`. `# non-fatal: daemon venv is already new; check_and_sync_apps_venv_sdk re-syncs on boot` says it.

tests/unit_tests/test_wireless_version.py
```diff
@@ -85,13 +92,51 @@ def test_build_git_ref_pip_fallback(monkeypatch):
...
+
+    With the bare `step1 && step2 || step3` shell precedence, a failed
+    step1 ran step3 (a plain PyPI upgrade): the latest wheel silently
+    replaced the requested git ref AND its exit code masked the failure.
+    The grouped form `step1 && ( step2 || step3 )` short-circuits instead.
+    """
+    monkeypatch.setattr(utils, "_check_uv_available", lambda: True)
+    cmd, _ = utils.build_install_command("wireless-version", git_ref="main")
+    step1, _, _ = _split_git_ref_cmd(cmd)
+    assert cmd.startswith(f"{step1} && ( ")
```
FabienDanieau: This can't fail for the reason its docstring gives. `_split_git_ref_cmd` already asserts `rest.startswith("( ") and rest.endswith(" )")` on every call, so `assert cmd.startswith(f"{step1} && ( ")` re-checks a shape that's checked in both git-ref tests — and no shell ever runs, so precedence is never exercised.

Either delete it (the grouping is covered), or make it real:

```python
assert subprocess.run("false && true || true", shell=True).returncode == 0       # old, masked
assert subprocess.run("false && ( true || true )", shell=True).returncode != 0   # new, honest
```

tests/unit_tests/test_wireless_version.py
```diff
@@ -85,13 +92,51 @@ def test_build_git_ref_pip_fallback(monkeypatch):
...
+    """Regression: a non-zero exit code must raise, not pass silently.
+
+    Before this check, a failed `pip install` sailed through and
+    `update_reachy_mini` restarted the daemon on the OLD version while
+    every route (REST job, BLE screen, WebRTC broadcast) reported success.
+    """
+    import logging
+
+    with pytest.raises(RuntimeError, match="exit code 7"):
+        await utils.call_logger_wrapper("exit 7", logging.getLogger("test"))
```
FabienDanieau, line 122: _(ponytail)_ These two are the right size of check — shelling out to `true` / `exit 7` is exactly it. Two nits: the 6-line docstring narrating the old bug belongs in the PR, and `import logging` appears inside five test bodies in this file now — one module-level import next to `shlex` / `Path` covers them.

tests/unit_tests/test_wireless_version.py
```diff
@@ -305,3 +350,65 @@ async def test_update_without_apps_venv(monkeypatch):
...
     assert calls.call_count == 2
     assert calls.await_args_list[-1].args[0] == "sudo systemctl restart reachy-mini-daemon"
+
+
+@pytest.mark.asyncio
+async def test_update_daemon_install_failure_aborts_before_restart(monkeypatch):
+    """A failed daemon venv install propagates and never restarts the daemon."""
+    calls = AsyncMock(side_effect=RuntimeError("pip install failed"))
+    monkeypatch.setattr(update, "call_logger_wrapper", calls)
+    monkeypatch.setattr(update.Path, "exists", lambda self: True)
```
FabienDanieau, line 364: Worth knowing this patches `pathlib.Path.exists` process-wide for the test, so *every* path exists, not just `/venvs/apps_venv/bin/python`. Matches the existing pattern in this file so I wouldn't change it, but it's load-bearing in a way that isn't visible.

tests/unit_tests/test_wireless_version.py
```diff
@@ -305,3 +350,65 @@ async def test_update_without_apps_venv(monkeypatch):
...
+
+@pytest.mark.asyncio
+async def test_update_apps_venv_failure_still_restarts(monkeypatch):
+    """An apps_venv failure is non-fatal: the restart still applies the update.
+
+    The daemon venv is already on the new version when apps_venv fails;
+    skipping the restart would leave it installed-but-not-running with
+    retries refused as "no update available". The startup check re-syncs
+    apps_venv on the next boot instead.
+    """
```
FabienDanieau: _(ponytail)_ Same as above — the 7-line rationale duplicates the comment at `update.py:56-61` and the PR body. The one-liner in `test_update_restart_failure_propagates` is the right length.

FabienDanieau (changes requested):
Good catch on both bugs, and the failure policy (daemon venv fatal → apps_venv non-fatal → restart fatal) is the right one. I checked the blast radius: `call_logger_wrapper` has a single caller module, and `bg_job_register.run_command` already wraps the coroutine in `try/except Exception` → `JobStatus.FAILED`, so the new exception surfaces correctly on the REST path; `daemon.py:761` likewise broadcasts `failed` and releases `busy_lock` in `finally`. No unhandled propagation, and no injection risk from the `( )` grouping since all three steps come out of `shlex.join`.

One thing blocks, the rest are notes.

**Blocking — the restart is now fatal, and it can report a *successful* update as failed.** `sudo systemctl restart reachy-mini-daemon` restarts the unit that owns the calling process. With systemd's default `KillMode=control-group`, SIGTERM hits the whole cgroup — including the `sudo systemctl` child — so `process.wait()` can legitimately return 143. The daemon installs no SIGTERM handler of its own, but it runs `uvicorn.Server` in-process (`daemon/app/main.py:575`) and uvicorn *does*, which leaves a graceful-shutdown window where the update thread observes 143, raises, and `_broadcast("failed", error=...)` reaches clients whose datachannel isn't torn down yet. Previously that exit code was harmless. Details inline.

Also worth adding to the PR body: the grouping fix has a second beneficiary you don't mention. `check_and_sync_apps_venv_sdk` (`startup_check.py:332`) builds the same git-ref chain and gates on `returncode` — it used to print "Successfully synced apps_venv SDK" after quietly installing the latest PyPI wheel instead of the requested ref. That's arguably the strongest argument for the change.

Separately, and this is a style call rather than a defect: the two source fixes are 1 line of logic each, carrying ~30 lines of comment and `Raises:` prose between them, plus test docstrings that narrate the old bug at length. Roughly 4:1 prose-to-code. Most of that archaeology belongs in this PR description, which already says it well — the comments marked _(ponytail)_ below are where I'd trim (~48 lines).

src/reachy_mini/utils/wireless_version/utils.py
tfrere: good catch, went with --no-block (fc68f02). systemctl now returns once the restart job is queued, so a non-zero exit means the queueing itself failed. also dropped the local logger.error, the raise alone carries it.

src/reachy_mini/utils/wireless_version/update.py
tfrere: right, changed to except Exception

tests/unit_tests/test_wireless_version.py
tfrere: replaced it with your subprocess version, it actually exercises the precedence now

src/reachy_mini/utils/wireless_version/utils.py
tfrere: good point, added it to the PR body

src/reachy_mini/utils/wireless_version/utils.py
tfrere: tried to revert it but the pre-commit ruff-format hook re-wraps any touched file, so this one has to stay

src/reachy_mini/utils/wireless_version/update.py
tfrere: fair, scoped the comment to the PyPI path

src/reachy_mini/utils/wireless_version/utils.py
tfrere: trimmed to the one-liner

src/reachy_mini/utils/wireless_version/update.py
tfrere: trimmed

tests/unit_tests/test_wireless_version.py
tfrere: docstring trimmed and import logging moved to module level

tests/unit_tests/test_wireless_version.py
tfrere: trimmed to a one-liner too

tests/unit_tests/test_wireless_version.py
tfrere: yeah, kept it as is to match the existing pattern in the file

FabienDanieau (commented):
I tested `fc68f02` on a real robot (`reachy-mini.local`, branch checked out at `3f19275`). **`--no-block` does not fix the false failure, and it breaks the restart entirely.** That's on me — `--no-block` was my suggestion in the previous review, and it's wrong for two independent reasons. Suggested changes are inline; the measurements are below so you can re-run them.

### 1. sudoers is argv-exact, so the new command cannot run at all

`/etc/sudoers.d/010-pollen-reachy` whitelists the literal argv:

```
Cmnd_Alias REACHY = \
    /usr/bin/systemctl restart reachy-mini-daemon, \
```

Adding any flag breaks the NOPASSWD match. Run exactly as `create_subprocess_shell` does, as `pollen`, with no tty:

| command | rc | daemon after |
|---|---|---|
| `sudo systemctl --no-block restart reachy-mini-daemon` (this PR) | **1** — `sudo: a terminal is required to read the password` | **inactive** |
| `sudo systemctl restart reachy-mini-daemon` (before) | **0** | **active** |

Strictness confirmed independently: `sudo -n rfkill unblock bluetooth` (exact match) succeeds, `sudo -n rfkill --json unblock bluetooth` → `a password is required`.

So on this branch the update installs the new version, the restart fails, and — because of this PR's own `raise` — it propagates: job reported FAILED, daemon never restarts, robot left installed-but-not-running with retries refused as "no update available". That's the trap your own `apps_venv` comment warns about, now on the primary path. Net effect vs `main`: before, the restart always worked and the report sometimes lied; now the report is honest and updates don't apply.

### 2. Even with sudoers fixed, `--no-block` still loses the race

Isolated transient unit, `KillMode=control-group` (same as the real unit — verified via `systemctl show`), update on a worker thread with a uvicorn-like graceful SIGTERM handler on the main thread:

| variant | rc | elapsed |
|---|---|---|
| `systemctl restart <self>` — blocking, in-cgroup | **-15** | 0.04s |
| `systemctl --no-block restart <self>` — in-cgroup | **-15** | 0.04s |
| control: `--no-block` from **outside** the cgroup | **0** | — |
| `systemd-run --collect --unit=… systemctl restart <self>` | **0** | 0.05s |

`--no-block` returns fast, but the `systemctl` child sits in the cgroup being SIGTERMed and the kill lands first. Escaping the cgroup does work — but `systemd-run` needs a **new sudoers entry**, which lives in the OS image, so it's a cross-repo change with a deployment-ordering problem (SDK updates would break on robots still carrying the old sudoers). Not viable here.

### The fix, verified end-to-end

Keep the whitelisted argv, tolerate only the self-inflicted signal. Applied on the robot and run from **inside the real daemon's cgroup** (`0::/system.slice/reachy-mini-daemon.service`), as uid 1000, no tty:

| | observed | daemon restarted? | job reports |
|---|---|---|---|
| `ok_returncodes=(0,)` | `RAISED … exit code -15` | ✅ yes | **FAILED** ← the original bug, reproduced |
| `ok_returncodes=(0, -15, 143)` | returned cleanly | ✅ yes | **DONE** |

Both `-15` and `143` because whether `sh` dies or relays its child's status decides which you see; I observed `-15` in every run on this robot, `143` is belt-and-braces.

`33 passed` with the suggestions applied (aarch64, py3.12.13, `--extra wireless-version`), and `ruff format --check` + `ruff check --select I --select D` clean at the pinned 0.12.0.

**One piece I can't attach as a suggestion**: line 120 of `utils.py` is outside the diff, so GitHub won't let me anchor there. Add the parameter to the signature by hand:

```python
 async def call_logger_wrapper(
     command: str,
     logger: logging.Logger,
     env: dict[str, str] | None = None,
+    ok_returncodes: tuple[int, ...] = (0,),
 ) -> None:
```

Everything else in `fc68f02` is good — the trimmed comments, `except Exception`, the real shell-precedence test all stand.

src/reachy_mini/utils/wireless_version/update.py
```diff
@@ -33,17 +38,29 @@ async def update_reachy_mini(
...
 
-    # Restart daemon to apply updates
-    await call_logger_wrapper("sudo systemctl restart reachy-mini-daemon", logger)
+    # Restart daemon to apply updates. --no-block: a blocking restart SIGTERMs
+    # our own cgroup (the systemctl child included), so it can exit 143 on a
+    # SUCCESSFUL update; --no-block returns once the job is queued, so a
+    # non-zero exit is a real failure.
+    await call_logger_wrapper(
+        "sudo systemctl --no-block restart reachy-mini-daemon", logger
+    )
```
FabienDanieau: This is the change that breaks it: `--no-block` makes the argv stop matching the sudoers NOPASSWD rule, so `sudo` demands a password, gets no tty, and the daemon never restarts. And per the table in my summary, `--no-block` wouldn't have avoided the `-15` anyway — the `systemctl` child is inside the cgroup being killed either way.

Measured working version:

```suggestion
    # Restart daemon to apply updates. Two constraints, both load-bearing:
    # - the argv must match /etc/sudoers.d/010-pollen-reachy exactly (any extra
    #   flag, --no-block included, makes sudo demand a password and fail);
    # - systemd SIGTERMs our own cgroup, killing this systemctl child, so
    #   143/-15 means "the restart we asked for is happening", not a failure.
    # Everything else (unknown unit, sudo refusal) still raises.
    await call_logger_wrapper(
        "sudo systemctl restart reachy-mini-daemon",
        logger,
        ok_returncodes=(0, -15, 143),
    )
```

src/reachy_mini/utils/wireless_version/utils.py
```diff
@@ -121,6 +126,9 @@ async def call_logger_wrapper(
         logger: logger object with .info and .error methods
         env: Optional environment variables dict. If None, inherits current environment.
 
+    Raises:
+        RuntimeError: If the command exits with a non-zero code.
```
FabienDanieau: Docstring for the new parameter (the signature line itself is outside the diff — see the summary for that one-line hand edit).

```suggestion
        logger: logger object with .info and .error methods
        env: Optional environment variables dict. If None, inherits current environment.
        ok_returncodes: Exit codes treated as success. Only the daemon restart
            needs more than ``(0,)`` - see ``update_reachy_mini``.

    Raises:
        RuntimeError: If the command exits with a code outside *ok_returncodes*.
```

src/reachy_mini/utils/wireless_version/utils.py
```diff
@@ -148,4 +156,6 @@ async def stream_output(
         tasks.append(asyncio.create_task(stream_output(process.stderr, logger.error)))
 
     await asyncio.gather(*tasks)
-    await process.wait()
+    returncode = await process.wait()
+    if returncode != 0:
+        raise RuntimeError(f"Command failed with exit code {returncode}: {command}")
```
FabienDanieau:
```suggestion
    if returncode not in ok_returncodes:
        raise RuntimeError(f"Command failed with exit code {returncode}: {command}")
```

Default stays `(0,)`, so every other call site keeps the strict behaviour this PR is adding.

tests/unit_tests/test_wireless_version.py
```diff
@@ -85,13 +94,34 @@ def test_build_git_ref_pip_fallback(monkeypatch):
...
+async def test_call_logger_wrapper_success():
+    """A zero exit code returns without raising."""
+    await utils.call_logger_wrapper("true", logging.getLogger("test"))
+
+
+@pytest.mark.asyncio
+async def test_call_logger_wrapper_nonzero_raises():
+    """Regression: a non-zero exit code must raise, not pass silently."""
+    with pytest.raises(RuntimeError, match="exit code 7"):
+        await utils.call_logger_wrapper("exit 7", logging.getLogger("test"))
```
FabienDanieau, line 122: Two tests worth adding here — the second one is what would have caught this: it pins the restart argv so a flag can't creep back in without a failing test.

```suggestion
@pytest.mark.asyncio
async def test_call_logger_wrapper_nonzero_raises():
    """Regression: a non-zero exit code must raise, not pass silently."""
    with pytest.raises(RuntimeError, match="exit code 7"):
        await utils.call_logger_wrapper("exit 7", logging.getLogger("test"))


@pytest.mark.asyncio
async def test_call_logger_wrapper_tolerated_returncode():
    """A code listed in ok_returncodes does not raise (the restart's 143/-15)."""
    await utils.call_logger_wrapper(
        "exit 143", logging.getLogger("test"), ok_returncodes=(0, -15, 143)
    )
    with pytest.raises(RuntimeError, match="exit code 5"):
        await utils.call_logger_wrapper(
            "exit 5", logging.getLogger("test"), ok_returncodes=(0, -15, 143)
        )


def test_restart_argv_matches_sudoers_whitelist():
    """The restart argv is whitelisted verbatim in /etc/sudoers.d/010-pollen-reachy.

    Any extra flag (e.g. --no-block) stops sudo matching the NOPASSWD rule, so the
    restart fails with "a password is required" and the update never applies.
    """
    src = Path(update.__file__).read_text()
    assert '"sudo systemctl restart reachy-mini-daemon"' in src
```

tests/unit_tests/test_wireless_version.py
```diff
@@ -284,12 +314,10 @@ async def test_update_with_apps_venv(monkeypatch):
     monkeypatch.setattr(update, "call_logger_wrapper", calls)
     monkeypatch.setattr(update.Path, "exists", lambda self: True)
 
-    import logging
-
     await update.update_reachy_mini(logging.getLogger("test"))
 
     assert calls.call_count == 3
-    assert calls.await_args_list[-1].args[0] == "sudo systemctl restart reachy-mini-daemon"
+    assert calls.await_args_list[-1].args[0] == "sudo systemctl --no-block restart reachy-mini-daemon"
```
FabienDanieau:
```suggestion
    assert calls.await_args_list[-1].args[0] == "sudo systemctl restart reachy-mini-daemon"
```

tests/unit_tests/test_wireless_version.py
```diff
@@ -299,9 +327,57 @@ async def test_update_without_apps_venv(monkeypatch):
     monkeypatch.setattr(update, "call_logger_wrapper", calls)
     monkeypatch.setattr(update.Path, "exists", lambda self: False)
 
-    import logging
-
     await update.update_reachy_mini(logging.getLogger("test"), pre_release=True)
 
     assert calls.call_count == 2
-    assert calls.await_args_list[-1].args[0] == "sudo systemctl restart reachy-mini-daemon"
+    assert calls.await_args_list[-1].args[0] == "sudo systemctl --no-block restart reachy-mini-daemon"
```
FabienDanieau:
```suggestion
    assert calls.await_args_list[-1].args[0] == "sudo systemctl restart reachy-mini-daemon"
```

tests/unit_tests/test_wireless_version.py
```diff
@@ -299,9 +327,57 @@ async def test_update_without_apps_venv(monkeypatch):
...
+    # Only the daemon venv install ran: no apps_venv step, no restart.
+    assert calls.call_count == 1
+
+
+@pytest.mark.asyncio
+async def test_update_apps_venv_failure_still_restarts(monkeypatch):
+    """An apps_venv failure is non-fatal: the restart still applies the update."""
+    commands: list[str] = []
+
+    async def fake_call(cmd, logger, env=None):
```
FabienDanieau: The fake needs the new kwarg or `update_reachy_mini` raises `TypeError`. (This is how I caught the signature change — these two fakes failed first.)

```suggestion
    async def fake_call(cmd, logger, env=None, ok_returncodes=(0,)):
```

tests/unit_tests/test_wireless_version.py
```diff
@@ -299,9 +327,57 @@ async def test_update_without_apps_venv(monkeypatch):
...
+        if len(commands) == 2:  # second call = apps_venv install
+            raise RuntimeError("apps_venv install failed")
+
+    monkeypatch.setattr(update, "call_logger_wrapper", fake_call)
+    monkeypatch.setattr(update.Path, "exists", lambda self: True)
+
+    await update.update_reachy_mini(logging.getLogger("test"))
+
+    assert len(commands) == 3
+    assert commands[-1] == "sudo systemctl --no-block restart reachy-mini-daemon"
```
FabienDanieau:
```suggestion
    assert commands[-1] == "sudo systemctl restart reachy-mini-daemon"
```

tests/unit_tests/test_wireless_version.py
```diff
@@ -299,9 +327,57 @@ async def test_update_without_apps_venv(monkeypatch):
...
+
+
+@pytest.mark.asyncio
+async def test_update_restart_failure_propagates(monkeypatch):
+    """A failed restart command surfaces as a failed job, not a silent pass."""
+    commands: list[str] = []
+
+    async def fake_call(cmd, logger, env=None):
```
FabienDanieau:
```suggestion
    async def fake_call(cmd, logger, env=None, ok_returncodes=(0,)):
```

tfrere:
pushed the fix in b632709e and merged main to bring the branch up to date. tl;dr: dropped --no-block, back to the whitelisted argv with ok_returncodes=(0, -15, 143), exactly as you measured. 33 passed locally with the wireless-version extra, ruff clean. i still want to re-run the end-to-end update on the robot before calling this green, but the unit side pins the argv now so the flag cannot creep back silently. thanks for the on-hardware archaeology, the sudoers argv-exact bit was invisible from my side.

src/reachy_mini/utils/wireless_version/update.py
tfrere: yep, reverted --no-block and applied your version verbatim in b632709e: exact whitelisted argv + ok_returncodes=(0, -15, 143). the sudoers argv-exact match was the part i could not have caught from here, thanks for running it on the real robot.

src/reachy_mini/utils/wireless_version/utils.py
tfrere: applied. also added the ok_returncodes param to the signature by hand since line 120 was outside the diff.

src/reachy_mini/utils/wireless_version/utils.py
tfrere: done in b632709e.

tests/unit_tests/test_wireless_version.py
tfrere: added both. the tolerated-returncode test also covers a real -15 via "kill -TERM $$", not just the 143 exit, so it exercises the actual signal path. argv pin test is in too.

tests/unit_tests/test_wireless_version.py
tfrere: done.

tests/unit_tests/test_wireless_version.py
tfrere: done.

tests/unit_tests/test_wireless_version.py
tfrere: good catch, added the kwarg to both fakes.

tests/unit_tests/test_wireless_version.py
tfrere: done.

tests/unit_tests/test_wireless_version.py
tfrere: done.

FabienDanieau (approved):
Approving — the blocker is fixed and I re-verified it on hardware against the committed code, not against my patch this time.

Tested at `a9ad7495` on `reachy-mini.local` (aarch64, py3.12.13), clean tree:

- **Update suite: 33 passed.**
- **Your two new test cases hit genuinely distinct real codes**, which is the bit I wanted to measure rather than assume: `kill -TERM $$` → returncode **-15**, `exit 143` → **143**, both via `asyncio.create_subprocess_shell`. So each branch of `(0, -15, 143)` is exercised for the right reason, not because dash collapses them onto one value. Your `kill -TERM $$` addition does real work that my `exit 143`-only suggestion would have missed — good catch.
- **End-to-end on the committed `update_reachy_mini`**, with only `build_install_command` neutered to `true` (the restart step, `call_logger_wrapper` and the `ok_returncodes` wiring were all the shipped code). Run as uid 1000, no tty, from inside the real daemon cgroup `0::/system.slice/reachy-mini-daemon.service`, with a uvicorn-like graceful SIGTERM window on the main thread:

  ```
  RESULT: update_reachy_mini returned cleanly after 0.05s -> job reports DONE
  daemon ActiveEnterTimestamp: 10:46:00 -> 10:46:14   (it really did restart)
  ```

  Same harness reported `RAISED … exit code -15 -> job reports FAILED` before the fix, so this is the original bug closing, measured.

Two non-blocking things I'd still sort out before merge:

1. **The PR description is now stale and contradicts the code.** It still says the restart uses `systemctl --no-block` and that "`--no-block` returns once the job is queued, so a non-zero exit is a real failure" — that's the reverted approach. Given how detailed this body is it will likely become the squash message, so it would ship a description of behaviour the code no longer has. "31 passed" is stale too (33 now).

2. **`a9ad7495` (the `read_jpeg` timeout) is unrelated scope.** It's a pre-existing flake — `test_read_jpeg_encodes_bgr_frame` is on `main` as well — fixed here to turn CI green. The change itself is fine: the 20 ms default is preserved so the three video-rate call sites are untouched, and only `read_jpeg` opts into 1 s (whose sole caller `media_manager.get_frame_jpeg` has no callers at all today, so the blast radius is nil). But it's a different subsystem, and it makes this PR non-atomic to revert. I'd split it into its own PR rather than block on it. FWIW it passes on aarch64 too — skipped by default there via `importorskip("cv2")`, 3/3 with `--extra opencv`.

Everything else from my earlier rounds is resolved. Nice work chasing this one down through three iterations.

tfrere:
both sorted, thanks. updated the description to match the code (dropped the --no-block wording, now documents the whitelisted argv + ok_returncodes=(0, -15, 143) with the sudoers argv-exact reasoning, and 31 -> 33 passed). and split the read_jpeg timeout out into #1328 against main, so this PR is back to just the wireless_version change. appreciate the three passes on this one.

PR #1313, opened 2026-08-03
feat(daemon): expose the IMU reading over the data channel (get_imu)

tfrere:

New `get_imu` command on the transport-agnostic dispatcher: replies with the BMI088 reading (accelerometer, gyroscope, quaternion, temperature) or an explicit `imu: null` on IMU-less hardware (Lite version, simulation), so clients can tell "no IMU" from "old daemon that never replies".

- The reading already existed backend-side (published at 50 Hz on the local WebSocket); this exposes it to WebRTC clients too.
- `get_imu` is served from a cache refreshed by the 50 Hz control loop, not a direct BMI088 read: two concurrent readers on the sensor would interleave I2C transactions and inject spurious quaternion integration steps.
- Drive-by docs fix: the protocol docstring now also lists the `welcome` push in the server-to-client message inventory.

Behavioural pairing (fail-open, any merge order): the JS block's `getImu()` resolves `null` against older daemons.

Tests: `test_backend_process_command.py` (56 passed locally, incl. the two `get_imu` dispatcher tests). Validated on hardware (wireless robot).

### DIFF
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -35,6 +35,7 @@ from reachy_mini.io.protocol import (
     DeleteHfTokenCmd,
     FaceTarget,
     GetHardwareIdCmd,
+    GetImuCmd,
     GetMicrophoneVolumeCmd,
     GetMotorModeCmd,
     GetRobotNameCmd,
@@ -44,6 +45,7 @@ from reachy_mini.io.protocol import (
     GetVolumeCmd,
     GotoSleepCmd,
     GotoTargetCmd,
+    ImuDataMsg,
     LogLineMsg,
     LogStreamErrorMsg,
     MockupSimBackendStatus,
@@ -489,6 +491,15 @@ class Backend:
 
         """
         self.imu_publisher = publisher
+
+    def get_imu_data(self) -> ImuDataMsg | None:
+        """Get the current IMU reading, or None when the backend has no IMU.
+
+        Overridden by the real-robot backend (BMI088 on the wireless
+        version); the default covers simulation backends and the Lite
+        version so `get_imu` can answer uniformly on every transport.
+        """
+        return None
 
     def update_target_head_joints_from_ik(
         self,
@@ -1541,6 +1552,19 @@ class Backend:
             from reachy_mini.utils.hardware_id import get_hardware_id
 
             send_response({"hardware_id": get_hardware_id()})
+
+        elif isinstance(cmd, GetImuCmd):
+            # `imu` is None on IMU-less hardware (Lite, simulation) so
+            # clients can tell "no IMU" from "no reply" (old daemon).
+            imu = self.get_imu_data()
+            send_response(
+                {
+                    "command": "get_imu",
+                    "imu": imu.model_dump(exclude={"type"})
+                    if imu is not None
+                    else None,
+                }
+            )
 
         elif isinstance(cmd, (GetRobotNameCmd, SetRobotNameCmd)):
             # Robot display name is a persistent, robot-wide string stored on
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -493,12 +493,7 @@ class Backend:
         self.imu_publisher = publisher
 
     def get_imu_data(self) -> ImuDataMsg | None:
-        """Get the current IMU reading, or None when the backend has no IMU.
-
-        Overridden by the real-robot backend (BMI088 on the wireless
-        version); the default covers simulation backends and the Lite
-        version so `get_imu` can answer uniformly on every transport.
-        """
+        """Latest IMU reading; None on backends without an IMU."""
         return None
 
     def update_target_head_joints_from_ik(
@@ -1557,14 +1552,8 @@ class Backend:
             # `imu` is None on IMU-less hardware (Lite, simulation) so
             # clients can tell "no IMU" from "no reply" (old daemon).
             imu = self.get_imu_data()
-            send_response(
-                {
-                    "command": "get_imu",
-                    "imu": imu.model_dump(exclude={"type"})
-                    if imu is not None
-                    else None,
-                }
-            )
+            payload = imu.model_dump(exclude={"type"}) if imu is not None else None
+            send_response({"command": "get_imu", "imu": payload})
 
         elif isinstance(cmd, (GetRobotNameCmd, SetRobotNameCmd)):
             # Robot display name is a persistent, robot-wide string stored on
--- a/src/reachy_mini/daemon/backend/robot/backend.py
+++ b/src/reachy_mini/daemon/backend/robot/backend.py
@@ -135,6 +135,20 @@ class RobotBackend(Backend):
         else:
             self.bmi088 = None
 
+        # Latest IMU reading, refreshed by the control loop each tick and
+        # served to on-demand readers (`get_imu_data`). The loop is the only
+        # code that talks to the BMI088: its `get_quat(dt)` call integrates
+        # the orientation filter with a fixed loop period, so an on-demand
+        # read would both race the i2c conversation and inject a spurious
+        # integration step. Stored as (msg, monotonic ts) so a reader can
+        # judge freshness in one atomic reference grab.
+        self._last_imu: tuple[ImuDataMsg, float] | None = None
+
+    # A cached reading older than this is treated as absent: the control
+    # loop refreshes at 50 Hz, so anything past a few periods means the
+    # loop is stalled and the data no longer describes the present.
+    IMU_CACHE_FRESH_S = 0.5
+
     def run(self) -> None:
         """Run the control loop for the robot backend.
 
@@ -257,10 +271,12 @@ class RobotBackend(Backend):
                         )
                     )
 
-                    if self.imu_publisher is not None and self.bmi088 is not None:
-                        imu_msg = self.get_imu_data()
+                    if self.bmi088 is not None:
+                        imu_msg = self._read_imu()
                         if imu_msg is not None:
-                            self.imu_publisher.put(imu_msg)
+                            self._last_imu = (imu_msg, time.monotonic())
+                            if self.imu_publisher is not None:
+                                self.imu_publisher.put(imu_msg)
 
                 self.last_alive = time.time()
 
@@ -495,7 +511,27 @@ class RobotBackend(Backend):
         return np.array(self.get_all_joint_positions()[1])
 
     def get_imu_data(self) -> ImuDataMsg | None:
-        """Get current IMU data (accelerometer, gyroscope, quaternion, temperature).
+        """Return the latest IMU reading cached by the control loop.
+
+        On-demand readers (the `get_imu` data-channel command) must not
+        touch the BMI088 themselves - see the `_last_imu` comment. A cache
+        entry older than ``IMU_CACHE_FRESH_S`` reads as absent so a stalled
+        loop can't keep serving a stale orientation.
+
+        Returns:
+            An ImuDataMsg, or None if the IMU is missing or the cache is stale.
+
+        """
+        cached = self._last_imu
+        if cached is None:
+            return None
+        msg, ts = cached
+        if time.monotonic() - ts > self.IMU_CACHE_FRESH_S:
+            return None
+        return msg
+
+    def _read_imu(self) -> ImuDataMsg | None:
+        """Read the BMI088 directly. Control-loop only (see `_last_imu`).
 
         Returns:
             An ImuDataMsg, or None if IMU is not available.
@@ -696,5 +732,3 @@ class RobotBackend(Backend):
 
         result: bytes = bytes(self.c.write_raw_packet(packet))
         return result
-
-
--- a/src/reachy_mini/daemon/backend/robot/backend.py
+++ b/src/reachy_mini/daemon/backend/robot/backend.py
@@ -135,18 +135,14 @@ class RobotBackend(Backend):
         else:
             self.bmi088 = None
 
-        # Latest IMU reading, refreshed by the control loop each tick and
-        # served to on-demand readers (`get_imu_data`). The loop is the only
-        # code that talks to the BMI088: its `get_quat(dt)` call integrates
-        # the orientation filter with a fixed loop period, so an on-demand
-        # read would both race the i2c conversation and inject a spurious
-        # integration step. Stored as (msg, monotonic ts) so a reader can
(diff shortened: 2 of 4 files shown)


src/reachy_mini/daemon/backend/robot/backend.py
```diff
@@ -135,6 +135,20 @@ def __init__(
...
+        # the orientation filter with a fixed loop period, so an on-demand
+        # read would both race the i2c conversation and inject a spurious
+        # integration step. Stored as (msg, monotonic ts) so a reader can
+        # judge freshness in one atomic reference grab.
+        self._last_imu: tuple[ImuDataMsg, float] | None = None
+
+    # A cached reading older than this is treated as absent: the control
+    # loop refreshes at 50 Hz, so anything past a few periods means the
+    # loop is stalled and the data no longer describes the present.
+    IMU_CACHE_FRESH_S = 0.5
```
FabienDanieau: The `(msg, ts)` tuple + `IMU_CACHE_FRESH_S` + the comparison guard a window in which the loop is already dead: `_update` sets `self.error = "No response from the robot's motor for the last second."` and re-raises after a second of motor silence, and the daemon status already reports it. Three coherent options — worth picking one:

- **bare cache**: `self._last_imu: ImuDataMsg | None = None`, return it. -8 lines, and in every non-pathological case the client sees exactly what it sees today.
- **expose it**: keep the timestamp and put an `age_s` in the reply, so the *client* judges freshness instead of the daemon deciding on its behalf with a fixed window.
- **keep as is** + a test for the stale branch (see the note on `get_imu_data`).

Two smaller things here: 0.5 s at 50 Hz is 25 periods, not the "few periods" the comment claims. And this 7-line rationale is repeated on the constant just below and again in the `get_imu_data` docstring — state it once, on `_read_imu`, where the BMI088 is actually touched.

src/reachy_mini/daemon/backend/robot/backend.py
```diff
@@ -495,7 +511,27 @@ def get_present_antenna_joint_positions(
...
+            An ImuDataMsg, or None if the IMU is missing or the cache is stale.
+
+        """
+        cached = self._last_imu
+        if cached is None:
+            return None
+        msg, ts = cached
+        if time.monotonic() - ts > self.IMU_CACHE_FRESH_S:
+            return None
+        return msg
```
FabienDanieau, line 527: Docstring restates the `_last_imu` comment almost verbatim; one line plus the existing `Returns:` covers it.

More importantly, this is the only piece of genuinely new logic in the PR and nothing exercises it — both new tests mock `get_imu_data` on the sim backend, so the `> IMU_CACHE_FRESH_S` branch never runs. Five lines, using the pattern already in `test_robot_backend_status.py` / `test_enable_motors_pin.py`:

```python
b = RobotBackend.__new__(RobotBackend)
b._last_imu = (msg, time.monotonic() - 1.0)
assert RobotBackend.get_imu_data(b) is None
```

src/reachy_mini/daemon/backend/robot/backend.py
```diff
@@ -257,10 +271,12 @@ def _update(self) -> None:
...
 
-                    if self.imu_publisher is not None and self.bmi088 is not None:
-                        imu_msg = self.get_imu_data()
+                    if self.bmi088 is not None:
+                        imu_msg = self._read_imu()
                         if imu_msg is not None:
-                            self.imu_publisher.put(imu_msg)
+                            self._last_imu = (imu_msg, time.monotonic())
+                            if self.imu_publisher is not None:
+                                self.imu_publisher.put(imu_msg)
```
FabienDanieau, line 279: The cache refresh inherits three guards that have nothing to do with the IMU: it sits inside `if self.joint_positions_publisher is not None and self.pose_publisher is not None:`, inside the `try` that re-raises the motor `RuntimeError`, and inside `if not self.is_shutting_down:`.

Net effect: `get_imu` answers `null` during shutdown and after a motor read failure. That's the right outcome, but the IMU cache now silently depends on the pose/joint publishers being wired — true today only because `WSServer.start()` sets all four at once. Either hoist the block out of the nest, or note the dependency in the comment above.

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -1542,6 +1553,19 @@ def _maybe_ignore(field: str) -> bool:
...
+            # clients can tell "no IMU" from "no reply" (old daemon).
+            imu = self.get_imu_data()
+            send_response(
+                {
+                    "command": "get_imu",
+                    "imu": imu.model_dump(exclude={"type"})
+                    if imu is not None
+                    else None,
+                }
+            )
```
FabienDanieau: 11 lines for one dict; the multi-line ternary is the bulk of it. Three lines, all under 88 chars:

```python
imu = self.get_imu_data()
payload = imu.model_dump(exclude={"type"}) if imu else None
send_response({"command": "get_imu", "imu": payload})
```

Keep the comment — the explicit `null` is the non-obvious part.

Worth noting the reply carries no age, so a client can't distinguish a 5 ms-old reading from a 480 ms-old one (see the freshness note in `robot/backend.py`).

src/reachy_mini/daemon/backend/abstract.py
```diff
@@ -490,6 +492,15 @@ def set_imu_publisher(self, publisher: Publisher) -> None:
         """
         self.imu_publisher = publisher
 
+    def get_imu_data(self) -> ImuDataMsg | None:
+        """Get the current IMU reading, or None when the backend has no IMU.
+
+        Overridden by the real-robot backend (BMI088 on the wireless
+        version); the default covers simulation backends and the Lite
+        version so `get_imu` can answer uniformly on every transport.
+        """
```
FabienDanieau: Five lines of docstring for `return None`. `"""Latest IMU reading; None on backends without an IMU."""` says it — "overridden by the real-robot backend" is discoverable from the subclass, and it's the kind of line that goes stale the moment a sim backend synthesises an IMU.

src/reachy_mini/io/protocol.py
```diff
@@ -284,6 +285,17 @@ class GetHardwareIdCmd(BaseModel):
     type: Literal["get_hardware_id"] = "get_hardware_id"
 
 
+class GetImuCmd(BaseModel):
+    """Query the current IMU reading (BMI088, wireless version only).
+
+    The response carries ``imu``: the `ImuDataMsg` payload
+    (accelerometer, gyroscope, quaternion, temperature) or ``None``
+    when the robot has no IMU (Lite version, simulation).
+    """
```
FabienDanieau: `GetStateCmd` / `GetVersionCmd` / `GetHardwareIdCmd` immediately above are all one-line docstrings, and the payload shape is already documented on `ImuDataMsg`. One line keeps the file uniform:

```python
"""Query the current IMU reading (BMI088, wireless only; null without one)."""
```

FabienDanieau (commented):
Read the whole thing — the design call is right. Serving `get_imu` from the 50 Hz cache instead of an on-demand BMI088 read is the correct choice, and the stated reason holds: `get_quat(dt)` integrates with a fixed period, so a second reader would race the I2C conversation *and* inject a bogus integration step.

Checked, all clean:

- the only in-repo caller of `RobotBackend.get_imu_data()` was the control loop, so re-pointing that name at the cache breaks nothing;
- `test_js_protocol_sync.py` validates JS → protocol only, so a Python-only command can't fail CI — the "any merge order" claim really does hold for daemon-first;
- `run()` is a `Thread` (`daemon.py:400`), same process as the dispatcher, so the cache is genuinely shared and the tuple swap is atomic;
- dropping the `imu_publisher is not None` guard costs no extra I2C — `WSServer.start()` wires all four publishers together. Arguably a fix: the orientation filter no longer depends on a subscriber existing.

Two things worth a decision, both inline:

1. the staleness machinery — bare cache, or expose the timestamp, or keep the window *and* test it. Right now it's the most code of the three options and the stale branch never runs in the tests.
2. ~30 lines of comment/docstring for ~20 lines of code, with the "control loop is the only BMI088 reader" rationale written three times.

Design question, not blocking: #1296 just moved pose *off* polling onto a pushed channel, and `get_imu` is a poll. Fine for one-shot reads, but a client wanting continuous IMU will poll the reliable-ordered channel at rate — and every data-channel command hits `_cancel_idle_reset()` (`abstract.py:2569`), so IMU polling suppresses idle-sleep indefinitely. Do the #1315 `getImu()` consumers need a stream, or is this genuinely one-shot?

Net across the inline notes: ~35 lines shorter, same behaviour. Nothing blocking.

<sub>Review assisted by Claude (`claude-opus-5[1m]`).</sub>

src/reachy_mini/daemon/backend/robot/backend.py
tfrere: went with option 3: kept the window, added the test (test_get_imu_data_cache_freshness covers empty / fresh / stale). Also fixed the comment, it now says 25 periods and points at the motor-error death as the scenario it guards. Rationale dedup done too, it lives on _read_imu now. 8db367cff

src/reachy_mini/daemon/backend/robot/backend.py
tfrere: docstring trimmed to the Returns block, and the stale branch is now covered with exactly the __new__ pattern you sketched. 8db367cff

src/reachy_mini/daemon/backend/robot/backend.py
tfrere: noted the dependency in a comment rather than hoisting: the block also wants the try's motor-error handling and the is_shutting_down guard, so pulling it out would mean duplicating those. The comment says the wiring is only true because WSServer.start() sets all four together. 8db367cff

src/reachy_mini/daemon/backend/abstract.py
tfrere: done, kept the comment. On the age: getImu() in #1315 is a one-shot read, so anything under the 0.5 s window is fresh enough for what it's used for; if a consumer ever needs to judge freshness we can add age_s then without breaking anyone. 8db367cff

src/reachy_mini/daemon/backend/abstract.py
tfrere: took your line verbatim. 8db367cff

src/reachy_mini/io/protocol.py
tfrere: one-lined, matches the neighbours now. 8db367cff

tfrere:
All inline notes addressed in 8db367cff: rationale stated once on _read_imu, freshness comment fixed (25 periods), publisher-coupling noted, reply collapsed, both docstrings one-lined, and the stale branch now has a test. Net -15 lines outside tests, same behaviour, 57 passed locally.

On the poll vs push question: getImu() in #1315 is genuinely one-shot, one roundtrip per call, no polling loop in the SDK. It's there for spot reads (orientation check before a move, temperature). If a client wants continuous IMU they should get a pushed channel like #1296 did for pose - the 50 Hz publisher already exists backend-side, so wiring it to a set_imu_provider hook on the media server would be a small follow-up PR rather than something to squeeze in here. And agreed the idle-reset interaction makes aggressive polling a bad pattern; a push channel sidesteps that too.

FabienDanieau (approved):
LGTM — approving. Every inline point is addressed cleanly in 8db367cff, and CI is green on this head including the WebRTC step that the earlier segfault had skipped (that red was a pre-existing intermittent teardown crash in the Ubuntu audio+video step — same signature on `feat/js-sdk-host-block`, a TypeScript-only branch — not this PR).

**Correction to my own review, first.** My idle-reset warning was wrong, and you agreed with it, so it's on the record twice: `request_idle_reset()` is called only when the shared `RobotAppLock` transitions to *free* — session ended or local app exited (`daemon.py:268`) — not on an inactivity timer. `_cancel_idle_reset()` on an incoming command just means "a new owner grabbed the robot", which is correct. A live peer polling `get_imu` has nothing to suppress, and once it drops no commands arrive. That removes an argument *against* polling rather than adding one — sorry for the noise it caused in your reasoning.

**Verified end to end on the wireless robot.** Daemon running from an editable install at `8db367cff` (`git diff` vs the PR head empty). The reply path is confirmed by the `sdk-js-demo-app` Space, which can only be fed by this PR: `StateSnapshot` carries no IMU (`protocol.py:114-121`), the 50 Hz `imu_data` publisher reaches only local-WebSocket clients, and the Space calls `reachy.getImu()` (`src/embed.ts:921`) via the vendored #1315 SDK build. Live values in that panel = `get_imu` round-tripping over the data channel.

**Hardware numbers, and I withdraw my "tighten to ~0.1 s".** 30 s of the live stream: 48.9 Hz, 389/391 distinct accelerometer samples, `|quat|` exactly 1.000000, `|accel|` 9.39-10.40 m/s². Server-side `control_loop_stats` over 30 windows: 47.6-49.2 Hz, worst single interval **48.1 ms** — 9.6 % of `IMU_CACHE_FRESH_S`. So 0.5 s has ~10x margin and 0.1 s would have had ~2x. Keeping the window was the right call. (Transport-side I saw gaps up to 1.5 s on Wi-Fi while the loop held ~48 Hz — which is a decent argument that serving from the cache beats "use the last streamed frame".)

**One thing for a follow-up PR, explicitly not for this one.** `WSServer._handle_command` passes a no-op `send_response` (`ws_server.py:168`, *"SDK commands are fire-and-forget"*), so over the local WebSocket `get_imu` never answers and "no IMU" is indistinguishable from "old daemon" — the property this PR is built on holds on WebRTC only. Pre-existing and true of every `Get*` command, so nothing to change here, but it means the Python SDK keeps needing `.imu` from the stream. Natural pairing with the `set_imu_provider` push channel you sketched: one follow-up that gives the WS a reply path and WebRTC a pushed IMU channel like #1296 did for pose. Happy to open the issue if useful.

Minor, non-blocking, take or leave: `test_get_imu_data_cache_freshness` is a `RobotBackend` unit test living in the dispatcher module (whose docstring scopes it to `abstract.py`) — `test_robot_backend_status.py` is the existing home for unbound-`RobotBackend` tests; and its `import time` / `import RobotBackend` can be top-level, as `test_enable_motors_pin.py:18` already proves.

Not verified anywhere but the unit test: `imu: null` on genuinely IMU-less hardware (no Lite to hand). Wouldn't hold the PR for it.

Aside: the demo app polls `getImu()` every 500 ms (`src/embed.ts:30`), so "no polling loop, spot reads" isn't quite how it's used in practice — harmless at 2 Hz, and with the idle-reset objection withdrawn there's no reason to care.

<sub>Review assisted by Claude (`claude-opus-5[1m]`).</sub>

PR #1314, opened 2026-08-03
feat(sdk/host): WebRTC resilience, daemon-update gate, host UX refresh

tfrere:
Split out of #1304 (see the split plan there). Everything under `ts/` - the SDK and host ship from the same package and the host is the SDK's first consumer, so they move together. 26 commits, original history preserved.

Every daemon-facing addition is fail-open against older daemons (unknown commands are silently dropped and the SDK resolves `null` on the shared 4 s timeout), so this has no build-time ordering constraint with the daemon PRs (#1310 DoA, #1311 handoff grace, #1312 update exit code, #1313 IMU) - only behavioural pairings, noted in each daemon PR.

## SDK

- **Build version**: `ReachyMini.version` / `instance.sdkVersion`, generated into `lib/version.ts` by a `prebuild:sdk` hook.
- **Automatic session re-dial**: ICE-blip grace (2.5 s disconnected / 1 s failed), network-change awareness, visibility-deferred judgement for backgrounded tabs, then a bounded re-dial loop (4 attempts, ~22 s) emitting `sessionReconnecting` / `sessionReconnected`. `setAutoReconnect()` lets the update flow disable it so a reboot's `sessionStopped` isn't absorbed.
- **Data-channel silence watchdog**: inbound traffic is the liveness signal (500 ms poll / 30 Hz pose stream); 2.5 s of silence triggers a `get_state` nudge, 8 s hands over to the re-dial. Visibility-aware to avoid false positives in throttled tabs.
- **Typed `startDaemonUpdate` surface**: `UpdateProgressEvent`, `StartDaemonUpdateOptions`, method on `ReachyMiniInstance`.
- **Refactors**: keyed reply-slot registry + public `peerConnection` getter; resilience policy extracted into `session-supervisor.ts`; all pending-promise mechanisms unified into `pending-replies.ts` with a single `settleAll()` on every teardown path. `reachy-mini.ts` shrinks by ~780 lines, public API unchanged.
- **`getImu()`**: one-shot IMU query, `null` on IMU-less robots and against older daemons (pairs with #1313).
- **DoA**: `DoA` type, optional `RobotState.doa` parsed off state frames (pairs with #1310).
- **Central fleet listener moved into the SDK**: the bare SSE listener (roles=["listener"], no peer slot) used by the host picker and the mobile scan screen is now one canonical exported module instead of two drifting copies. The TTL-eviction path now actually reconnects, and `onConnect` fires only after the `setPeerStatus` registration settles.
- **Generic `request()` escape hatch**: send any `{type, ...}` command and await the reply, fail-open `null` - so a future daemon command is usable by apps without waiting for an SDK release.

## Host (shell + embed)

- **Stable-id dialing**: central rotates peer ids on relay reconnect, so the embed re-resolves the live peer id right before `startSession()`, with the daemon `hardware_id` travelling through both handoff channels. Fail-open on every miss. Also `awaitFullyAwake()` (bounded 8 s) instead of fire-and-forget `ensureAwake()`.
- **Daemon update gate + embed SDK staleness self-check**, and reboot completion now requires an offline observation first (pairs with #1312).
- **Sleep-on-leave, mobile parity**: the embed plays `gotoSleep` + forces motors `disabled` while the session is still up, waits (bounded 1 s) for the state frame echoing `disabled`, then acks `embed:left`; the host waits for the ack instead of a fixed sleep (pairs with #1311).
- **Picker / top bar refresh**: SSE-driven list, spinner instead of the Refresh button, identity row + direct logout, outlined "End session".
- **Flicker-free boot + error boundary**: `authResolved` gate, single continuous splash across the boot legs, `HostErrorBoundary` with a themed fallback.
- **Silent OAuth sign-in + sliding token idle window**: prompt-less `authorize_url` retry on boot for returning users, token idle window slid on activity.

## CI / tests

- Vitest infra for `ts/` with wire-compat and version-policy suites; the TS test suite runs before the npm package build.
- 137 tests / 12 files passing, `tsc --noEmit` clean.
- Validated end-to-end on hardware (wireless robot) and in a full-software sim (daemon mockup + local central + host harness): dead-transport re-dial, backgrounding, daemon update, handoffs.


tfrere:
Superseded: split into #1315 (SDK half, base main) and #1316 (host half, stacked on #1315). Both branches were rebuilt commit-by-commit from this one and verified content-identical to it (empty tree diff); the SDK branch also builds and passes the suite with the CURRENT host code, proving the additive-API claim. Branch kept for reference.

PR #1315, opened 2026-08-03
feat(sdk-js): WebRTC resilience, additive APIs, internals refactor

tfrere:
Split out of #1304 (see the split plan there). This is the SDK half of the former JS block (#1314, closed): everything under `ts/lib` + package tooling. The host consumer half is stacked on top as a separate PR.

**Contract: the public *runtime* surface is additive** - no released method or event changes shape, and every daemon-facing addition is fail-open against older daemons (unknown commands are silently dropped; the SDK resolves `null` on the shared 4 s timeout). The current host builds and passes its suite against this branch unchanged (`tsc --noEmit` clean, old host code untouched). Two caveats for downstream (per review):

- The *type* surface is not strictly additive: `ReachyMiniConstructor` gains a static `version` and `ReachyMiniInstance` gains new members, so anything *implementing* those interfaces (realistically: mocks in downstream test suites) needs a `tsc` run and possibly new stubs. Plain consumers are unaffected.
- `autoReconnect` defaults to **on**, which is a behavioural change, not an addition: an app keying its "disconnected" UI off the `error` event now sees `sessionReconnecting` first and only hears a terminal event after the bounded re-dial loop (~22 s) gives up. Opt out with `autoReconnect: false` or `setAutoReconnect(false)`.

## Resilience

- **Automatic session re-dial**: ICE-blip grace (2.5 s disconnected / 1 s failed), network-change awareness, visibility-deferred judgement for backgrounded tabs, then a bounded re-dial loop (4 attempts, ~22 s) emitting `sessionReconnecting` / `sessionReconnected`. `setAutoReconnect()` lets flows like the daemon update disable it so a reboot's `sessionStopped` isn't absorbed.
- **Data-channel silence watchdog**: inbound traffic is the liveness signal (500 ms poll / 30 Hz pose stream); 2.5 s of silence triggers a `get_state` nudge, 8 s hands over to the re-dial (or emits the fatal `error` once and stands down when auto-reconnect is off). Visibility-aware to avoid false positives in throttled tabs.

## Additive APIs

- `ReachyMini.version` / `instance.sdkVersion` (generated `lib/version.ts`, `prebuild:sdk` hook, re-stamped by CI after the release version bump).
- Typed `startDaemonUpdate` surface (`UpdateProgressEvent`, `StartDaemonUpdateOptions`).
- `getImu()`: one-shot IMU query, `null` on IMU-less robots and against older daemons (pairs with #1313).
- `DoA` type + optional `RobotState.doa` parsed off state frames (pairs with #1310).
- `openCentralListener`: the bare SSE fleet listener (roles=["listener"], no peer slot) as one canonical exported module instead of drifting per-app copies; the TTL-eviction path now actually reconnects.
- `request()`: generic command round-trip escape hatch - send any `{type, ...}` command, await the reply, fail-open `null` - so a future daemon command is usable without waiting for an SDK release.
- Token store: sliding idle window on activity.

## Internals refactor (public API unchanged)

Keyed reply-slot registry + public `peerConnection` getter; resilience policy extracted into `session-supervisor.ts`; all pending-promise mechanisms unified into `pending-replies.ts` with a single `settleAll()` on every teardown path. `reachy-mini.ts` shrinks by ~780 lines.

## CI / tests

Vitest infra for `ts/`; the TS test suite now runs before the npm package build. 86 tests passing on this branch, `tsc --noEmit` clean. Validated end-to-end on hardware (wireless robot) and in a full-software sim (daemon mockup + local central + host harness).

## Addendum (2026-08-08): wake/sleep lifecycle hardening

Two more commits close the wake/sleep contract holes found while auditing
what happens when an app leaves motors in a non-default state:

**`feat(daemon)`: reset to sleep from any inherited motor state.** The idle
reset used to skip whenever the mode was already `Disabled` (even head
drooped mid-pose) and to run a bare `goto_sleep()` that degrades silently
under gravity compensation or limp motors. New `reset_to_sleep()` (quiesce,
torque on, lift to init, then the normal sleep) is used by the idle reset
and `Daemon.stop()`; the skip predicate is now "Disabled AND physically at
the sleep pose"; per-motor `set_torque(ids=...)` now forces the next
`set_motor_control_mode(Enabled)` to re-apply for real. Compatible with the
host-shell sleep-on-leave in #1316: its happy path still skips the reset,
and its degraded path (robot disabled mid-pose after a wedged goto_sleep)
now gets parked properly instead of staying drooped.

**`feat(sdk-js)`: `ensureAwake()` behavior change, deliberate.** It now
awaits the wake trajectory (internal 5 s budget, never rejects) instead of
fire-and-forget, so `connectToHost()`/`autoConnect()` resolve only once the
robot actually follows position targets; and it flips an inherited
`gravity_compensation` back to `enabled` (no emote) - previously such an
app booted "awake" with every position target silently ignored. Covered by
a new vitest suite (`ensure-awake.test.ts`, 6 tests; suite total 94).

Plus `docs(apps)`: new guide section 14.7 on clocking robot logic off a Web
Worker so hidden-tab throttling doesn't freeze pose streams.


ts/scripts/gen-version.mjs
```diff
@@ -0,0 +1,29 @@
...
+// own build version at runtime.
+//
+// Why generate instead of hardcoding: package.json is the single source
+// of truth npm publishes, and CI bumps it at release time (see RELEASE).
+// A hand-written constant would drift; deriving it here keeps the runtime
+// value byte-identical with what ends up on npm — zero maintenance.
+//
+// Runs automatically via the `prebuild:sdk` npm lifecycle hook. The output
+// (lib/version.ts) is committed with the placeholder value so typecheck /
+// editors work without a build; CI regenerates it after the version bump.
```
FabienDanieau, line 11: 🔴 **Blocking — this claim isn't true, and the feature is inert because of it.**

`.github/workflows/npm-publish.yml` runs, in order:

```
npm ci  →  npm test  →  npm run build   (L61)  →  LFS check  →  npm version "$ver"  (L127)  →  publish
```

`prebuild:sdk` fires `gen-version.mjs` during **`npm run build`** — before `npm version` bumps `package.json`. There's no `prepack` / `prepare` / `preversion` script to re-run it afterwards, and `npm publish` won't rebuild on its own.

So `dist/version.js` ships the committed placeholder on *every* release, which is precisely the `0.0.0-managed-by-ci` sentinel documented as "unreleased/branch build". `ReachyMini.version` never reports a real version on npm.

Fix — pick one:
- move the `Compute and apply version from pyproject.toml` step **above** `Build SDK and host bundles`;
- add `"prepack": "npm run gen:version && npm run build"`;
- insert `npm run gen:version && npm run build:sdk` right after `npm version`.

The third option is the one that finally gives `gen:version` (package.json L48) a caller — right now it's a duplicate of `prebuild:sdk` that nothing invokes.

_Minor, same file:_ L19's `typeof pkg.version === "string" ? ... : "0.0.0-unknown"` is a defensive check on your own `package.json`; npm guarantees the field.

ts/lib/session-supervisor.ts
```diff
@@ -0,0 +1,662 @@
...
+
+        // Dead transport. Stop judging (the re-dial tears this interval
+        // down anyway via the class's teardown; the error path keeps the
+        // session nominally up, so re-baseline to avoid a repeat every
+        // tick) and escalate through the same funnel as the ICE grace.
+        this._lastDcInboundAt = now;
+        this._dcSilenceNudged = false;
+        const cause = `No data-channel traffic for ${silence}ms`;
+        if (this.maybeBeginRedial(cause)) return;
+        this._deps.emitError(new Error(cause));
```
FabienDanieau: 🔴 **Blocking — this emits a fatal `error` every ~8 s, forever.**

The comment says the re-baseline avoids "a repeat every tick", and it does — but it doesn't avoid a repeat every *cycle*. `_lastDcInboundAt = now` leaves the interval running and the session nominally `streaming`, so a permanently dead channel just re-runs the whole escalation: nudge at 2.5 s, `error` at 8 s, repeat, indefinitely.

`types.ts` documents `autoReconnect: false` as _"restore the previous fatal-`error` behaviour"_ — which fired **once**.

```suggestion
        // Dead transport, and no redial to hand it to: emit once and
        // stand down. The next startDcWatchdog() re-arms.
        this._dcSilenceNudged = false;
        const cause = `No data-channel traffic for ${silence}ms`;
        if (this.maybeBeginRedial(cause)) return;
        this.stopDcWatchdog();
        this._deps.emitError(new Error(cause));
```

The re-baseline drops out with the stale comment: once the interval is stopped there's nothing left to re-baseline for, and the redial branch has the class teardown stop it anyway.

ts/lib/dc-watchdog.test.ts
```diff
@@ -0,0 +1,207 @@
...
+
+        expect(errors).toHaveLength(1);
+        expect(String((errors[0]!.error as Error).message)).toMatch(
+            /No data-channel traffic/,
+        );
+        expect(internals._supervisor.redialing).toBe(false);
+        // The error path keeps the session nominally up: the baseline was
+        // reset, so the very next tick doesn't re-emit.
+        await vi.advanceTimersByTimeAsync(2000);
+        expect(errors).toHaveLength(1);
```
FabienDanieau, line 161: Follow-on from the watchdog loop above: this only advances **2 000 ms**, which is below `DC_SILENCE_FATAL_MS` (8 000). So it verifies "the very next tick doesn't re-emit" and never reaches the second escalation.

Swap the `2000` for `10_000` and this assertion fails today — which is what makes the bug shippable. Worth keeping the extended version as the regression guard once it's fixed.

ts/lib/central-listener.ts
```diff
@@ -0,0 +1,399 @@
...
+      reconnectTimer = null;
+    }
+    if (abortController !== null) {
+      abortController.abort();
+      abortController = null;
+    }
+  };
+
+  const scheduleReconnect = (reason: string): void => {
+    if (closed) return;
```
FabienDanieau: ⚠️ `reconnectTimer` is overwritten without clearing the previous one, so two reconnects can be armed at once.

Reachable path:
1. the stream hits EOF → `scheduleReconnect('stream ended')` calls `stopHeartbeat()` and arms **timer 1**;
2. a heartbeat POST *already in flight* resolves with 400 → the eviction branch (L262-268) aborts and calls `scheduleReconnect('listener evicted')`, arming **timer 2** and orphaning timer 1.

Both fire → two concurrent `runStream()` → two SSE connections, and `abortController` only holds the second, so the first is unclosable and lingers until the server culls it.

One line at the top of the function:

```suggestion
  const scheduleReconnect = (reason: string): void => {
    if (closed || reconnectTimer !== null) return;
```

A cheap re-entrancy guard in `runStream()` wouldn't hurt either.

ts/lib/types.ts
```diff
@@ -444,9 +631,14 @@ export interface ReachyMiniInstance extends EventTarget {
...
 
-export type ReachyMiniConstructor = new (
+export type ReachyMiniConstructor = (new (
     options?: ReachyMiniOptions,
-) => ReachyMiniInstance;
+) => ReachyMiniInstance) & {
+    /** Build version of the SDK (npm package `version`), reachable without
+     *  an instance via `ReachyMini.version`. Mirror of the instance
+     *  `sdkVersion`. */
+    readonly version: string;
```
FabienDanieau, line 656: ⚠️ This one isn't additive: `ReachyMiniConstructor` goes from a bare construct signature to an intersection with `{ readonly version: string }`. Anything currently assignable to the type **without a static `version`** stops typechecking.

Same story for the `ReachyMiniInstance` additions in this PR (`peerConnection`, `sdkVersion`, `request()`, `getImu()`, `setAutoReconnect()`, `subscribePose()`, `unsubscribePose()`, `startDaemonUpdate()`) — fine for consumers, breaking for anyone *implementing* the interface. Realistically that means mocks in downstream test suites.

The runtime surface really is additive; worth narrowing the PR description's claim to that so nobody skips a `tsc` run downstream on the strength of it.

FabienDanieau (commented):
## Inline pass

Re-posting the review anchored to the code, per request — the standalone comment above has been removed in favour of this.

**Verdict: approve after the two 🔴 items.** Both are small, targeted fixes, but one makes a shipped feature inert (`ReachyMini.version` will read `0.0.0-managed-by-ci` on every npm release) and the other breaks the documented `autoReconnect: false` contract (fatal `error` on an 8 s loop instead of once). The ⚠️ SSE reconnect race is a two-line guard worth folding in while you're there.

The refactor itself is the strongest part of this PR. `PendingReplies.settleAll()` is a real correctness win — `disconnect()` previously never rejected pending RPC calls, so those promises just hung. The `settle`-identity trick in `slotRoundtrip` is right, and it's honest about the cross-talk it *doesn't* fix. And the DI on `SessionSupervisorDeps` is what makes the test suite cheap: 86 tests over the paths that actually matter, including the gating table, the give-up path, TTL eviction and "doesn't steal replies owned by typed handlers".

**Security:** clean. Token in the `Authorization` header rather than a URL query for the SSE stream, with the reasoning recorded; `sessionStorage` stays tab-scoped; the sliding idle window is a sane replacement for the pagehide wipe; 401/403 correctly terminal with no retry loop against a bad token.

**Test gaps** beyond the two flagged inline: nothing exercises `_teardownForRedial`'s fire-and-forget `endSession` against a null `_token`.

**Not worth an anchor:**
- `central-listener.ts` L229 — `sendListenerRegistration` uses `abortController?.signal`; after `close()` nulls it, a POST scheduled in that window isn't abortable. Cosmetic.
- Mixed indentation across `ts/lib/`: `reachy-mini.ts`, `session-supervisor.ts`, `token-store.ts`, `pending-replies.ts` use 4 spaces; `central-listener.ts` and three of the test files use 2. Worth a prettier pass before it sets the house style.

<sub>Comments tagged _"complexity pass"_ are a separate read hunting only for lines to delete — none of them are bugs, and they total roughly **-190 lines**. On comment density generally: these files run ~25 % prose, and while a few passages genuinely earn it (the TTL-eviction comment clearly encodes real debugging pain), most restate the code beneath them and will rot against it.</sub>

---

*Review assisted by Claude (`claude-opus-5[1m]`).*

ts/lib/types.ts
```diff
@@ -84,6 +119,41 @@ export interface ReachyMiniOptions {
...
+     * re-dials the same robot through the signaling server — up to 5
+     * attempts with backoff (~22 s window). The app sees
+     * `sessionReconnecting` per attempt, then either `sessionReconnected`
+     * followed by a fresh `streaming` event, or a terminal
+     * `sessionStopped` with reason `reconnect_failed`. Failures during
+     * the INITIAL `startSession()` handshake still reject the promise —
+     * the caller owns that retry. Set to `false` to restore the previous
+     * fatal-`error` behaviour.
+     */
+    autoReconnect?: boolean;
```
FabienDanieau, line 136: ⚠️ Defaulting to `true` makes this a behavioural change, not an addition — worth calling out loudly in the release notes.

An existing app that keys its "disconnected" UI off the `error` event now sits silent for ~22 s before hearing anything, and `sessionRejected` is suppressed for the duration (both `_failSessionRejected` and the `_handleEndSession` path gate on `_supervisor.redialing`).

That's the right behaviour long-term. But if any already-published Space embeds a frozen bundle that assumes the old semantics, shipping `false` for one release and flipping the default in the next is the cheaper rollback story.

ts/lib/session-supervisor.ts
```diff
@@ -0,0 +1,662 @@
...
+        if (this._redialTimer !== null) {
+            clearTimeout(this._redialTimer);
+            this._redialTimer = null;
+        }
+        if (this._redialWake) {
+            const wake = this._redialWake;
+            this._redialWake = null;
+            wake();
+        }
+    }
```
FabienDanieau, line 511: ⚠️ `setAutoReconnect(false)` during an in-flight redial lands here, the loop returns at its next `if (!this._redialing) return;` — and **nothing terminal is emitted**. But `_runRedialLoop` already called `teardownForRedial()` on entry, so the pc is closed and the session is gone.

Net effect: the app is left session-less with no `sessionStopped`, no `error`, nothing. It works for the daemon-update flow because that flow gets its terminal signal from elsewhere, but it's a sharp edge for any other caller.

Either document it on the public `setAutoReconnect()` ("aborts the redial and leaves the session torn down — the caller owns the terminal event"), or emit `sessionStopped` when cancelling a loop that has already torn the transport down.

ts/lib/reachy-mini.ts
```diff
@@ -1463,88 +1236,83 @@ export class ReachyMini extends EventTarget implements ReachyMiniInstance {
...
+            .catch((err: unknown): null => {
+                // Fail-open on the waiter timeout only; teardown rejections
+                // (settleAll) propagate to the caller like every other
+                // in-flight round-trip.
+                if (err instanceof Error && err.message.startsWith('broadcast timeout')) {
+                    return null;
+                }
+                throw err;
+            });
+        this._sendCommand(command);
```
FabienDanieau: Three small things in this block:

**L1302** — `.then((m): Record<string, unknown> | null => m)` is an identity callback that exists only to widen the type. Annotate `reply` instead and drop the line.

**L1307** — fail-open detection by `err.message.startsWith('broadcast timeout')` couples this to a string literal in `pending-replies.ts`. Reword that message and fail-open silently becomes a throw. A sentinel error class (or a `symbol` marker on the error) makes it a type check.

**L1312** — `_sendCommand`'s `false` return is dropped, so a call on a closed-mid-flight channel hangs to the timeout and resolves `null` instead of rejecting. `_sendCommandAwaitCompletion` rejects in that case — worth matching.

ts/lib/reachy-mini.ts
```diff
@@ -375,16 +349,21 @@ export class ReachyMini extends EventTarget implements ReachyMiniInstance {
...
         const opts: { clientId?: string } = {};
         if (this._clientId) opts.clientId = this._clientId;
-        window.location.href = await oauthLoginUrl(opts);
+        let url = await oauthLoginUrl(opts);
+        // OIDC prompt param. `oauthLoginUrl` doesn't expose it, but the HF
+        // authorize endpoint honours it (verified empirically): with
+        // `prompt=none` an already-authorized user comes straight back with
+        // a code and no screen, anyone else comes back with `?error=...`
+        // instead of landing on the HF login page.
+        if (options?.prompt) url += `&prompt=${options.prompt}`;
```
FabienDanieau: `url += ` assumes `oauthLoginUrl()` always returns a URL that already has a query string, and skips encoding. Both are true today, and `'none'` is the only permitted value — but the safe form is the same length:

```suggestion
        // OIDC prompt param. `oauthLoginUrl` doesn't expose it, but the HF
        // authorize endpoint honours it (verified empirically): with
        // `prompt=none` an already-authorized user comes straight back with
        // a code and no screen, anyone else comes back with `?error=...`
        // instead of landing on the HF login page.
        if (options?.prompt) {
            const u = new URL(url);
            u.searchParams.set('prompt', options.prompt);
            url = u.toString();
        }
```

ts/lib/token-store.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+export function consumeOAuthErrorParams(): string | null {
+    try {
+        const url = new URL(window.location.href);
+        const error = url.searchParams.get('error');
+        if (error === null) return null;
+        url.searchParams.delete('error');
+        url.searchParams.delete('error_description');
+        url.searchParams.delete('state');
+        window.history.replaceState(null, '', url.toString());
```
FabienDanieau, line 124: This strips `state` and `error_description` whenever **any** `error` param is present — including an app's own unrelated `?error=`, on a page load that has nothing to do with OAuth. It also rewrites history in that case.

Since it runs before `oauthHandleRedirectIfPresent()`, gate it on the error codes the silent flow actually returns:

```ts
const OAUTH_ERRORS = new Set(['login_required', 'consent_required', 'interaction_required', 'access_denied']);
if (error === null || !OAUTH_ERRORS.has(error)) return null;
```

ts/lib/types.ts
```diff
@@ -286,6 +414,20 @@ export interface ReachyMiniInstance extends EventTarget {
     readonly preselectedRobotId: string | null;
     /** `true` iff `preselectedRobotId !== null`. UX branching helper. */
     readonly isEmbedded: boolean;
+    /**
+     * Live RTCPeerConnection, or `null` between sessions. Read-only escape
+     * hatch for stats sampling (`getStats()`); mutating it is unsupported.
+     * Auto-reconnect re-dials REPLACE this object, so re-read it on every
+     * use — never capture it across ticks.
+     */
+    readonly peerConnection: RTCPeerConnection | null;
```
FabienDanieau, line 419: This and `_pc` (L434, just below) are now two public ways to read the same handle, both documented, both non-null-between-sessions.

Assuming `peerConnection` is meant to win, mark the old one so the next reader knows which to reach for:

```ts
    /** @deprecated Use `peerConnection`. Kept for compatibility. */
    _pc: RTCPeerConnection | null;
```

ts/vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    environment: 'jsdom',
+    include: ['host/src/**/*.test.{ts,tsx}', 'lib/**/*.test.ts'],
```
FabienDanieau, line 6: This glob and the `@testing-library/react` devDep both land ahead of their content — there are no `host/src/**` tests in this PR (they're in the stacked one). Harmless, just noting it so the empty half doesn't look like a mistake later.

Worth confirming `npm test` stays green in CI if the stacked PR merges in the other order.

ts/lib/central-listener.ts
```diff
@@ -0,0 +1,399 @@
...
+ *
+ * Reconnect policy
+ * ────────────────
+ * Network blips, HF Space cold-restarts, and proxy idle culls all
+ * surface as a stream EOF or fetch error. We retry with exponential
+ * backoff capped at `MAX_BACKOFF_MS`, jittered to avoid thundering
+ * herds on a central restart. Fatal auth errors (401/403) end the
+ * stream instead - retrying with the same bad token would only get
+ * us rate-limited.
+ */
```
FabienDanieau: 47-line prologue. The one genuinely non-obvious thing in it is the "why not the SDK class" paragraph — central's 1:1 `token → peer_id` mapping means a fleet-watching UI would clobber the peer slot. That's worth keeping.

The wire-surface table, the transport notes and the reconnect-policy essay all restate code that's a screen away and will drift from it. ~6 lines covers the same ground.

<sub>_Complexity pass — no bug here, purely about the diff getting shorter._</sub>

ts/lib/central-listener.ts
```diff
@@ -0,0 +1,399 @@
...
+  const raw = welcomeMsg['recommended_heartbeat_interval_seconds'];
+  if (typeof raw === 'number' && Number.isFinite(raw) && raw > 0) {
+    return clampHeartbeatIntervalMs(raw * 1000);
+  }
+  const lease = welcomeMsg['lease_seconds'];
+  if (typeof lease === 'number' && Number.isFinite(lease) && lease > 0) {
+    return clampHeartbeatIntervalMs((lease * 1000) / 3);
+  }
+  return HEARTBEAT_DEFAULT_INTERVAL_MS;
+};
```
FabienDanieau, line 53: 23 lines to read one optional number.

The MIN/MAX clamp defends against a misconfigured **first-party** central telling you to heartbeat every 1 ms — if that ships, the clamp isn't the fix. Dropping it also removes the need for `clampHeartbeatIntervalMs`:

```ts
const HEARTBEAT_DEFAULT_INTERVAL_MS = 10_000;

const negotiateHeartbeatIntervalMs = (w: Record<string, unknown>): number => {
  const s = (w['recommended_heartbeat_interval_seconds'] ?? Number(w['lease_seconds']) / 3) as number;
  return Number.isFinite(s) && s > 0 ? s * 1000 : HEARTBEAT_DEFAULT_INTERVAL_MS;
};
```

~-16 lines.

<sub>_Complexity pass — no bug here, purely about the diff getting shorter._</sub>

ts/lib/central-listener.ts
```diff
@@ -0,0 +1,399 @@
...
+export interface CentralStreamProducer {
+  id: string;
+  meta?: Record<string, unknown>;
+  busy?: boolean;
+  activeApp?: string | null;
+}
+
+export interface CentralListEvent {
+  producers: CentralStreamProducer[];
+}
```
FabienDanieau: `CentralListEvent` wraps a single array field. `onList?: (producers: CentralStreamProducer[]) => void` says the same thing without the indirection at the call site.

<sub>_Complexity pass — no bug here, purely about the diff getting shorter._</sub>

ts/lib/central-listener.ts
```diff
@@ -0,0 +1,399 @@
...
+          peerId: typeof msg.peerId === 'string' ? msg.peerId : '',
+          busy: msg.busy === true,
+          activeApp:
+            typeof msg.activeApp === 'string' ? msg.activeApp : null,
+          meta:
+            msg.meta && typeof msg.meta === 'object'
+              ? (msg.meta as Record<string, unknown>)
+              : undefined,
+        });
+        return;
```
FabienDanieau: 22 lines hand-coercing every field of two frames — while `producers` in the `list` branch just above is cast wholesale with a single `Array.isArray` check.

Both frames come from the same first-party server, so the inconsistency is the real finding: either validate all three or cast all three. Casting is ~6 lines and matches what the SDK's own signaling parser already does.

<sub>_Complexity pass — no bug here, purely about the diff getting shorter._</sub>

ts/lib/session-supervisor.ts
```diff
@@ -0,0 +1,662 @@
...
+ * unpredictably, so silence measured there is meaningless: ticks
+ * re-baseline while hidden, and a tick arriving after a large gap
+ * (throttled / suspended timer) re-baselines too instead of judging
+ * stale data. 8 s of zero inbound traffic on a link that normally
+ * delivers 2-30 messages per second is unambiguous, and still ~4×
+ * faster than the RFC 7675 consent-freshness teardown (~30 s).
+ */
+const DC_WATCHDOG_TICK_MS = 1000;
+const DC_SILENCE_NUDGE_MS = 2500;
+const DC_SILENCE_FATAL_MS = 8000;
```
FabienDanieau, line 75: 85 lines of rationale for 8 constants.

Two of them genuinely earn it — the RFC 7675 consent-freshness derivation behind `MAX_VISIBILITY_DEFER_MS`, and the observed `failed → connected` flip on iOS AP roams that justifies debouncing a state the spec calls terminal. Neither is guessable from the code.

The rest restates the constant beneath it in prose ("Backoff schedule … one entry per attempt, in ms of wait BEFORE that attempt"). ~20 lines total keeps everything load-bearing.

<sub>_Complexity pass — no bug here, purely about the diff getting shorter._</sub>

ts/lib/session-supervisor.ts
```diff
@@ -0,0 +1,662 @@
...
+        downlink?: number;
+        rtt?: number;
+        saveData?: boolean;
+    }): void;
+}
+
+/* ─── Supervisor ───────────────────────────────────────────────────── */
+
+export class SessionSupervisor {
+    private readonly _deps: SessionSupervisorDeps;
```
FabienDanieau: Seven dep members that all do exactly `this._emit(name, detail)` in the constructor. One typed member covers them:

```ts
emit<K extends keyof ReachyMiniEventMap>(name: K, detail: ReachyMiniEventMap[K]['detail']): void;
```

Costs a `.filter(c => c[0] === 'error')` in the test asserts, saves ~12 lines of interface + 7 arrows in the constructor + 7 `vi.fn()` in the mock factory. Judgement call on whether the test ergonomics are worth it.

<sub>_Complexity pass — no bug here, purely about the diff getting shorter._</sub>

ts/lib/session-supervisor.ts
```diff
@@ -0,0 +1,662 @@
...
+    }
+
+    /**
+     * An explicit dial from the app supersedes any in-flight
+     * auto-reconnect (possibly towards a different robot). No-op when
+     * the caller is the redial loop's own dial attempt.
+     */
+    noteExternalDial(): void {
+        if (this._redialing && !this._redialInternalDial) this.cancelRedial();
+    }
```
FabienDanieau: `noteExternalDial()` + `_redialInternalDial` (L192) + the try/finally at L507-515 exist solely to tell "the redial loop is dialling" from "the app is dialling".

Pointing the `dial` dep at a private `_startSessionInternal` removes the distinction entirely — the public `startSession` becomes `cancelRedial()` then `_startSessionInternal()`, and the loop calls the private one. ~-12 lines and one less piece of mutable state to reason about.

<sub>_Complexity pass — no bug here, purely about the diff getting shorter._</sub>

ts/lib/session-supervisor.ts
```diff
@@ -0,0 +1,662 @@
...
+        if (ms <= 0) return Promise.resolve();
+        return new Promise<void>((resolve) => {
+            this._redialWake = resolve;
+            this._redialTimer = setTimeout(() => {
+                this._redialTimer = null;
+                this._redialWake = null;
+                resolve();
+            }, ms);
+        });
+    }
```
FabienDanieau, line 496: `_redialTimer` exists only to be cleared by `cancelRedial`. But `_runRedialLoop` rechecks `if (!this._redialing) return;` immediately after every `await this._redialSleep(...)`, so a timer that fires post-cancel is already a no-op.

Keep `_redialWake`, drop the timer field and its clear in `cancelRedial`. ~-6 lines.

<sub>_Complexity pass — no bug here, purely about the diff getting shorter._</sub>

ts/lib/session-supervisor.ts
```diff
@@ -0,0 +1,662 @@
...
+/** Reject `p` with `label` if it hasn't settled within `ms`. */
+function withTimeout<T>(p: Promise<T>, ms: number, label: string): Promise<T> {
+    return new Promise<T>((resolve, reject) => {
+        const timer = setTimeout(() => reject(new Error(label)), ms);
+        p.then(
+            (v) => { clearTimeout(timer); resolve(v); },
+            (e) => { clearTimeout(timer); reject(e); },
+        );
+    });
+}
```
FabienDanieau, line 596: Hand-rolled timeout with `clearTimeout` bookkeeping, one call site, 15 s ceiling:

```suggestion
function withTimeout<T>(p: Promise<T>, ms: number, label: string): Promise<T> {
    return Promise.race([
        p,
        new Promise<T>((_, rj) => setTimeout(() => rj(new Error(label)), ms)),
    ]);
}
```

The leaked timer is bounded by `ms` and there's exactly one caller.

<sub>_Complexity pass — no bug here, purely about the diff getting shorter._</sub>

ts/lib/session-supervisor.ts
```diff
@@ -0,0 +1,662 @@
...
+                removeEventListener?: (type: string, listener: () => void) => void;
+            };
+        }).connection;
+        if (conn && this._connectionChangeHandler && typeof conn.removeEventListener === 'function') {
+            conn.removeEventListener('change', this._connectionChangeHandler);
+        }
+        this._onlineHandler = null;
+        this._offlineHandler = null;
+        this._connectionChangeHandler = null;
+    }
```
FabienDanieau, line 402: 53 lines whose entire job is re-emitting `window.online` / `offline` / `connection.change` verbatim as SDK events. Nothing inside the SDK consumes them — `emitNetworkOnline` and friends have no internal callers, and apps can call `window.addEventListener('online')` themselves.

**Pre-existing** (moved from `reachy-mini.ts`, not added here), so it's outside this PR's scope and I've left it out of the line count. Flagging it only because the code is already in your hands this week.

<sub>_Complexity pass — no bug here, purely about the diff getting shorter._</sub>

ts/lib/reachy-mini.ts
```diff
@@ -902,200 +863,69 @@ export class ReachyMini extends EventTarget implements ReachyMiniInstance {
...
-        if (conn && typeof conn.addEventListener === 'function') {
-            const onChange = (): void => this._emit('networkChange', {
-                effectiveType: conn.effectiveType,
-                downlink: conn.downlink,
-                rtt: conn.rtt,
-                saveData: conn.saveData,
-            });
-            conn.addEventListener('change', onChange);
-            this._connectionChangeHandler = onChange;
+    private _teardownForRedial(): void {
```
FabienDanieau, line 898: This is the **fourth** copy of the same teardown sequence — `clearIceGrace`, `uninstallNetworkListeners`, both intervals, `stopDcWatchdog`, `endSession`, micStream stop, pc close, flag reset — alongside `stopSession()`, `disconnect()` and `_failSessionRejected()`.

They differ only in which error they settle with, whether they emit, and where `_state` lands. One `_closeTransport()` holding the common part would leave each caller with its own three or four lines, and would make it structurally impossible for the next teardown path to forget a step (which is the same argument `PendingReplies.settleAll()` makes, and makes well).

Biggest single cut available here: ~-35 lines.

<sub>_Complexity pass — no bug here, purely about the diff getting shorter._</sub>

ts/lib/pending-replies.ts
```diff
@@ -0,0 +1,302 @@
...
+        return new Promise<unknown>((resolve, reject) => {
+            const timer = setTimeout(() => {
+                this._pendingRpc.delete(id);
+                reject(new Error(`rpcCall(${method}) timed out after ${timeoutMs}ms`));
+            }, timeoutMs);
+            this._pendingRpc.set(id, { resolve, reject, timer });
+        });
+    }
+
+    /**
```
FabienDanieau: `nextRpcId()` then `registerRpc(id, ...)` is a two-call API for one operation, and it makes the caller thread the id between them.

`slotRoundtrip` right above already has the better shape — mint internally, take a `send` callback:

```ts
rpcRoundtrip(method: string, timeoutMs: number, send: (id: string) => boolean): Promise<unknown>
```

`rpcCall` then loses its id juggling too. ~-8 lines across both files.

<sub>_Complexity pass — no bug here, purely about the diff getting shorter._</sub>

ts/lib/types.ts
```diff
@@ -249,6 +351,30 @@ export interface NetworkChangeEventDetail {
...
+ * Auto-reconnect succeeded. A fresh `streaming` event fires alongside
+ * (same robot, new sessionId); video re-attaches automatically through
+ * the `videoTrack` listener installed by `attachVideo()`. App-level
+ * state that lives daemon-side per session (motor mode override, a
+ * replaced audio sender track, …) must be re-asserted by the app.
+ */
+export interface SessionReconnectedEventDetail {
+    /** The attempt number that got through (1-based). */
+    attempt: number;
+}
```
FabienDanieau: A named exported interface for `{ attempt: number }`. Inlining it in the event map costs nothing and removes a name consumers have to learn.

(`SessionReconnectingEventDetail` above carries three fields and earns its name — this one doesn't.)

<sub>_Complexity pass — no bug here, purely about the diff getting shorter._</sub>

ts/scripts/gen-version.mjs
tfrere: good catch, the claim was indeed false. fixed in 68fc30a77 with your third option: a workflow step right after the bump reruns gen:version (which finally has a caller) and recompiles the SDK, so dist/version.js ships the real version. dropped the typeof check too.

ts/lib/session-supervisor.ts
tfrere: fixed in 340299a2e, took your suggestion as-is: the fatal path now stops the watchdog before emitting, so the error fires once and the next startDcWatchdog() re-arms. stale comment went with it.

ts/lib/dc-watchdog.test.ts
tfrere: done - the test now advances a full extra escalation cycle (10s) after the error and asserts it fired exactly once. it does fail against the old code, kept as the regression guard.

ts/lib/central-listener.ts
tfrere: guard added in 340299a2e, plus a comment explaining the EOF + in-flight heartbeat race so it doesn't get simplified away later. skipped the runStream() re-entrancy guard for now since scheduleReconnect is the only path that arms it.

ts/lib/types.ts
tfrere: fair. PR description now says the *runtime* surface is additive and spells out the type-surface caveat (interface implementers, i.e. downstream mocks, need a tsc run).

ts/lib/types.ts
tfrere: called out in the PR description now (behavioural change, opt-out spelled out). keeping the default on for this release: published Space embeds freeze their own SDK bundle so they keep the old semantics, and the consumers that pick this version up (host, mobile app) are ours and ship with it.

ts/lib/session-supervisor.ts
tfrere: documented on the public surface in 340299a2e (types.ts + reachy-mini.ts): aborting an in-flight redial leaves the session torn down and the caller owns the terminal event. kept the no-emit behaviour since the update flow relies on it and an unsolicited sessionStopped there would flip the update gate.

ts/lib/reachy-mini.ts
tfrere: all three in 48fb6eca0: send now happens first (replies arrive on a later task) so a failed send rejects like _sendCommandAwaitCompletion; the fail-open detects a BroadcastTimeoutError sentinel instead of the message string; and the identity .then() is gone (the catch already widens the type).

ts/lib/reachy-mini.ts
tfrere: done in 48fb6eca0, took the URL API form.

ts/lib/token-store.ts
tfrere: done in 48fb6eca0 - gated on the four codes a prompt=none round trip can return, so an app's own ?error= is left alone.

ts/lib/types.ts
tfrere: done, _pc is @deprecated pointing at peerConnection.

ts/vitest.config.ts
tfrere: intentional, yes - the stacked host PR brings the host/src tests. merge order is safe either way: vitest is fine with a glob matching nothing, and npm test stays green.

tfrere:
Thanks for the pass, all of it was on point. Pushed three commits:

- 68fc30a77 fixes the version stamping (your third option - gen:version rerun + SDK rebuild right after the bump)
- 340299a2e fixes the watchdog error loop (your suggestion as-is, test extended to 10s as the regression guard), adds the scheduleReconnect re-entrancy guard, and documents the setAutoReconnect mid-redial edge on the public surface
- 48fb6eca0 takes the smaller inline items: request() now rejects on a failed send and fail-opens via a BroadcastTimeoutError sentinel instead of the message string, login() builds the prompt param with the URL API, consumeOAuthErrorParams() is gated on the four silent-flow error codes, and _pc is @deprecated in favour of peerConnection

PR description is updated too: the additive claim is scoped to the runtime surface with the type-surface caveat spelled out, and the autoReconnect default is called out as a behavioural change.

On the complexity pass: agreed on _closeTransport, it's the same argument settleAll makes and I want it - but it touches every teardown path so I'd rather land it as a follow-up than re-churn this diff under review. Planning to take the rest of the pass (comment trimming included), the prettier/indentation sweep, and the null-token teardown test gap in that same follow-up.

ts/lib/central-listener.ts
tfrere: trimmed in 1c4443f2: kept the why-not-the-SDK-class paragraph and the Authorization-header line, dropped the wire-surface table and the transport/reconnect essays.

ts/lib/central-listener.ts
tfrere: took your shape in 1c4443f2 - clamp and clampHeartbeatIntervalMs are gone, negotiation is recommended seconds, lease/3 fallback, default.

ts/lib/central-listener.ts
tfrere: done in 1c4443f2: onList now takes the producers array directly and CentralListEvent is gone (barrel export too). the stacked host PR gets a one-line fixup.

ts/lib/central-listener.ts
tfrere: cast all three in 1c4443f2, with a short comment recording the first-party trust level. activeApp is optional now since the cast no longer coerces a missing field to null.

ts/lib/session-supervisor.ts
tfrere: trimmed in 1c4443f2 - kept the RFC 7675 ceiling derivation and the failed->connected debounce rationale, the rest is one short block per constant group.

ts/lib/session-supervisor.ts
tfrere: took the judgement call in 1c4443f2: one typed emit(name, detail) dep. the supervisor keeps a tiny private _emitError wrapper so the eight error sites don't each spell the webrtc payload shape; the mock factory is down to one vi.fn() and the asserts filter on the event name.

ts/lib/session-supervisor.ts
tfrere: done in 1c4443f2 - the dial dep points at a private _startSessionInternal, the public startSession is cancelRedial() + that, and noteExternalDial / the internal-dial flag / the try-finally are gone. the supersede test now drives the real public startSession.

ts/lib/session-supervisor.ts
tfrere: dropped in 1c4443f2. one nuance kept: the timer callback nulls _redialWake only if it still owns it, so a post-cancel fire can't orphan a newer sleep's wake.

ts/lib/session-supervisor.ts
tfrere: took the Promise.race form as-is in 1c4443f2.

ts/lib/session-supervisor.ts
tfrere: agreed nothing SDK-internal rides them, but networkOnline/Offline/Change are on the released event surface (moved, not added, as you noted), so pulling them is a semver-major move rather than a line cut. leaving the block in place and flagging the events for a deprecation follow-up.

ts/lib/reachy-mini.ts
tfrere: landed it in this PR after all (1c4443f2) rather than the follow-up - it fell out naturally while taking the rest of the pass. _closeTransport() holds the common sequence (ledger settle, resolver reject, resilience plumbing, timers, mic, pc/dc, flags) and returns the session id; each caller keeps its settling error, its endSession policy, its events and where _state lands. the four copies are gone.

ts/lib/pending-replies.ts
tfrere: done in 1c4443f2, exactly that shape. rpcCall loses the id juggling, and a send against a closed channel now rejects inside the ledger, same as slotRoundtrip owns its timeout.

ts/lib/types.ts
tfrere: inlined in 1c4443f2 (the doc moved onto the event-map entry); SessionReconnectingEventDetail keeps its name.

tfrere:
Follow-up landed in this PR after all instead of a separate one - two commits:

- [1c4443f2](https://github.com/pollen-robotics/reachy_mini/commit/1c4443f239fc9006e84e1e1b1b26313f792c967a) takes the complexity pass (every anchor answered inline): _closeTransport() unifies the four teardown copies, rpcRoundtrip() replaces the two-call RPC API, the supervisor loses noteExternalDial/_redialTimer and its seven emit deps collapse into one typed emit(), withTimeout is a Promise.race, central-listener drops the heartbeat clamp / the CentralListEvent wrapper / the hand-coercion, and the two big prologues are trimmed to the parts that don't restate the code. Also covers the null-token teardown test gap you flagged. ~-130 lines, 88 tests green, tsc clean.
- [78026b86](https://github.com/pollen-robotics/reachy_mini/commit/78026b86320f2886e7411611d6707bf9c718d9b1) is the indentation sweep (the three 2-space holdouts move to 4 spaces, whitespace-only - `git diff -w` is empty).

One decline, reasoned inline: the network-listener block stays, since networkOnline/Offline/Change are on the released event surface and pulling them is a semver-major move - flagged for a deprecation follow-up.

Heads-up for the stacked host PR: onList now receives the producers array directly, so its useRobots callback needs a one-line fixup there.

FabienDanieau (approved):
Approving. Basis, so it's on record: I reviewed the resilience path in the diff (`session-supervisor.ts`, the re-dial loop, `ontrack`, `_teardownForRedial`) and exercised this SDK **extensively on hardware** via the #1321 branch — ICE grace, the re-dial loop, the DC-silence watchdog and the central listener, across three blip shapes and repeated cycles. It holds up: sub-grace flickers stay silent, >3 s outages re-dial and recover on attempt 1, and the slot-freeing `endSession` in `_teardownForRedial` does its job.

I have *not* read the token store, the pending-replies refactor, `getImu`, `DoA`, `request()`, the version tooling or the CI changes.

Two follow-ups, neither blocking:

**1. `ontrack` announces video only** (`reachy-mini.ts:656-668`) — no audio equivalent. Combined with `autoReconnect` defaulting on, the SDK silently swaps the `RTCPeerConnection` under every consumer and re-announces half the media. Audio consumers are left holding a receiver from a closed PC, and it fails *silently*: `onaudioprocess` keeps firing on the dead track, so the uplink looks healthy while the backend hears nothing.

This is not hypothetical — it cost a full debugging session on the mobile client and is one of four bugs fixed in pollen-robotics/reachy_mini_mobile_app#85. The host embed only escapes it because it separately clears its media cache on `sessionReconnecting`. An `audioTrack` event (or a `tracksChanged` on re-dial) would make it hard to get wrong instead of something every consumer must remember.

It's additive, so it *can* land later — but the SDK is about to be published and this contract has already cut one consumer. Your call whether that argues for doing it here.

**2. The re-dial reuses the stored peer id.** `dial: (robotId) => this._startSessionInternal(robotId)` (`:238`), no re-resolution. Meanwhile #1316's embed re-resolves right before `startSession` *because central rotates peer ids on relay reconnect*, and the mobile client added its own `resolvePeerId` for the same reason. Two consumers independently working around the SDK suggests it belongs in the loop. Concretely: when the **robot's** link drops (its Wi-Fi, an AP reboot) central re-registers it under a fresh id, so all 4 attempts dial a corpse for ~22 s before giving up.

**Cross-cutting:** the validation here covers "does it reconnect", not "does audio still flow after it reconnects". One integration test that re-dials and then asserts a live receiver track would have caught (1) and most of the mobile-side fallout.

tfrere:
Heads-up for re-review: three new commits since approval - the daemon wake/sleep backstop (reset_to_sleep + tightened idle-reset predicate + per-motor torque override fix), the ensureAwake behavior change (awaits the wake trajectory, clears inherited gravity compensation), and a new guide section on background-tab throttling. Details in the description addendum. Everything else is untouched; #1316/#1321 have been rebased on top.

PR #1316, opened 2026-08-03
feat(host): update gate, resilience UX, picker/boot refresh

tfrere:
Split out of #1304 (see the split plan there). This is the host half of the former JS block (#1314, closed), **stacked on #1315** (the SDK half) - the base is `feat/sdk-js-core` so the diff below is host-only; it will retarget to `main` automatically when #1315 merges.

## Embed (app-side iframe runtime)

- **Reconnect bridge**: mirrors the SDK's automatic re-dial into host app-states - full-screen "Reconnecting…" while the SDK retries, back to `live` on success, fatal error card on terminal failure. Registered with plain string event names, so an app that bundled an older SDK is a silent no-op.
- **Sleep-on-leave (mobile parity)**: the embed plays `gotoSleep` + forces motors `disabled` while the session is still up, waits (bounded 1 s) for the state frame echoing `disabled`, then acks `embed:left`; the host waits for the ack instead of a fixed sleep (pairs with #1311 so the daemon doesn't replay its own goto_sleep after the handoff).
- **Stable-id dialing**: central rotates peer ids on relay reconnect, so the embed re-resolves the live peer id right before `startSession()`, with the daemon `hardware_id` travelling through both handoff channels. Fail-open on every miss. Also `awaitFullyAwake()` (bounded 8 s) instead of fire-and-forget `ensureAwake()`.
- **`sdkVersion` reporting** on `embed:ready` (telemetry only - the staleness overlay initially built on it was dropped in review, as a de-facto kill switch for published Spaces) and RTT sampling that re-reads the live `RTCPeerConnection` (the re-dial replaces it mid-session).

## Shell

- **Daemon update gate**: blocks entry into an app while the robot's daemon is out of date, drives the update, and reboot completion now requires an offline observation first (pairs with #1312, which makes the failure path honest). A host-side stall timeout disarms the embed (`host:cancel-update`), and the blocking tier keeps an exit back to the picker.
- **Picker / top bar refresh**: SSE-driven list on the SDK's `openCentralListener` (drops the host's local listener copy), spinner instead of the Refresh button, identity row + direct logout, outlined "End session".
- **Flicker-free boot + error boundary**: `authResolved` gate (with an 8 s escape hatch when the SDK bundle never loads), single continuous splash across the boot legs, neutral splash before the auth state is known, `HostErrorBoundary` with a themed fallback.
- **Silent OAuth sign-in**: prompt-less `authorize_url` retry on boot for returning users.

## Token lifetime posture (cross-PR dependency)

This PR drops the `pagehide` token wipe (it fired on every F5, not just tab close), and the guide recommends `hf_oauth_expiration_minutes: 43200` (30 days) to app authors. The mitigation lives in the stacked #1315: the SDK's token store enforces a **24 h sliding idle window** on the sessionStorage-scoped token, so the realistic exposure is a still-open tab used within the last day - a resurrected session-restore tab re-auths. The other bound is scope: the recommended setup keeps the default `openid profile`, so the token identifies the user but grants no Hub/API access. Both bounds are now documented in `APP_CREATION_GUIDE.md` next to the expiration recommendation.

## Tests

Wire-compat (`useHostBridge.compat`), version-policy (`daemonRelease`), protocol, reboot-watch and update-gate suites; full `ts/` suite at 155 with the SDK half, `tsc --noEmit` clean on both packages. Validated on hardware and in the full-software sim (daemon mockup + local central + host harness): dead-transport re-dial, backgrounding, daemon update, handoffs.


ts/host/src/hooks/useOAuth.ts
```diff
@@ -64,6 +104,13 @@ export function useOAuth(sdk: ReachyMiniInstance | null): OAuthState {
   const [isPostOauthReturn, setPostOauth] = useState<boolean>(() =>
     readOAuthPendingOnce(),
   );
+  // Whether the boot-time auth resolution has settled. Starts true only when
+  // there's nothing async to wait for: the SDK already reports authenticated,
+  // or the user explicitly signed out earlier. Otherwise we must wait for
+  // `authenticate()` before deciding between SignInView and the picker.
+  const [authResolved, setAuthResolved] = useState<boolean>(() =>
```
FabienDanieau, line 128: **The boot splash has no escape hatch when the SDK never loads.**

`useSdk` leaves `sdk === null` until `window.ReachyMini` exists, and the effect below opens with `if (!sdk) return;` — so `authResolved` never flips. In `ReachyHostShell` that makes `showBootSplash` (`hostPhase === 'signing-in' && !authResolved`) hold the neutral spinner forever.

Before this PR the same state rendered `SignInView`, including the `isLocalDevMissingConfig` hint — which can now never be reached. Any failure to load the SDK bundle (blocked CDN, stale Space asset, bad build) becomes a permanent spinner instead of a diagnosable screen.

Suggest gating the splash on `sdkReady` too, or resolving `authResolved` after a bounded wait.

ts/host/src/components/DaemonUpdateGate.tsx
```diff
@@ -0,0 +1,460 @@
...
+    }
+    if (progress.status === 'failed') {
+      setFailure(progress.error ?? null);
+      setPhase('failed');
+      return;
+    }
+    // `rebooting` / `done`: the install is out of our hands. Drop the
+    // iframe so the shell can resume central polling - that's the only
+    // way we get to see the robot come back.
+    setPhase('rebooting');
```
FabienDanieau, line 163: **A host-side update timeout leaves the embed armed, and a later end-session resurrects this gate.**

`UPDATE_STALL_TIMEOUT_MS` flips *local* phase to `failed`, but nothing tells the embed: `startDaemonUpdateForHost` only calls `abandonUpdate()` on a daemon-reported `failed`. So `autoReconnect` stays off and the `sessionStopped → embed:update-progress{rebooting}` translator stays registered.

Repro: daemon acks `start_update` then goes silent → host times out at 180 s → user dismisses (`clearedRef = true`, `updateProgress = null`) and keeps using the app → they end the session normally → embed posts `rebooting` → this effect runs with `sessionLostRef` still `false` → full-screen "Reachy is rebooting…" pops over the picker and `onSessionLost()` fires mid-leave.

Cheapest fix: bail out of this effect when `phase === 'failed'`. Better: a `host:cancel-update`, or an embed-side self-disarm on its own timer.

ts/host/src/components/DaemonUpdateGate.tsx
```diff
@@ -0,0 +1,460 @@
...
+
+  return (
+    <Box
+      sx={{
+        position: 'fixed',
+        inset: 0,
+        // Above the connecting / leaving overlays (1300): once an update
+        // is running, its narrative outranks the reconnection churn it
+        // is about to cause.
+        zIndex: 1400,
```
FabienDanieau, line 244: **The blocking tier is a dead end.**

`position: fixed; inset: 0; zIndex: 1400` sits above the TopBar, so a user on a pre-1.8.2 daemon gets two outbound links and nothing else — no back-to-picker, no sign out, no end session.

And `daemonVersion` only arrives *after* `startSession()` succeeds, so the session is live and the robot is awake behind the block. "The app never becomes interactive" is true visually only: the app-slot lock stays held with no release path short of closing the tab.

Worth adding a "Back to my Reachies" action wired to `endSession`.

ts/host/src/components/ReachyHostShell.tsx
```diff
@@ -385,20 +522,20 @@ function ReachyHostShellNormal({
...
         name: picked?.meta?.name ?? null,
         transport: picked?.transport ?? null,
+        hardwareId: picked?.hardwareId ?? null,
       });
       initSentForRef.current = null;
       embedReadyPendingRef.current = false;
+      setDaemonVersion(null);
+      setUpdateProgress(null);
+      setAwaitingReboot(null);
+      setGateKey((k) => k + 1);
```
FabienDanieau, line 526: **`rebootWatch` isn't reset here, which breaks the invariant documented at its declaration.**

The comment on `rebootWatch` states it's "Held at `waiting-offline` whenever `awaitingReboot` is null", but only `dismissDaemonUpdate` resets it — `selectRobot` clears `awaitingReboot` and bumps `gateKey` while leaving the watch wherever it was.

A pre-armed `waiting-online` would let a *stale* pre-reboot listing complete a subsequent watch in a single observation: exactly the regression `rebootWatch.ts` exists to prevent. I couldn't build a reachable path today (the `rebooting` overlay blocks selection), but one `setRebootWatch('waiting-offline')` here makes the invariant real instead of incidental.

ts/host/src/components/DaemonUpdateGate.tsx
```diff
@@ -0,0 +1,460 @@
...
+/** Where to get the desktop app, which can update a daemon too old for
+ *  OTA. Same target the mobile gate uses. */
+const DESKTOP_APP_URL = 'https://huggingface.co/spaces/pollen-robotics/reachy-mini';
+
+/** Public troubleshooting docs. */
+const TROUBLESHOOTING_URL = 'https://huggingface.co/docs/reachy_mini/troubleshooting';
+
+/** If the daemon acked the job but never restarted within this window,
+ *  something failed silently - stop pretending it's still working. */
+const UPDATE_STALL_TIMEOUT_MS = 180_000;
```
FabienDanieau, line 47: **This is a budget, not a stall timer.**

The effect that consumes it depends on `[phase]` only, so it never resets on `in_progress` lines — it's a hard 180 s ceiling on the whole install. A slow-but-progressing pip download on weak wifi trips it, and the copy ("The robot never started the update.") is then wrong.

Either reset the timer on each progress line, or rename to `UPDATE_BUDGET_MS` and reword the failure text.

ts/host/src/components/ReachyHostShell.tsx
```diff
@@ -51,7 +64,29 @@ import { SignInView } from './SignInView';
 import { TopBar, type HostPhase } from './TopBar';
 import { WelcomeBackOverlay } from './WelcomeBackOverlay';
 
-const LEAVING_TIMEOUT_MS = 1500;
+// Hard cap on how long the host stays on the leaving overlay waiting for the
+// embed's `embed:left` ack (app cleanup + host-owned sleep/disable). Larger
+// than the embed's own sleep hard-cap (~6.5 s) so a well-behaved embed's ack
+// wins the race; a wedged/older embed that never acks just falls through here.
+const LEAVING_ACK_CAP_MS = 8000;
```
FabienDanieau: **The margin is thinner than the comment claims.**

The embed's worst case isn't ~6.5 s: it's `LEAVE_SLEEP_HARD_TIMEOUT_MS` (6500) **+** `LEAVE_DISABLE_CONFIRM_TIMEOUT_MS` (1000) = 7.5 s, because `awaitMotorsDisabled` runs after the sleep race. That leaves 500 ms for postMessage + React commit before this cap fires and unmounts a well-behaved embed mid-ack.

Bump to ~9–10 s.

FabienDanieau (commented):
## Review: host shell refresh

Read the full diff. Two passes below — correctness first, then a complexity pass on the same code. Inline comments carry the detail; this is the map.

### Verified plumbing

- `openCentralListener` in the SDK (`ts/lib/central-listener.ts`, base branch) **does** retain the heartbeat negotiation and the HTTP-4xx→abort eviction fix from the deleted host copy, and now takes `appName`. Dropping `host/src/lib/centralListener.ts` is a safe 357-line dedup, not a silent regression of the "picker never sees `busy=false`" bug.
- `hardwareId` is mapped in both the SSE (`producerToRobotInfo`) and REST (`centralRest.ts`) paths, and exists on `RobotInfo` in `ts/lib/types.ts` — stable-id dialing is actually wired end to end.

### Correctness — ranked

1. **Boot splash can hang forever** when the SDK bundle never loads (`sdk === null` → `authResolved` never flips → neutral spinner, no `SignInView`, no `isLocalDevMissingConfig` hint). `useOAuth.ts`
2. **A host-side update timeout leaves the embed armed**; a later normal end-session resurrects the gate into a bogus full-screen "rebooting" and fires `onSessionLost` mid-leave. `DaemonUpdateGate.tsx`
3. **The blocking tier is a dead end** — `zIndex: 1400` covers the TopBar, and the session is live behind it with no release path. `DaemonUpdateGate.tsx`
4. **`rebootWatch`'s documented invariant isn't maintained** by `selectRobot`. `ReachyHostShell.tsx`
5. `UPDATE_STALL_TIMEOUT_MS` is a hard budget, not a stall timer, and the failure copy is wrong for a slow install.
6. `LEAVING_ACK_CAP_MS` margin is 500 ms, not the ~1.5 s the comment implies (6500 + 1000).

Plus: the SDK-staleness modal is the "global kill switch" this PR argues against, applied to apps; and the token posture change leans entirely on #1315's idle window.

### Tests

The pure logic is genuinely well covered, and the golden-envelope suite is the right instinct — asserting the *outbound* direction against a frozen 1.9.0 `isProtocolMessage` is the half everyone forgets.

The gap is that coverage is inverted relative to risk: `DaemonUpdateGate` — three timers, two refs-as-latches, phase derived during render, and the source of findings 2, 3 and 5 — is the one new stateful component with no tests, while the pure functions around it have three suites. A handful of fake-timer tests over `progress → phase` would catch #2 directly.

### Complexity pass

Same diff, complexity only. Biggest cuts: the `RefreshIndicator` timer/fade machinery (−75), `rebootWatch`'s state machine collapsing to one boolean (−85), `DaemonUpdateGate`'s log panel + copy helpers + button wrapper (−75), the imperative-DOM modal (−45), and `daemonRelease`'s memo layer over a synchronous localStorage read (−25).

**net: −470 lines possible**, −600 if the `notice` tier goes and the four new "something's wrong + one button" full-screens (`HostErrorBoundary` fallback, `ErrorView`, the gate's `failed` state, `showSdkOutdatedOverlay`) become one `FullScreenNotice`.

### Nice work

`vite.harness.config.ts`'s `envDir` pin is a real token leak fixed, and the comment says *how* it was found. `resolveLivePeerId` failing open on every miss, `daemonUpdateSeverity` refusing to act on any version it can't parse, and `readDaemonVersion` degrading to `null` are the right instincts for code that gates someone's own hardware. Comment density is high, but the good ones carry non-obvious *why* — peer-id rotation, central's stale listing, the SCTP send-queue race behind `awaitMotorsDisabled` — rather than restating the code.

ts/host/src/embed/index.ts
```diff
@@ -667,9 +1057,113 @@ function pushAppState(
...
+ * this very code - a frozen bundle can't warn about itself. Only a
+ * parent that updates independently of the app could catch that (the
+ * mobile app, via the absence of `sdkVersion` on app-state); the web
+ * shell can't, being part of the same frozen bundle.
+ *
+ * Silent when: the version is a dev placeholder (`0.0.0-*`) or
+ * unparseable, the latest release can't be fetched, or we're simply up
+ * to date. Never blocks - one click dismisses it for the session.
+ */
+async function maybeWarnSdkOutdated(): Promise<void> {
```
FabienDanieau: **This is the global kill switch the PR argues against, aimed at apps instead of daemons.**

The daemon policy is deliberately two-tiered because "a Space is public: applying that rule here would make every daemon release a global kill switch." This does exactly that for apps: any patch release of the SDK makes every already-published Space (frozen bundles, never rebuilt) show a full-screen scrim to end users, with no persistence — back on every reload.

It also compares the **npm bundle version** against the **`pollen-robotics/reachy_mini` GitHub release tag** (`fetchLatestDaemonVersion` reused verbatim). If those two version lines ever diverge, every app warns spuriously.

Suggest minor-version tolerance, a `localStorage` dismissal, or a non-modal notice — the same reasoning you applied one tier up. (The card is also hardcoded light: `background:#fff` inside apps that may be dark.)

ts/host/src/components/ReachyHostShell.tsx
```diff
@@ -481,10 +690,17 @@ function ReachyHostShellNormal({
     const onPageHide = (): void => {
       // Best-effort: tell the embed it's leaving so it can
       // disconnect its SDK. We don't wait for an ack.
+      //
+      // Deliberately NOT wiping the HF token here anymore: pagehide
```
FabienDanieau, line 708: **Worth stating the cross-PR dependency explicitly.**

Dropping the wipe is well reasoned (F5 ≠ tab close), but combined with the guide now *recommending* `hf_oauth_expiration_minutes: 43200` to every app author, an HF user token lives in `sessionStorage` for the tab's lifetime and travels to the iframe in the URL hash.

The entire mitigation is the sliding idle window in the stacked #1315. Please call that out in the PR body, and confirm the window is short (hours, not days) — a 24 h idle window is a long tail for a 30-day token.

ts/host/src/hooks/useOAuth.ts
```diff
@@ -81,16 +128,44 @@ export function useOAuth(sdk: ReachyMiniInstance | null): OAuthState {
...
     let alive = true;
+    // True once we've committed to the silent-auth redirect: the page is
+    // about to unload, so `authResolved` must stay false to keep the
+    // neutral splash up instead of flashing SignInView for a frame.
+    let redirecting = false;
     void (async () => {
       try {
         const ok = await sdk.authenticate();
         if (!alive) return;
+        if (!ok && shouldAttemptSilentSignIn()) {
```
FabienDanieau, line 175: **Silent `prompt: 'none'` now bounces every anonymous first-time visitor through huggingface.co.**

No prior grant → `consent_required` → return with `?error=`, before they ever see the sign-in view. The four guards (signed-out, one-per-tab, dev token, iframe) look right and re-arming on explicit click is a nice touch, but three things to confirm:

1. deep-link query params (`?config=`, etc.) survive the round trip via `redirect_uri`;
2. crawlers / social-preview bots aren't redirected;
3. `authenticate()` really does strip `?error=` as the comment asserts.

ts/host/src/embed/index.ts
```diff
@@ -332,16 +472,38 @@ async function bootOnce(
     throw err;
   }
 
+  // Read the daemon version now that the data channel is up, and before
+  // the wake, so a host that gates on it (the web shell's update gate)
+  // can decide while the user is still on the connecting splash rather
+  // than after the app has painted. Fail-open: stays `null` on timeout.
+  daemonVersion = await readDaemonVersion(sdk);
```
FabienDanieau, line 449: **Boot latency budget, for the record.**

`readDaemonVersion` (≤2.5 s) + `awaitFullyAwake` (≤1 s state wait + ≤8 s trajectory) is up to ~11.5 s added before `live` in the degraded case. Deliberate and defensible — just confirm ConnectingView's ~6 s slow hint stays honest across the whole window.

ts/host/src/components/PickerView.tsx
```diff
@@ -245,6 +281,123 @@ function RobotsHeader({
   );
 }
 
+/* ─────────────────── Refresh activity indicator ─────────────────── */
+
+// Floor we keep the indicator on screen once a refresh starts, so an
+// instant (cached) fetch still shows a complete, smooth turn instead of
+// a one-frame flash that reads as "cut off".
+const MIN_VISIBLE_MS = 900;
```
FabienDanieau: `delete:` 110 lines of `MIN_VISIBLE_MS` / `TAIL_DEBOUNCE_MS` / `mounted`+`shown` double state / `requestAnimationFrame` to stop a spinner from flashing.

An always-mounted icon does it in one element:

```tsx
<RefreshIcon sx={{
  opacity: isRefreshing ? REST_OPACITY : 0,
  transition: 'opacity 320ms ease',
  animation: `${spin} 1.4s linear infinite`,
}} />
```

The 320 ms fade-out already coalesces a two-fetch burst, which is what the trailing debounce was for. **−~75 lines.**

Minor a11y note while here: `aria-label` on an MUI SVG with no `role` isn't reliably announced. The glyph is decorative — `aria-hidden` throughout is more honest than `aria-hidden={!visible}` plus a label.

ts/host/src/lib/rebootWatch.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/** What the shell remembers about the robot it dropped mid-update. */
+export interface RebootTarget {
+  /** Stable hardware id (serial), the strongest match. */
+  hardwareId: string | null;
+  /** Advertised name, fallback for daemons without a hardware id. */
+  name: string | null;
+}
+
+export type RebootWatchPhase = 'waiting-offline' | 'waiting-online' | 'back';
```
FabienDanieau: `shrink:` a 3-state machine plus a switch for what is one sticky boolean.

In the shell: `sawOffline` state, `if (!listed) setSawOffline(true)`, `rebootedRobotBack = sawOffline && listed`. That deletes this module and most of its test file. `back`'s terminality is unused — the gate latches `done` on the edge and never re-reads.

**−~85 lines** (module + test). The test's stale-listing case is worth keeping as an assertion on the boolean.

ts/host/src/lib/daemonRelease.ts
```diff
@@ -0,0 +1,205 @@
...
+  const min = parseSemver(MIN_SUPPORTED_DAEMON_VERSION);
+  return !!c && !!min && compareSemver(c, min) >= 0;
+}
+
+interface CacheEntry {
+  value: string | null;
+  at: number;
+}
+
+let memo: CacheEntry | null = null;
```
FabienDanieau: `delete:` `memo` + `isFresh(memo)` + `cachedLatestDaemonVersion` are a second cache in front of a *synchronous* localStorage read.

Drop all three and read localStorage directly: the first call writes before the second reads, so the fetch-called-once test still passes. It also removes the `useState<string|null>(cachedLatestDaemonVersion)` puzzle in `useLatestDaemonVersion` (that's the lazy-initializer overload, correct but it reads like a missing `()`).

**−~25 lines.**

ts/host/src/embed/index.ts
```diff
@@ -667,9 +1057,113 @@ function pushAppState(
...
+    showSdkOutdatedOverlay(appSdkVersion as string, latest as string);
+  } catch {
+    /* purely advisory - never let it interfere with a live session */
+  }
+}
+
+/** Plain-DOM overlay (the embed is framework-agnostic): dark scrim,
+ *  centred card, one dismiss button. Styling is self-contained so it
+ *  renders the same over any app theme. */
+function showSdkOutdatedOverlay(current: string, latest: string): void {
```
FabienDanieau: `shrink:` seven `createElement` calls and seven `cssText` arrays for one modal. One `document.body.insertAdjacentHTML` with a template literal and a `<style>` block is ~15 lines. **−~45.**

ts/host/src/components/DaemonUpdateGate.tsx
```diff
@@ -0,0 +1,460 @@
...
+  latestVersion,
+  progress,
+  robotBackOnline,
+  appLive,
+  onStartUpdate,
+  onSessionLost,
+  onDismiss,
+}: DaemonUpdateGateProps): JSX.Element | null {
+  const [phase, setPhase] = useState<Phase>('idle');
+  const [logLines, setLogLines] = useState<string[]>([]);
```
FabienDanieau: `delete:` `logLines` + `LOG_TAIL` + `InstallLog` — six lines of pip output nobody acts on. The spinner and "Installing the latest software" already carry the message, and the `in_progress` branch collapses to a no-op. **−~35 lines.**

ts/host/src/components/DaemonUpdateGate.tsx
```diff
@@ -0,0 +1,460 @@
...
+        borderRadius: `${RADIUS.pill}px`,
+        px: 3,
+      }}
+    >
+      {children}
+    </Button>
+  );
+}
+
+function PhaseIcon({ phase }: { phase: Phase }): JSX.Element {
```
FabienDanieau: `shrink:` `PhaseIcon` + `titleFor` + `bodyFor` are three switches over the same 6-value enum. One `const COPY: Record<Phase, { icon, title, body }>`, indexed once.

Same block: `PrimaryButton` is a wrapper for two call sites that differ only in label — a `PRIMARY_SX` const on `<Button variant="contained">` does it. **−~40 lines.**

ts/host/src/components/DaemonUpdateGate.tsx
```diff
@@ -0,0 +1,460 @@
+/**
+ * Daemon version gate for the standalone web shell.
```
FabienDanieau: `shrink:` 33 lines of essay before the first import, three-quarters of it restating the two-tier policy that's already in `daemonRelease.ts`'s header *and* in APP_CREATION_GUIDE §13.6.

Keep the "why the update outlives the iframe" paragraph — that one is genuinely non-obvious. Cut the rest.

ts/host/src/embed/index.ts
```diff
@@ -918,6 +1433,64 @@ function createRobotMedia(sdk: ReachyMiniInstance): RobotMedia {
...
+ * uses. On success the SDK re-fires `streaming` (video re-attaches on
+ * its own) and we flip back to `live`; on a terminal
+ * `sessionStopped { reason: 'reconnect_failed' }` we report a fatal
+ * error so the shell offers reload / back-to-picker.
+ *
+ * Registered with plain string event names so an app that shipped an
+ * OLDER SDK bundle (no such events) is a silent no-op.
+ */
+function installReconnectBridge(sdk: ReachyMiniInstance): void {
+  const on = (
```
FabienDanieau: `delete:` the `on()` indirection and its try/catch. `addEventListener` with an unknown event name never throws on an EventTarget — it just never fires. Call `sdk.addEventListener` directly, three times.

Same dead defensiveness in the `try { sdk.addEventListener('sessionStopped', …) } catch` inside `startDaemonUpdateForHost`. (The `setAutoReconnect` feature-detect *is* warranted — the embed rides whatever SDK the app bundled.) **−~15 lines.**

ts/host/src/components/PostOAuthSplash.tsx
```diff
@@ -49,7 +68,17 @@ const pulseKeyframes = keyframes`
   }
 `;
 
-export function PostOAuthSplash(): JSX.Element {
+export interface PostOAuthSplashProps {
```
FabienDanieau: `yagni:` a `variant` prop whose `neutral` branch renders one `CircularProgress`, threaded through four `variant === 'signing-in' &&` guards.

The shared frame is `position:fixed; inset:0` plus centering — inline a 6-line neutral splash at the call site instead. **−~25 lines.**

ts/host/src/components/HostErrorBoundary.tsx
```diff
@@ -0,0 +1,116 @@
...
+  };
+
+  render(): ReactNode {
+    const { error } = this.state;
+    if (!error) return this.props.children;
+
+    return (
+      <Box
+        sx={{
+          minHeight: '100%',
```
FabienDanieau, line 63: `shrink:` 58 lines of MUI `sx` for a screen that says "reload". Emoji + `Typography` + `Button` in a centered Box is ~15; the `grayscale(0.4)` filter and the monospace font stack on the `<pre>` aren't earning their keep.

The boundary itself is right (React needs the class) — this is just the fallback chrome. **−~30 lines.**

ts/host/src/components/ReachyHostShell.tsx
```diff
@@ -410,26 +547,37 @@ function ReachyHostShellNormal({
...
       }
       setHostPhase('leaving');
 
-      await sleep(LEAVING_TIMEOUT_MS);
+      // Wait for the embed to finish its host-owned tear-down (app onLeave +
+      // gotoSleep + motors disabled) and ack via `embed:left`, so the robot is
+      // actually asleep before we drop the session. Bounded by
+      // `LEAVING_ACK_CAP_MS` so a wedged or older embed (which never acks)
+      // still falls through and unmounts.
+      await new Promise<void>((resolve) => {
```
FabienDanieau: `shrink:` a settled flag, a timer handle and ref-nulling around one ack:

```ts
await Promise.race([
  new Promise<void>((r) => { leftAckResolveRef.current = r }),
  sleep(LEAVING_ACK_CAP_MS),
]);
leftAckResolveRef.current = null;
```

`race` ignores the loser, so the double-settle guard is dead weight. Needs the `sleep` helper this PR deletes three hunks down. **−8 lines.**

ts/host/src/components/ReachyHostShell.tsx
```diff
@@ -190,6 +251,22 @@ function ReachyHostShellNormal({
    *  reports authenticated) so the celebratory beat plays out
    *  even on a fast auth confirmation. */
   const [welcomeBackShown, setWelcomeBackShown] = useState<boolean>(false);
+  /** Latches true once the WelcomeBackOverlay has finished (its
+   *  `onDone`). Needed because `welcomeBackShown` flips back to
+   *  false on `onDone` while `isPostOauthReturn` stays true in the
+   *  real-OAuth path - without this the post-OAuth splash would
+   *  re-trigger in `picking` and loop "Signing you in…" forever.
+   *  Reset on sign-out so a re-sign-in replays the sequence. */
+  const [welcomeBackDone, setWelcomeBackDone] = useState<boolean>(false);
```
FabienDanieau: `shrink:` `welcomeBackShown` + `welcomeBackDone` + `welcomeBackShownOnceRef` are three latches over one three-position sequence.

One `welcomeBack: 'idle' | 'showing' | 'done'` removes a `useState`, a ref, and a clause from `showPostOAuthSplash`. **−~15 lines.**

ts/host/src/hooks/useRobots.ts
```diff
@@ -198,7 +211,19 @@ export function useRobots(opts: {
     void doFetch();
   }, [doFetch, enabled, hfToken]);
 
-  return { robots, isLoading, isRefreshing, error, refresh };
+  // Report "loading" for the whole first-fetch window of an `enabled`
+  // session - including the transition frame before the fetch effect has run -
+  // so the picker never flashes its empty state between `picking` and the
+  // first central response. Once settled, defer to the real `isLoading`
+  // (which stays false on background polls, driving only the refresh spinner).
+  const isInitialLoading = enabled && !firstFetchSettled;
```
FabienDanieau, line 215: `shrink:` `isLoading` is `isRefreshing && !hasRobots` at every call site — drop the prop and PickerView's re-derivation of it. Keep `firstFetchSettled` itself; that one fixes a real frame flash.

Also: `refresh` is now dead — no consumer survives the sticky-bar removal. **−~10 lines.**

ts/host/src/embed/index.ts
```diff
@@ -64,6 +72,120 @@ const SDK_READY_TIMEOUT_MS = 8000;
...
+ * Ordering:
+ *   1. If the motor mode is still unknown (fresh session), ask for a state
+ *      snapshot and wait briefly for it so `isAwake()` is meaningful.
+ *   2. Already awake -> return immediately (no trajectory to play).
+ *   3. Asleep -> await the full `wakeUp()` trajectory. A timeout or error is
+ *      swallowed and we reveal the app anyway (degraded but reachable),
+ *      never trapping the user on the connecting splash.
+ */
+async function awaitFullyAwake(sdk: ReachyMiniInstance): Promise<void> {
+  if (sdk.robotState.motor_mode === undefined) {
```
FabienDanieau: `shrink:` 13 lines waiting on a state snapshot so `isAwake()` is meaningful before `wakeUp()`.

If the daemon no-ops `wake_up` on an already-awake robot, this is `await sdk.wakeUp({ timeoutMs }).catch(() => {})` — one line. Worth confirming daemon-side before cutting, since the point is avoiding a replayed animation.

ts/host/src/components/DaemonUpdateGate.tsx
```diff
@@ -0,0 +1,460 @@
...
+   *  offline-first reboot watch, see `lib/rebootWatch.ts` - a stale
+   *  pre-reboot listing must not complete the gate). Only meaningful
+   *  after the shell has torn the iframe down and resumed polling -
+   *  `false` at every other point in the flow. */
+  robotBackOnline: boolean;
+  /** True once the embedded app is interactive. The soft notice waits
+   *  for it so an optional offer doesn't land on top of the connecting
+   *  splash; the blocking tier ignores it, since the whole point is to
+   *  stop the user before the app opens. */
+  appLive: boolean;
```
FabienDanieau, line 84: `yagni:` (product call, not mine) the whole `notice` tier — dismissable card, this `appLive` prop, the second severity value — exists to tell a *working* robot it's a patch behind. The `block` tier is the one preventing a broken session.

**−~70 lines** if it goes. Flagging it because the PR body defends it explicitly; if it stays, it stays.

ts/host/src/components/ErrorView.tsx
```diff
@@ -85,7 +85,7 @@ export function ErrorView({
           <Button variant="outlined" onClick={onBackToPicker}>
             Back to picker
           </Button>
-          <Button variant="contained" onClick={onReload}>
+          <Button variant="outlined" onClick={onReload}>
```
FabienDanieau, line 90: Both buttons are now `outlined`, so neither reads as the primary action. Was this intentional, or fallout from the "outlined End session" pass?

ts/host/src/components/TopBar.tsx
```diff
@@ -267,162 +264,74 @@ function AccountMenu({
...
-            },
-          },
-        }}
+      </Stack>
+      <IconButton
+        onClick={onLogout}
+        disabled={disabled}
+        color="primary"
+        size="small"
+        aria-label={username ? `Sign out @${username}` : 'Sign out'}
```
FabienDanieau, line 333: One-click sign-out drops the guard the removed docstring explicitly justified ("prevents accidental sign-out clicks"). Mobile parity is a fine reason to change it — just noting the rationale was deleted rather than answered.

ts/host/src/hooks/useHostBridge.ts
```diff
@@ -51,9 +62,24 @@ export interface UseHostBridgeOptions {
   /** App requests a clean leave (in-app button, etc.). Host
    *  should run the same tear-down as a user-initiated end. */
   onRequestLeave(): void;
+  /** App finished its `host:leaving` tear-down (onLeave + the SDK's
+   *  host-owned sleep/disable). The host can unmount now instead of
+   *  waiting out its safety cap. */
+  onLeft(): void;
```
FabienDanieau: Nit: `onLeft` is required while `onUpdateProgress` is optional, for two equally additive callbacks. Making `onLeft` optional too keeps `UseHostBridgeOptions` non-breaking for anyone implementing it outside this repo.

ts/APP_CREATION_GUIDE.md
```diff
@@ -713,6 +713,7 @@ app_build_command: npm ci && npm run build   # HF runs this on its builder
 app_file: dist/index.html                     # HF serves the build output as entry
 pinned: false
 hf_oauth: true
+hf_oauth_expiration_minutes: 43200   # 30 days (max); default is only 8 h
```
FabienDanieau, line 717: Recommending the 30-day maximum to *every* app author is a posture change worth spelling out: combined with the `pagehide` wipe removal, an HF user token now lives in `sessionStorage` for the tab's lifetime and is handed to the iframe in the URL hash.

Worth documenting the scope set the token carries, and the SDK idle window that bounds it, right here where authors will read it.

ts/host/src/components/DaemonUpdateGate.tsx
tfrere: Fixed, both halves. The gate now surfaces the give-up through a dedicated `onCancelUpdate` callback, which the shell relays as a new `host:cancel-update` protocol message; the embed's handler disarms the update-mode plumbing (drops the sessionStopped translator, restores auto-reconnect). Additive message, older embeds ignore it. The resurrection path is also closed on the gate side: progress frames are only folded into the phase while an update we drove is in flight (`phase === 'updating' | 'rebooting'`), so a stale `rebooting` frame on a later normal end-session no longer repaints a dismissed gate. Covered by fake-timer tests.

ts/host/src/components/ReachyHostShell.tsx
tfrere: Fixed, and the fragility you spotted is gone at the root: the state machine was replaced by a sticky `sawOffline` boolean (see the `rebootWatch.ts` thread), and `selectRobot` now resets it alongside `awaitingReboot`. A pre-armed watch can no longer leak into the next update by construction. The stale-listing regression test survives, re-expressed on the boolean.

ts/host/src/components/DaemonUpdateGate.tsx
tfrere: Fixed - went with the reset option. The effect now depends on `progress`, so every frame re-arms the timer: it measures silence since the last line, not total install time. A slow-but-reporting pip on weak wifi can run as long as it keeps talking, and the "never started" copy is only reachable when the daemon actually went quiet. The constant's comment now says exactly that.

ts/host/src/components/ReachyHostShell.tsx
tfrere: Right - I had missed that `awaitMotorsDisabled` runs after the sleep race, not inside it. Bumped to 9.5 s and rewrote the comment to spell out the sequential worst case (6.5 s + 1 s = 7.5 s), which leaves ~2 s for postMessage + a React commit.

ts/host/src/embed/index.ts
tfrere: Agreed - it contradicted the PR's own two-tier argument, and keying it on the npm bundle version made it worse. Reverted: the overlay and the semver comparison are gone. The `sdkVersion` field stays on `embed:ready` (additive, useful telemetry), but nothing acts on it. If an app-side gate is ever warranted it should be policy-driven (daemon/central side), not bundle-version-driven.

ts/host/src/components/ReachyHostShell.tsx
tfrere: Done - the PR body now states the dependency on the sliding idle window explicitly. On the numbers: the window is 24 h (`TOKEN_MAX_IDLE_MS`), sliding on every successful read, and the token is sessionStorage-scoped, so the realistic tail is "a still-open tab last used within a day", not 30 days. The other bound is scope: the token only carries `openid profile` (no Hub or API access), which the guide now documents next to the `hf_oauth_expiration_minutes` recommendation. We considered shrinking the window to hours, but that reintroduces the daily re-auth the 30-day recommendation exists to remove; 24 h is the shortest value that keeps a daily user signed in.

ts/host/src/embed/index.ts
tfrere: Confirmed. The slow hint is scoped to the `wake` step (`useStepElapsedPast(step, 'wake', 6000)`) and its timer resets on step change, so `readDaemonVersion` (which runs under the previous step) cannot consume the budget. The hint lands 6 s into the wake wait - the only window that can stretch to ~9 s - so the degraded tail runs under "taking a moment to settle in" rather than a silent spinner.

ts/host/src/components/PickerView.tsx
tfrere: Done, nearly verbatim: the icon is always mounted inside its fixed-width slot, opacity tracks `isRefreshing` with the 320 ms fade, and the mounted/shown double state, trailing debounce and rAF machinery are gone (~75 lines). On the a11y note: agreed, the glyph is decorative - it is `aria-hidden` now, the list content is the accessible signal.

ts/host/src/lib/rebootWatch.ts
tfrere: Done. The state machine and its switch are gone; the shell holds a sticky `sawOffline` boolean and `sawOffline && isTargetListed(...)` is the back-online edge. `isTargetListed` (the identity-matching half) stays in the module, and the stale-listing case survives as a test on the boolean, per your note. As a bonus the reset hole in `selectRobot` closed itself - resetting one boolean is hard to get wrong.

ts/host/src/lib/daemonRelease.ts
tfrere: Done. The memo, `isFresh` and `cachedLatestDaemonVersion` are gone; `fetchLatestDaemonVersion` reads localStorage directly (first call writes before the second reads, so the fetch-called-once test still passes, plus stale-on-error fallback). `useLatestDaemonVersion` now starts from `null` and resolves in an effect - the lazy-initializer puzzle went with it.

ts/host/src/embed/index.ts
tfrere: Moot in the best way: the whole modal went away with the staleness-overlay revert (see the thread above), so there is nothing left to shrink. If the embed ever needs host-driven chrome again, `insertAdjacentHTML` with a template literal is the shape it should take.

ts/host/src/components/DaemonUpdateGate.tsx
tfrere: Done. `logLines`, `LOG_TAIL` and the panel are gone, and the `in_progress` branch is now a documented early-return whose only job is re-arming the stall timer (its liveness value survived the panel).

ts/host/src/components/DaemonUpdateGate.tsx
tfrere: Done. One `COPY: Record<Phase, ...>` indexed once (bodies are small functions so the dynamic bits - versions, failure detail - stay in place), and `PrimaryButton` collapsed to a `PRIMARY_SX` const on `<Button variant="contained">`.

ts/host/src/components/DaemonUpdateGate.tsx
tfrere: Trimmed. Kept the "why the update outlives the iframe" paragraph plus a short pointer to `daemonRelease.ts` for the policy; the two-tier restatement is gone.

ts/host/src/embed/index.ts
tfrere: Done. `on()` and its try/catch are gone - three direct `sdk.addEventListener` calls - and the same dead try/catch inside `startDaemonUpdateForHost` went with it. The `setAutoReconnect` feature-detect stays, as you noted: the embed rides whatever SDK the app bundled.

ts/host/src/components/PostOAuthSplash.tsx
tfrere: Done. The `variant` prop and the four guards are gone; the component only knows the signing-in state, and the neutral splash is a 6-line inline Box at the call site in `ReachyHostShell`.

ts/host/src/components/HostErrorBoundary.tsx
tfrere: Done. The fallback is down to emoji + Typography + contained Button in a centered Box; the grayscale filter, the `<pre>` and its font stack are gone. The class boundary itself is untouched.

ts/host/src/components/ReachyHostShell.tsx
tfrere: Done, as written - `Promise.race` against `sleep(LEAVING_ACK_CAP_MS)`, resolver ref nulled after. The `sleep` helper is back (net win over the flag + timer handle + double-settle guard it replaces).

ts/host/src/components/ReachyHostShell.tsx
tfrere: Done. One `welcomeBack: 'idle' | 'showing' | 'done'` replaces the two booleans and the one-shot ref: the enum's own transitions ('idle' is the only state that arms) encode what the ref was guarding.

ts/host/src/hooks/useRobots.ts
tfrere: Done. `isLoading` is gone from the hook and from `PickerViewProps` - the one consumer derives `!hasRobots && isRefreshing` locally - and dead `refresh` went with it. `firstFetchSettled` stays, as you said.

ts/host/src/embed/index.ts
tfrere: Confirmed daemon-side and keeping the guard: `wake_up` is NOT a no-op on an already-awake robot. The backend unconditionally replays the wake choreography (sound + antenna/head animation) on every call - there is no awake-state check in `abstract.py`. So the 13 lines are load-bearing: without the state wait, every reconnect to an awake robot would replay the wake animation mid-app.

ts/host/src/components/DaemonUpdateGate.tsx
tfrere: Keeping it, with the product hat on: the notice tier is what makes the block tier acceptable. The self-serve update path needs mileage on robots that still work, so that by the time a version falls below the floor, users have already met (and trusted) the update flow in its dismissable form. Happy to revisit if the dismiss rate says nobody ever updates from it.

ts/host/src/components/ErrorView.tsx
tfrere: Fallout, good catch. "Back to picker" is now the contained primary - it recovers in place and keeps the OAuth session, whereas reload starts the world over - with a comment so the next outlined-pass doesn't flatten it again.

ts/host/src/components/TopBar.tsx
tfrere: Fair - the rationale deserved to be answered, not deleted. Restored as a comment on the button: mobile parity, plus the cost asymmetry (a mis-click costs one silent-auth round trip, not data loss), and a nod to the discoverability problem of the old menu-behind-avatar guard.

ts/host/src/hooks/useHostBridge.ts
tfrere: Done - `onLeft` is optional now, symmetric with `onUpdateProgress`, so `UseHostBridgeOptions` stays additive for outside implementers.

ts/APP_CREATION_GUIDE.md
tfrere: Documented right there in the frontmatter section: the recommendation now comes with the two bounds spelled out - scope (keep the default `openid profile`: identity only, no Hub/API access, so a leaked token identifies but cannot act) and the SDK's 24 h sliding idle window on the sessionStorage-scoped token (session restore re-auths, a plain reload does not). The PR body of the stacked host PR calls out the dependency too.

ts/host/src/hooks/useOAuth.ts
tfrere: Fixed. `SDK_LOAD_GRACE_MS` (8 s) now bounds the wait: if no SDK instance materialises within the grace, `authResolved` flips true and the shell lands on SignInView, local-dev missing-config hint included. A late-arriving SDK still runs `authenticate()` and upgrades the state; the SignInView flash in that race is the accepted cost of never spinning forever.

ts/host/src/components/DaemonUpdateGate.tsx
tfrere: Fixed. The blocking tier now carries a "Back to my Reachies" action wired to a new `onExitApp` callback: the shell runs the same tear-down as the top-bar end-session and returns to the picker, releasing the app-slot lock. And yes - the block only paints after `startSession()` succeeded, which is exactly why the escape has to end the live session rather than just unmount the overlay; the prop's doc comment now spells that out.

ts/host/src/hooks/useOAuth.ts
tfrere: All three confirmed:

1. `login()` builds the URL with `oauthLoginUrl()`, whose default `redirectUrl` is the current `window.location.href`, query string included - `?config=` and friends come back verbatim on the return leg.
2. Social-preview bots don't execute JS, so they never trigger the redirect. JS-rendering crawlers reach the Space through the hub page, where it is framed - the iframe guard stops the redirect there. A direct `*.hf.space` hit by a rendering crawler would bounce once; accepted, since that origin is not the indexed surface.
3. It does: `authenticate()` calls `consumeOAuthErrorParams()` before the cached-token fall-through, stripping `error` / `error_description` via `history.replaceState`. Covered in `token-store.test.ts`.

FabienDanieau (approved):
Approving, on the same basis as #1315: reviewed the embed media path and the reconnect bridge in the diff, and exercised the stack on hardware via the #1321 branch. Not reviewed: the update gate, the picker/top-bar refresh, the boot path and silent OAuth.

**Credit where due** — clearing the media cache on `sessionReconnecting` (`embed/index.ts:1310`) is exactly right, and the comment explaining *why* ("the SDK's auto re-dial replaces the RTCPeerConnection, so the cached tracks are dead") is what tipped me off that the mobile client was getting this wrong. That one listener is the difference between the embed surviving a re-dial and not.

One follow-up, not blocking:

**The rebuild is lazy, so it doesn't reach every app.** `robotStream` is a getter over `ensureCached()`, and the clear only takes effect for a consumer that *re-reads* it. An app that grabs `media.robotStream` once at boot and holds the `MediaStream` never re-reads, so the invalidation never reaches it — same dead-track class the clear is there to prevent, just moved into the app.

Two ways out, either fine: push an app-state / event on `sessionReconnected` that means "re-read your media", or document the contract in `APP_CREATION_GUIDE.md` next to the existing reconnect notes. Given the audience is Space authors who won't be thinking about peer-connection lifecycles, I'd lean toward the event.

Related: this is the same root shape as the missing `audioTrack` event I flagged on #1315 — if that lands, this gets easier to solve for everyone at once.

PR #1321, opened 2026-08-05
feat(host): SDK staleness self-check v2 (npm source, major-only, persisted dismissal)

tfrere:
> Stacked on #1316. Rebuild of the SDK staleness overlay that was pulled out of that PR following @FabienDanieau's review (see the `maybeWarnSdkOutdated` thread there).

## What the review caught

The v1 overlay had a structural flaw, not a cosmetic one:

- **Wrong version source.** It compared the npm bundle version against the **daemon's** GitHub release tag - two version lines that only agree while they happen to be tagged together. The first daemon-only release would have painted a warning over every up-to-date app.
- **Kill-switch semantics.** It fired on *any* gap, full-screen, on **every reload**, with no persisted dismissal. Published Spaces are frozen bundles that are never rebuilt: any SDK release would have nagged every user of every published app, forever. That's exactly the "global kill switch" behaviour the daemon update policy in #1316 argues against - for apps this time.
- **Hardcoded light card** over apps that may be dark.

Softening it in place wouldn't have fixed the version-source flaw, hence this dedicated PR.

## What v2 does instead

Every flagged choice is inverted:

| | v1 (reverted) | v2 (this PR) |
|---|---|---|
| Version source | daemon GitHub release tag | the package's own npm `latest` dist-tag (~200-byte CORS JSON) |
| Trigger | any version gap | **major** gap only |
| Dismissal | per-pageload | persisted (localStorage), covers the acknowledged major, re-arms on the next major |
| Theming | hardcoded light | follows `prefers-color-scheme` |

The rationale for "major-only": a major bump is the only signal that the app/robot contract may genuinely no longer hold. Minor/patch releases must never nag a working app.

Fail-open discipline throughout: dev placeholders (`0.0.0-managed-by-ci`), unparseable versions, an unreachable registry - all resolve to "stay silent". Advisory code must never cost the user their app.

## Where it lives

The whole feature is now a React-free module, `host/src/lib/sdkStaleness.ts`, imported by the embed runtime (it still runs *inside* the app iframe, so every host - web shell, mobile WebView - is covered with zero parent integration). Extracting it out of `embed/index.ts` is what makes it testable.

## Tests

15 new tests pin the behaviours the review called out:

- policy matrix: silent on minor/patch gaps, fires on major, silent on dev/unparseable versions
- registry cache: one network hit per TTL, stale-on-error fallback, never throws
- persisted dismissal: one click covers the major, re-arms on the next major, **no re-nag on reload**
- overlay: single mount, dismiss round-trip, versions re-serialised through the semver parser before touching `innerHTML`

`vitest` 171/171, `tsc` clean.

---

*Made with [Cursor](https://cursor.com)*


FabienDanieau (approved):
Approving — this one I read essentially in full (3 files, +423), and the v2 inversion addresses every point from the #1316 thread: npm dist-tag as the version source instead of the daemon release tag, major-only, persisted dismissal, `prefers-color-scheme`. The fail-open discipline throughout is the right instinct for advisory code.

Dismissal semantics look correct to me: `l.major <= dismissed.major` means one acknowledgement covers the major and a genuinely newer major re-arms.

Two notes, neither blocking:

**1. Worth stating that this can't be smoke-tested from a dev build.** `isSdkMajorBehind` returns false when `current.major === 0`, and anything built from these branches is `0.0.0-branch.*`. That's correct fail-open behaviour, but it means "I ran it and saw no overlay" is indistinguishable from a pass — which is a trap for whoever validates this next. The 15 unit tests *are* the validation; worth saying so in the PR body, along with the manual recipe for anyone who wants to see the thing:

```js
localStorage.setItem('reachy.sdkLatestRelease.v1', JSON.stringify({ value: '99.0.0', at: Date.now() }))
localStorage.removeItem('reachy.sdkStalenessDismissed.v1')
```

(plus a build carrying a non-zero major, since the `0.x` guard fires first).

**2. The 6 h TTL means a real major surfaces up to 6 hours late.** Almost certainly the right trade against hammering the registry — just worth being explicit, given the entire point is a channel to frozen Spaces that have no other way to hear anything.

PR #1326, opened 2026-08-07
deps: allow secure Starlette 1.x releases

### LINKED ISSUE #1317: starlette<1.0.0 pin holds downstream apps on a version with five open advisories
`reachy_mini` requires `starlette<1.0.0`. Starlette is currently at 1.3.1, and every version below it carries open advisories. Because the constraint is transitive, an application built on the SDK cannot resolve them on its own — the pin is the blocker, not anything in the app.

The five that apply, from Dependabot on an app using the SDK (installed `reachy-mini` 1.10.0rc2, which resolves `starlette` 0.52.1):

| Severity | Advisory | Fixed in | Summary |
|---|---|---|---|
| High | GHSA-82w8-qh3p-5jfq | 1.3.1 | `request.form()` limits silently ignored for `application/x-www-form-urlencoded`, enabling DoS |
| High | GHSA-wqp7-x3pw-xc5r | 1.1.0 | SSRF and NTLM credential theft via UNC paths in `StaticFiles` on Windows |
| Medium | GHSA-x746-7m8f-x49c | 1.1.0 | Arbitrary HTTP method dispatched to `HTTPEndpoint` attributes via `getattr` |
| Medium | GHSA-86qp-5c8j-p5mr | 1.0.1 | Missing Host header validation poisons `request.url.path`, bypassing path-based checks |
| Low | GHSA-jp82-jpqv-5vv3 | 1.3.0 | Unvalidated request path concatenated into authority, poisoning `request.url.hostname` |

**GHSA-wqp7-x3pw-xc5r looks the most relevant to this SDK specifically.** The app template mounts `StaticFiles` to serve its settings UI under `--ui`, and Windows is a supported platform, so that combination arises in normal use rather than an unusual one. Exposure is limited by the server binding to loopback, but it is reachable by any local process.

Would you consider relaxing the constraint to allow Starlette 1.x?

I have not opened a PR because I do not know whether the `<1.0.0` bound is guarding a specific incompatibility. If it is, knowing which API would help downstream projects plan.

Daniiiil1:
## Summary

- replace the obsolete `starlette<1.0.0` ceiling with `starlette>=1.3.1,<2.0.0`
- refresh the lockfile to FastAPI 0.141.1 and Starlette 1.3.1
- keep downstream applications off Starlette releases affected by the advisories documented in #1317

The TODO on the previous ceiling referred to a `TemplateResponse` call signature, but the repository no longer uses `TemplateResponse`, `Jinja2Templates`, or Starlette/FastAPI templating imports.

## Validation

- `uv lock --check`
- `ruff check .`
- 72 focused API/router tests passed on FastAPI 0.141.1 + Starlette 1.3.1
- broader non-hardware run: 701 passed; remaining failures required unavailable local optional Mujoco, Placo, GStreamer, or ReSpeaker dependencies

Closes #1317

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,8 @@ dependencies = [
     "rich",
     "rustypot>=1.4.2",
     "scipy>=1.15.3, <2.0.0",
-    "starlette<1.0.0", # TODO: update TemplateResponse(request, name, context) before allowing Starlette 1.x
+    # Starlette 1.3.1 fixes the last advisory affecting the previous 0.x range.
+    "starlette>=1.3.1,<2.0.0",
     "toml",
     "uvicorn[standard]",
     "websockets>=12,<16",
--- a/uv.lock
+++ b/uv.lock
@@ -828,7 +828,7 @@ wheels = [
 
 [[package]]
 name = "fastapi"
-version = "0.129.0"
+version = "0.141.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "annotated-doc" },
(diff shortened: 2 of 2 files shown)


FabienDanieau (changes requested):
Reviewed, and tested this bump on a real robot as well as locally. No regression found.

**Robot** (Reachy Mini Wireless, ReachyMiniOS 1.10.0), starlette swapped to 1.3.1 in the daemon venv, wheel checked against the hash in `uv.lock`:

- 25 HTTP and WebSocket checks pass, identical to the 0.52.1 baseline taken before and restored after: routes, `/docs`, `/openapi.json`, `FileResponse` with single and overlapping `Range` headers, CORS preflight and rejected origin, multipart upload with the extension allow-list, the non-audio rejection and the 25 MB cap returning 413, five websockets (`/logs/ws/daemon`, `/api/state/ws/full`, `/api/move/ws/updates`, `/api/move/ws/set_target`, `/ws/sdk`), and two real `goto` moves.
- `reachy_mini_conversation_app` started, initialized its realtime session and spoke its greeting; audio pipeline, WebRTC signalling server and the HF central relay all came up.
- No traceback and no deprecation warning in `journalctl -u reachy-mini-daemon` across the restart. `/openapi.json` was byte-identical on 0.52.1 and 1.3.1.

**Unit tests** (`uv sync --frozen --all-extras --group dev`, CI marker set):

| Env | starlette | Result |
|---|---|---|
| current main | 0.52.1 | 935 passed, 66 deselected |
| current main + `starlette==1.3.1` | 1.3.1 | 935 passed, 66 deselected |
| current main + this PR's specifier, fresh resolve | 1.6.0 | 935 passed, 66 deselected |

The hardware-marked tests CI never runs (`audio or video or wireless or webrtc`) fail identically with and without the change: 8 failed, 57 passed, and the 8 are the ReSpeaker board and attached-robot tests, which cannot pass on a laptop.

The floor is right, too: 1.3.1 is the first release clearing all five advisories open against 0.52.x, since GHSA-82w8-qh3p-5jfq needs exactly 1.3.1 and GHSA-jp82-jpqv-5vv3 needs 1.3.0.

**One change before merge: regenerate the OpenAPI spec.** fastapi 0.141.1 emits the OpenAPI 3.1 form for the upload field, so the committed spec drifts by one line:

```
-            "format": "binary",
+            "contentMediaType": "application/octet-stream",
```

`uv run python scripts/generate_openapi.py`, then commit `docs/source/API/openapi.json`. `lint.yml` triggers on push with `pyproject.toml` in its paths, so without this main's `check-api-docs` goes red immediately after the merge.

Two notes while you are in there:

- This branch has never had a successful CI run. Every workflow run on `e2acefbd` failed at startup with zero jobs created, which is why the PR reports no checks at all rather than red ones. Pushing the spec commit will re-trigger them, so please confirm they come back green.
- Filling in the PR template (Testing, Tested on, AI assistance) would be good; feel free to quote the robot results above and tick Reachy Mini Wireless.

This supersedes #1391, which proposed the same change with a `>=1.1.0` floor. I have closed that one and pointed it here.

PR #1328, opened 2026-08-07
fix(camera): give read_jpeg a longer sample timeout

tfrere:
`read_jpeg` pushes a frame into the one-shot JPEG encoder and then pulls with `get_sample`'s default 20 ms timeout. That is fine for video-rate pulling (a missed frame is fine) but too tight for a one-shot encode: on a loaded CI runner the pipeline has not prerolled and emitted the encoded buffer yet, so `try_pull_sample` returns None and `read_jpeg` returns None. It surfaces as a flaky `test_read_jpeg_encodes_bgr_frame` failing with `assert None is not None`.

This is a pre-existing flake (the test and code are already on `main`), split out of #1312 where it was turning CI red on an otherwise-unrelated change.

- `get_sample` gains an optional `timeout_ns` (default 20 ms), so the three video-rate call sites are untouched.
- `read_jpeg` opts into a 1 s timeout. Its only caller, `media_manager.get_frame_jpeg`, has no callers today, so the blast radius is nil.

Tested: `test_video.py::test_read_jpeg_encodes_bgr_frame` passes with `--extra opencv` (skipped by default via `importorskip("cv2")`), on both aarch64 and macOS.

### DIFF
--- a/src/reachy_mini/media/camera_base.py
+++ b/src/reachy_mini/media/camera_base.py
@@ -191,7 +191,10 @@ class CameraBase(ABC):
             self._build_jpeg_encoder(width, height)
         self._jpeg_pipeline.set_state(Gst.State.PLAYING)
         self._jpeg_appsrc.push_buffer(Gst.Buffer.new_wrapped(frame.tobytes()))
-        jpeg = get_sample(self._jpeg_appsink, self.logger)
+        # One-shot encode: unlike video-rate pulling, we must wait for the
+        # pipeline to preroll and emit the encoded buffer. The default 20 ms is
+        # too tight on a loaded CI runner and returns None there.
+        jpeg = get_sample(self._jpeg_appsink, self.logger, timeout_ns=Gst.SECOND)
         self._jpeg_pipeline.set_state(Gst.State.PAUSED)
         return jpeg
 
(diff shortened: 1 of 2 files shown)


FabienDanieau (approved):
Approving — diagnosis is right and the fix is minimal. Checked the five `get_sample` call sites: the default keeps the four video/audio-rate ones untouched, only `read_jpeg` opts in.

One correction to the description though: `get_frame_jpeg` **does** have a caller, just not in this repo. The conversation app's `camera` tool uses it:

```python
# reachy_mini_conversation_app, src/reachy_mini_conversation_app/tools/camera.py
jpeg_bytes = deps.reachy_mini.media.get_frame_jpeg()
if jpeg_bytes is None:
    logger.error("No frame available from camera")
    return {"error": "No frame available"}
```

So this isn't only a CI fix — the same 20 ms flake is what makes the assistant answer *"No frame available"* when you ask it what it can see on a loaded robot. Worth claiming that win in the description rather than calling the blast radius nil.

Trade worth knowing: `get_frame_jpeg()` is a blocking call inside an `async def` there, so the worst-case event-loop stall on the *failure* path goes from 20 ms to 1 s. Since `try_pull_sample` returns as soon as the buffer arrives, the success path is unchanged — only failures get slower to report. Not a blocker, and the sync-call-in-async is pre-existing, but it's a real ceiling on the voice loop now.

Unrelated heads-up: #1282 adds the identical `timeout_ns` parameter to `get_sample` (same signature, same default, same body — only the docstring differs, and yours is the better one). Whichever lands second should drop that hunk.

PR #1333, opened 2026-08-09
feat(daemon): reset to sleep from any inherited motor state

tfrere:
## Problem

`goto_sleep()` assumes it inherits a robot under position control. A crashed app, a killed tab or a dropped Wi-Fi link skip the client's own cleanup and can leave the robot in gravity compensation, globally limp, or with a per-motor torque cut. In all three cases `goto_sleep()` degrades silently: the control loop ignores position targets, the sleep trajectory is written into the void, and the final torque cut drops the head from wherever it happened to be.

Follow-up to #1294 / #1311.

## What changes

- **`reset_to_sleep()`**: re-establish the assumption first, then reuse the normal sequence. Quiesce aim sources (wobbling, speech offsets, head tracking), torque on (`enable_motors` pins every target to the measured pose, so nothing snaps), lift to the init pose - duration clamped to [0.3, 1.5] s so a fully drooped head never crawls back up and a head already at init still gets a real move - then `goto_sleep()`, ending limp at the sleep pose. Both the daemon's sleep path and the idle reset now go through it.
- **`_already_idle()`** replaces the motor-mode-only skip in the idle reset: motors disabled AND measured pose within the sleep radius. The radius (`SLEEP_POSE_MAGIC_ATOL`) is shared with `goto_sleep`'s "no travel needed" branch so the trajectory and the skip cannot drift apart. A robot left limp mid-pose by a crashed app is no longer mistaken for one parked down by a clean leave sequence.
- **Per-motor torque tracking** (`RobotBackend`): `set_motor_torque_ids()` records a `_partial_torque_override`, so the next `set_motor_control_mode()` re-applies the mode even when it looks unchanged instead of short-circuiting on a stale equal-looking value - that re-torques motors an app detorqued individually.

## Tests

- `test_backend_reset_to_sleep.py` (new): locks in the ordering that makes the whole thing work (quiesce -> enable -> lift -> sleep), the lift target and duration clamp, and the `is_at_sleep_pose` predicate (true at sleep pose, false at init, false before the first kinematics update).
- `test_backend_idle_reset.py` (updated): the idle reset now skips only when limp AND at the sleep pose; limp mid-pose triggers the reset.

17/17 unit tests pass locally.

### DIFF
--- a/src/reachy_mini/daemon/backend/abstract.py
+++ b/src/reachy_mini/daemon/backend/abstract.py
@@ -1194,6 +1194,11 @@ class Backend:
     # a (velocity-eased) tick short of the commanded endpoint.
     INIT_TARGET_ATOL: float = 1e-2
 
+    # Radius (magic units) around SLEEP_HEAD_POSE within which the robot counts
+    # as already down. Shared by goto_sleep's "no travel needed" branch and
+    # is_at_sleep_pose() so the trajectory and the skip can't drift apart.
+    SLEEP_POSE_MAGIC_ATOL: float = 10.0
+
     def is_awake_at_init_pose(self) -> bool:
         """Motors on and init is the pose the controller is actively holding.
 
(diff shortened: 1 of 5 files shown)


tfrere:
Superseded: this landed in #1315 (feat/sdk-js-core, commit 44e9c9cfc) so it ships with the already-approved SDK/host stack instead of waiting on a separate review. Branch kept for reference.

PR #1340, opened 2026-08-11
feat(daemon): persisted first wake-up flag over the data channel

tfrere:
## Issue

Extracted from #1286 (first wake-up wizard support).

The mobile app runs a one-time, post-connection hardware diagnostic wizard ("first wake-up"). Whether it has been completed must live on the robot, not in the app, so the wizard shows once per robot whichever client connects. The daemon currently has no way to store or expose that flag, so the app-side wizard can't ship gated.

## Description

- New data-channel commands `get_first_wake_up` / `set_first_wake_up` (`protocol.py` + handler in `abstract.py`).
- The flag is a `first_wake_up_completed` key in the existing `daemon_config.json` (no new file), through shared generic helpers in `startup_app_config.py`: `_get_bool(key)` for validated reads (now also backing `get_turn_enabled`) and `_update(key, value)` for read-modify-write. Read/write is fail-safe: errors degrade to "not completed" / "write failed" instead of raising, so a storage problem can't break the command loop.
- JS SDK: `getFirstWakeUp()` / `setFirstWakeUp(isCompleted)` on the existing `_slotRoundtrip` helper. Both resolve `null` when the channel is closed or the daemon predates the command, so clients fail-open (skip the wizard).

Purely additive; no HTTP route added. The BLE IDENTIFY half of #1286 will come separately.

## Testing

- `tests/unit_tests/test_startup_app_config.py` (17 new tests): direct coverage of the generic `_update` / `_get_bool` helpers (create/overwrite/clear, other keys preserved, corrupt-file recovery, write-error propagation, malformed-value warnings).
- `tests/unit_tests/test_first_wake_up.py` (9 new tests): default-to-false, roundtrip, reset, malformed value, other config keys preserved, fail-safe on corrupt JSON and write errors.
- `test_backend_process_command.py`, `test_backend_wake_up.py`, `test_autostart_app.py` still green (101 tests total); ruff and `npm run typecheck:sdk` clean.
- On-robot smoke test pending: `get_first_wake_up` -> false, `set` -> true, survives daemon restart.

## Tested on

- [ ] Reachy Mini Wireless (unit tests only so far; on-robot smoke test pending)
- [ ] Reachy Mini Lite

## AI assistance

Assisted-by: Cursor:fable-5

### DIFF
--- a/src/reachy_mini/daemon/startup_app_config.py
+++ b/src/reachy_mini/daemon/startup_app_config.py
@@ -17,6 +17,7 @@ logger = logging.getLogger(__name__)
 _KEY = "startup_app"
 _EQ_KEY = "speaker_eq_gains"
 _TURN_KEY = "turn_enabled"
+_FIRST_WAKE_UP_KEY = "first_wake_up_completed"
 # equalizer-10bands accepts per-band gains in [-24, +12] dB.
 _EQ_GAIN_MIN, _EQ_GAIN_MAX = -24.0, 12.0
 
@@ -88,6 +89,26 @@ def get_speaker_eq_gains() -> list[float] | None:
     return None
 
 
+def _get_bool(key: str) -> bool | None:
+    """Return the boolean at `key`, or None when unset or malformed.
+
+    A present-but-malformed value is warned about (so the user knows their
+    hand-edited config was ignored) and treated as unset.
+    """
+    config = _read()
+    if key not in config:
+        return None
+    value = config[key]
+    if isinstance(value, bool):
+        return value
+    logger.warning(
+        "Ignoring invalid '%s' in daemon config (need true or false); "
+        "using the built-in default.",
+        key,
+    )
+    return None
+
+
 def get_turn_enabled() -> bool | None:
     """Return whether the media server should offer TURN relay candidates.
 
@@ -96,30 +117,53 @@ def get_turn_enabled() -> bool | None:
     turning it off saves a background thread refreshing credentials on a
     robot that is only ever reached from its own network.
     """
-    config = _read()
-    if _TURN_KEY not in config:
-        return None
-    value = config[_TURN_KEY]
-    if isinstance(value, bool):
-        return value
-    # Present but malformed: warn so the user knows it was ignored.
-    logger.warning(
-        "Ignoring invalid '%s' in daemon config (need true or false); "
-        "using the built-in default.",
-        _TURN_KEY,
-    )
-    return None
+    return _get_bool(_TURN_KEY)
 
 
-def set_startup_app(name: str | None) -> None:
-    """Persist the startup app name; a falsy name clears it."""
+def _update(key: str, value: object | None) -> None:
+    """Read-modify-write a single config key; `None` clears it.
+
+    Other keys are preserved. May raise OSError on a write error; callers
+    that must never raise (e.g. the daemon command loop) wrap it themselves.
+    """
     config = _read()
-    if name:
-        config[_KEY] = name
+    if value is None:
+        config.pop(key, None)
     else:
-        config.pop(_KEY, None)
+        config[key] = value
 
     path = _config_path()
     path.parent.mkdir(parents=True, exist_ok=True)
     with path.open("w") as f:
         json.dump(config, f, indent=2)
+
+
+def set_startup_app(name: str | None) -> None:
+    """Persist the startup app name; a falsy name clears it."""
+    _update(_KEY, name if name else None)
+
+
+def get_first_wake_up_completed() -> bool:
+    """Return True if the first wake-up setup wizard has been completed.
+
+    Robot-wide, persistent flag (not per-session): the mobile / desktop apps
+    run a one-time, post-connection hardware diagnostic wizard, and gate it on
+    this so it only ever shows once, whichever client connects. Defaults to
+    False (wizard pending) when unset or malformed, so a config problem can
+    never trap the daemon command loop.
+    """
+    return _get_bool(_FIRST_WAKE_UP_KEY) is True
+
+
+def set_first_wake_up_completed(is_completed: bool) -> bool:
+    """Persist the first wake-up completion flag. Returns True on success.
+
+    Fail-safe: a write error is logged and reported as False instead of
+    raising, so a storage problem can't break the daemon command loop.
+    """
+    try:
+        _update(_FIRST_WAKE_UP_KEY, bool(is_completed))
+        return True
+    except OSError as e:
+        logger.warning(f"Could not persist first-wake-up flag: {e}")
+        return False
--- a/src/reachy_mini/daemon/startup_app_config.py
+++ b/src/reachy_mini/daemon/startup_app_config.py
@@ -134,8 +134,13 @@ def _update(key: str, value: object | None) -> None:
 
     path = _config_path()
     path.parent.mkdir(parents=True, exist_ok=True)
-    with path.open("w") as f:
+    # Write to a sibling then rename: a crash mid-write would otherwise
+    # truncate the file, and the next _update() rebuilds it from the empty
+    # dict _read() returns for unparseable JSON, dropping every other key.
+    tmp = path.with_suffix(".json.tmp")
+    with tmp.open("w") as f:
         json.dump(config, f, indent=2)
+    tmp.replace(path)
 
 
 def set_startup_app(name: str | None) -> None:
--- a/tests/unit_tests/test_startup_app_config.py
+++ b/tests/unit_tests/test_startup_app_config.py
@@ -1,0 +1,112 @@
+"""Unit tests for the generic daemon-config helpers.
+
+``_update`` / ``_get_bool`` are the shared primitives behind every per-key
+accessor in ``startup_app_config`` (startup app, TURN toggle, first-wake-up
+flag). Exercise them directly against a tmp config path (monkeypatched
+``_config_path``) so the real user config dir is never touched.
+"""
+
+import json
+from pathlib import Path
+
+import pytest
+
+from reachy_mini.daemon import startup_app_config
+
+
+@pytest.fixture
+def config_path(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> Path:
+    """Point the module at a throwaway daemon_config.json under a tmp dir."""
+    path = tmp_path / "reachy_mini" / "daemon_config.json"
+    monkeypatch.setattr(startup_app_config, "_config_path", lambda: path)
+    return path
+
+
+def _stored(config_path: Path) -> dict:
+    return json.loads(config_path.read_text())
+
+
+# ─── _update ─────────────────────────────────────────────────────────────
+
+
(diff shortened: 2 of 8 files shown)


FabienDanieau (changes requested):
Tested end to end: helpers on a real Wireless (arm64, isolated `XDG_CONFIG_HOME`), then the full roundtrip over the WebRTC data channel from a browser — `false` → `set(true)` → `true` → `set(false)` → `false`, persisted across processes, `startup_app` preserved. The daemon side works. Two blockers; the smaller stuff is inline as committable suggestions.

### 1. `getFirstWakeUp` / `setFirstWakeUp` reject when the channel isn't open

The JSDoc and the PR description both promise `null` so callers fail-open, but `_slotRoundtrip` rejects:

```
getFirstWakeUp: REJECTED -> Data channel not open
setFirstWakeUp: REJECTED -> Data channel not open
```

The wizard gate runs right after connect, which is exactly when the channel may not be open yet — a caller following the doc gets an unhandled rejection instead of a skipped wizard. Suggestion inline on both methods.

### 2. No test for the dispatch branch

`test_startup_app_config.py` is a good addition, but it goes *down* to the primitives — the untested code is *up* in `abstract.py`: the `status: ok|error` mapping and the error path that acks the **stored** value rather than the requested one. `test_backend_process_command.py` already covers this shape (`test_delete_hf_token_ok/_error`, `test_set_volume_failure_returns_current`). This passes as-is against the current head:

```python
from reachy_mini.daemon import startup_app_config
from reachy_mini.io.protocol import GetFirstWakeUpCmd, SetFirstWakeUpCmd


@pytest.fixture
def cfg(tmp_path, monkeypatch):
    monkeypatch.setattr(
        startup_app_config, "_config_path", lambda: tmp_path / "daemon_config.json"
    )


def test_first_wake_up_round_trips(sim_backend, cfg):
    assert _dispatch(sim_backend, GetFirstWakeUpCmd()) == [
        {"command": "get_first_wake_up", "is_completed": False}
    ]
    assert _dispatch(sim_backend, SetFirstWakeUpCmd(is_completed=True)) == [
        {"command": "set_first_wake_up", "status": "ok", "is_completed": True}
    ]
    assert _dispatch(sim_backend, GetFirstWakeUpCmd()) == [
        {"command": "get_first_wake_up", "is_completed": True}
    ]


def test_first_wake_up_write_failure_acks_stored_value(sim_backend, cfg, monkeypatch):
    def boom(*a, **k):
        raise OSError("read-only filesystem")

    monkeypatch.setattr(startup_app_config, "_update", boom)
    assert _dispatch(sim_backend, SetFirstWakeUpCmd(is_completed=True)) == [
        {"command": "set_first_wake_up", "status": "error", "is_completed": False}
    ]
```

### Non-blocking, no suggestion

On write failure the daemon re-reads and sends `is_completed`, but the SDK maps `status === 'error'` → `null` and discards it. Either drop the extra read or use it client-side.

ts/lib/reachy-mini.ts
```diff
@@ -1418,6 +1421,33 @@ export class ReachyMini extends EventTarget implements ReachyMiniInstance {
...
+     * predates the `get_first_wake_up` command (callers should fail-open
+     * and skip the wizard on `null`).
+     */
+    getFirstWakeUp(): Promise<boolean | null> {
+        return this._slotRoundtrip(
+            () => this._firstWakeUpResolve,
+            (next) => { this._firstWakeUpResolve = next; },
+            { type: 'get_first_wake_up' },
+        );
+    }
```
FabienDanieau, line 1440: Fail-open guard, so this matches the doc right above it and the PR description ("clients fail-open"). Same early-return shape as `rpcCall`. No race: nothing awaits between the check and `_slotRoundtrip`.

```suggestion
    getFirstWakeUp(): Promise<boolean | null> {
        // Fail-open, so this never rejects: the wizard gate runs right after
        // connect, which is exactly when the channel may not be open yet.
        if (!this._dc || this._dc.readyState !== 'open') return Promise.resolve(null);
        return this._slotRoundtrip(
            () => this._firstWakeUpResolve,
            (next) => { this._firstWakeUpResolve = next; },
            { type: 'get_first_wake_up' },
        );
    }
```

Alternative if you prefer no duplication: give `_slotRoundtrip` a `failOpen = false` 4th param and pass `true` from both methods. Same result, one guard, but it touches a line outside this diff so I can't suggest it inline.

ts/lib/reachy-mini.ts
```diff
@@ -1418,6 +1421,33 @@ export class ReachyMini extends EventTarget implements ReachyMiniInstance {
...
+     * Persist the first wake-up wizard completion flag on the robot.
+     * Resolves with the stored value (or `null` on channel-closed).
+     */
+    setFirstWakeUp(isCompleted: boolean): Promise<boolean | null> {
+        return this._slotRoundtrip(
+            () => this._firstWakeUpResolve,
+            (next) => { this._firstWakeUpResolve = next; },
+            { type: 'set_first_wake_up', is_completed: isCompleted },
+        );
+    }
```
FabienDanieau, line 1453: Same for the setter.

```suggestion
    setFirstWakeUp(isCompleted: boolean): Promise<boolean | null> {
        if (!this._dc || this._dc.readyState !== 'open') return Promise.resolve(null);
        return this._slotRoundtrip(
            () => this._firstWakeUpResolve,
            (next) => { this._firstWakeUpResolve = next; },
            { type: 'set_first_wake_up', is_completed: isCompleted },
        );
    }
```

With both applied, built and run against `dist/`: `getFirstWakeUp`/`setFirstWakeUp` resolve `null`, `getRobotName`/`getVolume` still reject. `typecheck:sdk` clean.

src/reachy_mini/daemon/startup_app_config.py
```diff
@@ -88,38 +89,81 @@ def get_speaker_eq_gains() -> list[float] | None:
...
+    if value is None:
+        config.pop(key, None)
     else:
-        config.pop(_KEY, None)
+        config[key] = value
 
     path = _config_path()
     path.parent.mkdir(parents=True, exist_ok=True)
     with path.open("w") as f:
         json.dump(config, f, indent=2)
```
FabienDanieau, line 142: `_update` is now the only writer, so one `tmp + replace` makes `startup_app`, `turn_enabled` and the new flag all crash-safe. Without it, a truncated file feeds `_read()`'s empty-dict fallback and the next write drops the other keys — which `test_update_recovers_from_corrupt_file` currently pins as intended.

```suggestion
    path = _config_path()
    path.parent.mkdir(parents=True, exist_ok=True)
    # Write to a sibling then rename: a crash mid-write would otherwise
    # truncate the file, and the next _update() rebuilds it from the empty
    # dict _read() returns for unparseable JSON, dropping every other key.
    tmp = path.with_suffix(".json.tmp")
    with tmp.open("w") as f:
        json.dump(config, f, indent=2)
    tmp.replace(path)
```

No `fsync` on purpose — `os.replace` alone covers the truncation case, which is the one that loses data.

tests/unit_tests/test_startup_app_config.py
```diff
@@ -0,0 +1,112 @@
...
+
+def test_update_raises_on_write_error(
+    config_path: Path, monkeypatch: pytest.MonkeyPatch
+):
+    def boom(*a, **k):
+        raise OSError("read-only filesystem")
+
+    monkeypatch.setattr(startup_app_config.Path, "open", boom)
+    with pytest.raises(OSError):
+        startup_app_config._update("some_key", "value")
```
FabienDanieau: `monkeypatch.setattr(startup_app_config.Path, "open", ...)` patches `pathlib.Path.open` process-wide, so the `_read()` at the top of `_update` raises too and gets swallowed by its own `except OSError`. The test passes via "everything fails" rather than "the write failed". A real unwritable path is tighter:

```suggestion
def test_update_raises_on_write_error(tmp_path: Path, monkeypatch: pytest.MonkeyPatch):
    # A plain file where the config's parent dir should be: mkdir and open
    # fail for real, without patching pathlib globally (which would also
    # break the _read() at the top of _update, masking what is under test).
    blocked = tmp_path / "blocked"
    blocked.write_text("not a directory")
    monkeypatch.setattr(
        startup_app_config, "_config_path", lambda: blocked / "daemon_config.json"
    )
    with pytest.raises(OSError):
        startup_app_config._update("some_key", "value")
```

Same pattern in `test_first_wake_up.py::test_set_is_failsafe_on_write_error`.

tests/unit_tests/test_startup_app_config.py
```diff
@@ -0,0 +1,112 @@
...
+
+def test_update_none_clears_key(config_path: Path):
+    startup_app_config._update("some_key", "value")
+    startup_app_config._update("some_key", None)
+    assert _stored(config_path) == {}
+
+
+def test_update_none_on_absent_key_is_noop(config_path: Path):
+    startup_app_config._update("some_key", None)
+    assert _stored(config_path) == {}
```
FabienDanieau: Not a no-op: it creates a config file where none existed. Worth saying so, since that is what users see on a fresh robot — clearing a key they never set leaves a `daemon_config.json` behind.

```suggestion
def test_update_none_on_absent_key_writes_empty_config(config_path: Path):
    # Not a no-op on disk: clearing a key a fresh install never had still
    # creates the config file, holding an empty object.
    startup_app_config._update("some_key", None)
    assert _stored(config_path) == {}
```

tfrere:
Thanks for the on-robot run, that's exactly the coverage I couldn't do today.

All addressed in three commits (no rewrite since review started):

- fail-open guards on both SDK methods, as suggested. Went with the inline guard rather than the _slotRoundtrip param to keep the diff local; happy to switch if you prefer.
- atomic tmp + replace in _update
- your two dispatch tests added as-is (just wrapped the cfg fixture), write-error tests now use a real unwritable path, and the noop test is renamed

On the non-blocking point: kept the re-read. The error ack reporting what's actually on disk seems like the only honest value, and the SDK maps error to null on purpose so a caller can't mistake a failed set for a stored value. Can revisit if you feel strongly.

FabienDanieau (approved):
All five suggestions applied faithfully, nothing else changed. Re-verified on `4958787f9`:

- 120 tests green on a real Wireless (arm64), including the two dispatch tests
- atomic write on the robot's own filesystem: `startup_app` + the flag both intact, no `.tmp` left behind
- built SDK: `getFirstWakeUp` / `setFirstWakeUp` resolve `null` with no channel, `getRobotName` / `getVolume` still reject
- ruff, mypy, `typecheck:sdk` clean

Agreed on keeping the re-read — an error ack that reports what's actually on disk is the useful one, and collapsing `error` → `null` client-side is what stops a caller mistaking a failed set for a stored value.

Non-blocking, for later: the temp file has a fixed name (`daemon_config.json.tmp`), so two writers racing could rename a half-written file into place. I didn't confirm whether the HTTP route and the data-channel handler can ever run on different threads — if they share the event loop, `_update` runs to completion and it can't happen. Not a regression, and a pid/uuid suffix closes it if those writers ever diverge.

LGTM.

Issue #3, opened 2025-05-27
Better client/server


apirrone:
Add pose goto, joints goto, get present positions etc

pierre-rouanet:
I've started a wip branch to abstract communication here:

https://github.com/pollen-robotics/stewart_little_control/tree/robust-client-server-experimentation

pierre-rouanet:
Fixed by #22

PR #1341, opened 2026-08-11
feat(api): expose IMU on REST state routes and WebRTC state snapshot

darkseed:
### Motivation

The wireless Reachy Mini's IMU (BMI088) was only reachable through the legacy WS protocol (`GetImuCmd` + 50 Hz `ImuDataMsg` broadcast on `/ws/sdk`). The newer API surfaces — the REST `/state/*` routes and the WebRTC state frames parsed by the JS SDK — had no IMU at all. This PR brings them to parity. The Lite and the simulator have no IMU: on those, every new surface fails silently (`null`/`undefined`, never an error), matching the existing `/state/doa` contract.

### What changed

**Wire model** (`src/reachy_mini/io/protocol.py`)
- New `ImuData` model carrying the four payload fields: `accelerometer` (3, m/s²), `gyroscope` (3, rad/s), `quaternion` (4, w-first), `temperature` (°C).
- `ImuDataMsg` now subclasses `ImuData`, adding only the `type: "imu_data"` discriminator — wire-compatible for existing clients (same fields and values; only JSON key order shifts, which nothing consumes: `ws_client.py` discriminates by model, not position).
- `StateSnapshot` gains an optional `imu` field (backward-safe for released JS SDKs: new optional field, existing fields untouched).

**Backend** (`src/reachy_mini/daemon/backend/abstract.py`)
- `build_state_snapshot()` fills `imu` from the existing `get_imu_data()` cache. This single point feeds both the 30 Hz WebRTC pose-channel push and the `get_state` reply on every transport. Because the snapshot field is typed `ImuData`, the legacy `type` discriminator never leaks into snapshot JSON. No robot-backend changes — the cache refresh already runs in the control loop.

**REST** (`daemon/app/models.py`, `daemon/app/routers/state.py`)
- `GET /state/imu` → the reading, or HTTP 200 with `null` on IMU-less robots / stale cache.
- `with_imu: bool = false` flag on `GET /state/full` and the streaming `WS /state/ws/full`, adding an `imu` field to `FullState` — so REST clients can also stream IMU at any requested frequency.

**TS SDK** (`ts/lib/types.ts`, `ts/lib/reachy-mini.ts`)
- `RobotState` gains optional `imu` (reusing the `ImuData` interface #1315 introduced), mirrored from incoming state frames. The mirror clears when the daemon sends `imu: null` (stale cache / no IMU), so consumers never act on a frozen reading — same pattern as `doa`.

### Not in scope

- A dedicated 50 Hz WebRTC IMU data channel. The 30 Hz pose-channel piggyback covers tilt/pickup/motion detection; the `pose` channel (`subscribe_pose`) is the template if a sensor-rate use case appears later.
- IMU configuration/calibration endpoints.
- Legacy WS protocol and Python SDK (`reachy.imu`) — already working, unchanged.

### Testing

New `tests/unit_tests/test_state_imu_api.py` (7 tests): fail-silent nulls against the mockup backend on all three surfaces; positive path with a stubbed reading; and the assertion that the `type` discriminator never reaches REST or snapshot JSON. On the JS side, `imu.test.ts` gains a `robotState.imu` mirror test pinning both the set and the stale-`null` clear. Verified alongside the pose-stream, DoA, protocol-sync, process-command, and WS server/client suites; `npm run typecheck:sdk` and the IMU vitest suite pass. (Pre-existing, unrelated: `test_daemon_wireless_client_disconnection` fails on a clean tree too.)

### Relation to #1315

Complementary: #1315 gave the JS SDK a one-shot `getImu()` over the `get_imu` data-channel command; this PR adds the *streamed* IMU in the 30 Hz state snapshot plus the REST endpoints, backed by the same daemon cache. Rebased on current main now that #1315 has landed — this branch reuses its `ImuData` interface and extends its `imu.test.ts` suite with the state-mirror test.

### DIFF
--- a/docs/source/API/daemon.mdx
+++ b/docs/source/API/daemon.mdx
@@ -180,6 +180,8 @@ released its media handles before the local app opens them.
 
 [[autodoc]] reachy_mini.daemon.app.routers.state.get_doa
 
+[[autodoc]] reachy_mini.daemon.app.routers.state.get_imu
+
 [[autodoc]] reachy_mini.daemon.app.routers.state.get_full_state
 
 [[autodoc]] reachy_mini.daemon.app.routers.state.ws_full_state
--- a/docs/source/API/openapi.json
+++ b/docs/source/API/openapi.json
@@ -2347,6 +2347,33 @@
         }
       }
     },
+    "/api/state/imu": {
+      "get": {
+        "summary": "Get Imu",
+        "description": "Get the latest IMU reading (accelerometer, gyroscope, quaternion, temperature).\n\nAccelerations are in m/s\u00b2, angular rates in rad/s, the quaternion is\nw-first and the temperature in \u00b0C.\nReturns None on versions without an IMU (Lite, simulation) or when the\ncached reading has gone stale.",
+        "operationId": "get_imu_api_state_imu_get",
+        "responses": {
+          "200": {
+            "description": "Successful Response",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "anyOf": [
+                    {
+                      "$ref": "#/components/schemas/ImuData"
+                    },
+                    {
+                      "type": "null"
+                    }
+                  ],
+                  "title": "Response Get Imu Api State Imu Get"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
     "/api/state/full": {
       "get": {
         "summary": "Get Full State",
@@ -2461,6 +2488,16 @@
               "type": "boolean",
               "default": false,
               "title": "With Doa"
+            }
+          },
+          {
+            "name": "with_imu",
+            "in": "query",
+            "required": false,
+            "schema": {
+              "type": "boolean",
+              "default": false,
+              "title": "With Imu"
             }
           },
           {
@@ -3342,6 +3379,16 @@
                 "type": "null"
               }
             ]
+          },
+          "imu": {
+            "anyOf": [
+              {
+                "$ref": "#/components/schemas/ImuData"
+              },
+              {
+                "type": "null"
+              }
+            ]
           }
         },
         "type": "object",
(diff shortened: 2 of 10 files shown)


FabienDanieau:
Re-checked this against current `main` now that #1315, #1347 and #1348 have all landed. The Python half auto-merges with zero OpenAPI drift; after resolving the two TS files locally: ruff/mypy clean, 889 unit tests pass (only hardware-bound failures plus the known `test_daemon_wireless_client_disconnection`), `npm run typecheck` and all 189 vitest tests pass. The merge-order note in the description is obsolete, you can rebase now.

Two changes on the TS side while rebasing:

1. **Drop the `ImuData` interface**: main already has an identical one from #1315 (`ts/lib/types.ts:67`). All that's left to add is `imu?: ImuData;` on `RobotState`.
2. **Mirror the `doa` line in the state handler, not `face_target`**:
   ```ts
   if ('imu' in s) this._robotState.imu = s.imu ?? undefined;
   ```
   The daemon deliberately sends `imu: null` once its 0.5 s cache goes stale, but `if (s.imu)` keeps the last reading forever, so a tilt/pickup consumer can act on a frozen value. Main's handler already clears `doa` this way for exactly that reason.

Nits:
- The 5-line comment on `imu=` in `build_state_snapshot` can be one line; the tests already pin the no-discriminator contract.
- `test_full_state_carries_imu` re-asserts what `test_get_imu_returns_reading` covers; could be folded.
- "byte-identical" in the description isn't quite right: subclassing reorders keys (`type` now serializes last). Nothing consumes key order (`ws_client.py` discriminates by `isinstance`), just worth rewording.
- Optional: #1315 brought `ts/lib/imu.test.ts`, so a two-assert state-mirror test (set on frame, cleared on `null`) is cheap now.

Python/REST side LGTM as-is.

darkseed:
Applied (pushed as d719ae3ed, force-push after rebase):
  - Rebased onto current main the branch now sits on top of #1315/#1347/#1348; the obsolete merge-commit history is gone.
  - Dropped our duplicate `ImuData` verified main's interface is field-identical; RobotState now just gains imu?: ImuData and reuses main's type (which reachy-mini.ts already imported for `getImu()`).
  - The stale-null fix this was the one real bug in the feedback: `if (s.imu)` would have served a frozen reading forever after the daemon's 0.5 s cache nulls out. Now `if ('imu' in s) this._robotState.imu = s.imu ?? undefined;`, mirroring main's doa clearing.
  - Comment shrink in `build_state_snapshot()` five lines down to one.
  - Took the optional vitest suggestion  added a mirror test to `imu.test.ts` (set on frame, cleared on null) it directly pins fix #3. 6/6 pass.
  - PR description "byte-identical" reworded to wire-compatible with the key-order caveat, and the merge-order section replaced with the post-rebase reality

FabienDanieau (approved):
Re-reviewed the reworked branch (rebased on main, 4 commits). All points from my previous comment are addressed:

- `ts/lib/types.ts` reuses #1315's `ImuData` and only adds `RobotState.imu`.
- The state handler mirrors the `doa` pattern (`'imu' in s`, `?? undefined`), so a stale `null` clears the mirror instead of latching a frozen reading, and `imu.test.ts` now pins exactly that.
- Snapshot comment shrunk, duplicate positive-path tests folded.

Verified locally on the branch: IMU pytest suite 6/6, ruff + mypy clean, zero OpenAPI drift, `typecheck:sdk` clean, `imu.test.ts` 6/6. LGTM.

PR #1343, opened 2026-08-11
perf(sdk): replace scipy Rotation with a minimal numpy one (-1.1 s cold import, 2.5x faster interpolation)

### LINKED ISSUE #1342: scipy is imported only for Rotation, and costs ~1.7 s of cold import on the robot
`from scipy.spatial.transform import Rotation` is the SDK's only use of scipy, but importing it eagerly pulls in `scipy.spatial`'s KD-tree, `scipy.sparse`, the array-API compat layer and `numpy.f2py` (a Fortran source parser), costing ~1.25 s warm and ~1.7 s cold on the CM4 for eight operations on 3x3 matrices that are ~250 lines of numpy.

RemiFabre:
## Issue

Closes #1342. Follows up on the review of #1308: rather than lazy-importing scipy, drop it from the SDK's import path entirely.

`import reachy_mini` pays ~1.1 s of cold import on the robot's CM4 for `scipy.spatial.transform.Rotation` alone. Reproduce with `python -X importtime -c "import reachy_mini"` on the robot: `scipy.spatial` eagerly loads a KD-tree, `scipy.sparse`, the array-API compat layer and `numpy.f2py` on the way to what we use — quaternion-scale arithmetic on single 3x3 matrices.

## Description

The SDK uses exactly eight `Rotation` operations, always on a single rotation: `from_euler` (extrinsic, with `degrees=`), `from_matrix`, `from_rotvec`, `as_matrix`, `as_euler("xyz")`, `as_rotvec`, `*` and `.inv()`. Grepped every call site to confirm; no `Rotation` object escapes a public function.

Those eight now live in `utils/rotation.py` (~250 lines of numpy, conventions and input rejections matched to scipy, including the gimbal-lock convention), and the twelve import sites — ten in the SDK, two examples — point at it. scipy moves to the `dev` dependency group, kept purely as the test oracle, so a runtime install sheds 118 MB.

Alternative considered: lazy-importing scipy (#1308) hides the cost until first use; removing it also drops the per-call overhead (scipy's single-rotation path is its worst case) and the 118 MB.

**Breaking**: apps that `import scipy` without declaring it lose the transitive install. One-line fix in the app's `pyproject.toml`; needs a release note.

## Testing

- New `tests/unit_tests/test_rotation_matches_scipy.py`: 14 tests fuzzing every operation against scipy (500 random rotations each, atol 1e-12), including `from_rotvec` at degenerate magnitudes, exact gimbal lock (pitch = ±90°), orthonormalisation of drifted matrices, rejection parity on invalid input, shape/dtype parity, and `linear_pose_interpolation` end-to-end against a scipy-backed reference. scipy stays installed in CI (`dev` group) as that oracle.
- Measured on a wireless robot: cold import of `reachy_mini` −1147 ms, `linear_pose_interpolation` 2.5–2.8x faster per call. Full measurements in the comments below.
- Verified live on the robot on this branch: daemon and gpio-shutdown active, no import errors, `create_head_pose(5, -10, 20)` round-trips through `as_euler` to `[5, -10, 20]` exactly.
- Verified in MuJoCo simulation (`--sim --headless`, daemon and client both on this branch): head driven through 4 distinct poses via `goto_target`, read-back orientation within controller settle error (<2.2°), and `scipy` confirmed absent from `sys.modules` for the whole session.
- Verified on a Reachy Mini Lite, A/B against `main` on the same robot and the same 4-pose sequence: worst read-back error 3.7° on this branch vs 3.5° on `main` (same pattern, worst at neutral — steady-state controller/gravity offset, present identically on both). Differences are within run-to-run noise.
- Full unit suite passes; the two failures on this branch also reproduce on clean `main` (details in comments).

## Tested on

- [x] Reachy Mini Wireless
- [x] Reachy Mini Lite
- [x] MuJoCo simulation (`--sim`)

## AI assistance

Assisted-by: Claude:claude-opus-5
Assisted-by: Claude:claude-fable-5


RemiFabre:
## On-robot validation, main vs this branch

Run on a wireless Reachy Mini (CM4), driven from a laptop over WiFi. Every motion below exercises a call site this PR changes. Both versions were installed on the same robot in turn, and the real joint positions were sampled at 100 Hz throughout each motion.

### 1. The maths is identical (deterministic, no robot)

`linear_pose_interpolation`'s exact rotation pipeline run through scipy and through `utils/rotation.py`, over 40 random pose pairs and the whole of `t ∈ [0, 1]`, for both the `yaw_as_scalar` branch and the geodesic one:

**worst difference across every sample: 1.67e-15 rad**, flat across `t`, three orders of magnitude below the 1e-12 tolerance the unit tests use.

### 2. The robot moves the same way

Twelve motions, joint trajectories overlaid. Travel is the total joint excursion; `p99 step` is the 99th-percentile change between consecutive 100 Hz samples, i.e. how smooth the motion is.

| motion | travel main | travel ours | p99 step main | p99 step ours |
|---|---|---|---|---|
| `goto` linear | 0.407 | 0.405 | 0.0178 | 0.0141 |
| `goto` minjerk | 0.407 | 0.402 | 0.0218 | 0.0156 |
| `goto` ease_in_out | 0.408 | 0.410 | 0.0187 | 0.0175 |
| `goto` cartoon | 0.488 | 0.483 | 0.0402 | 0.0356 |
| yaw −80° → 80° (`yaw_as_scalar`) | 0.612 | 0.617 | 0.0354 | 0.0318 |
| roll +35° | 0.571 | 0.571 | 0.0224 | 0.0187 |
| pitch −35° | 0.414 | 0.416 | 0.0151 | 0.0138 |
| roll/pitch/yaw combo | 0.767 | 0.767 | 0.0371 | 0.0202 |
| `set_target` sine, 100 Hz | 0.706 | 0.707 | 0.0276 | 0.0276 |
| recorded move `dance1` | 0.867 | 0.870 | 0.1201 | 0.1195 |
| recorded move `curious1` | 1.605 | 1.605 | 0.0721 | 0.0735 |
| `look_at` sweep | 0.433 | 0.433 | 0.0299 | 0.0261 |

Travel agrees to ~0.005 rad and the step distributions have the same shape everywhere.

**How to read this table.** These are two separate physical runs, so any difference is the robot, not the code. I deliberately did not include a "max difference between the two runs" column: it comes out at 0.02–0.13 rad, and that is entirely run-to-run timing jitter. Quoting it as an implementation difference would be misleading. Section 1 is the number that says the implementations agree.

**One artefact worth flagging:** in the roll/pitch/yaw combo plot, the *main* trace has a brief flat-then-drop around t ≈ 1.2–1.35 s that the branch trace does not. That is a sampling/network hiccup during the main run, not smoother motion on this branch.

### 3. Threshold checks

An earlier pass asserted, for each of the same motions, that the trajectory is continuous (no jump between consecutive samples that isn't physical motion), that the robot arrives at the commanded pose, and that no joint value is NaN or infinite: **15/15 passed**.

### Coverage of the changed call sites

| call site | covered by |
|---|---|
| `utils/interpolation.py` (`from_matrix`, `as_euler`, `from_euler`, `inv`, `as_rotvec`, `from_rotvec`, `as_matrix`) | all four `goto` methods, both interpolation branches |
| `utils/__init__.py` `create_head_pose` (`from_euler`) | every motion |
| `kinematics/*` (`from_matrix` → `as_euler`) | workspace extremes |
| `motion/recorded_move.py` | `dance1`, `curious1` |
| `vision/look_at.py` (`from_rotvec`) | `look_at` sweep |
| `reachy_mini.py` wake-up roll (`from_euler`, degrees) | `wake_up` |

The driver script is ~150 lines and reproducible against any robot: it takes a label, records every trajectory to an `.npz`, and a second script overlays two runs.

RemiFabre:
rotation_offline_exactness
rotation_robot_trajectories

askuric:
It looks good to me. 
I did not go into the detail of the `as_euler` and `as_rotvec` function, but if the tests pass, I trust that they work well. 
The fucntions `from_matrix`, `from_euler` and `from_rotvec` look good. Although I'd add some comments and references. 
I had to look them up, and some of them are very clever and need some time to understand even with the explanation :D

I'd also add some gimbal lock tests in the test script. 
Even though they are usually outside the reachable space.

RemiFabre:
## Measurements

All on the wireless robot (CM4, 2026-08-11), same venv, `@62a9c1b4e` (main) vs this branch, 5 warm runs and 2 cold runs each with `drop_caches` between.

### Where the time goes on main

`-X importtime` on the robot:

```
scipy.spatial.transform         1064 ms
  scipy.spatial._kdtree          891 ms
    scipy.spatial._ckdtree       539 ms
      scipy.sparse               533 ms
        scipy._lib._array_api    481 ms
          array_api_compat.numpy 429 ms
            numpy.f2py           189 ms
              numpy.f2py.crackfortran  145 ms
  scipy.spatial.distance         347 ms
  scipy.special                  183 ms
  scipy.linalg                   148 ms
```

`scipy.spatial` has no lazy `__getattr__`, so there is no import path that avoids this.

### Import time

| `import reachy_mini` | before | after | saved |
|---|---|---|---|
| warm | 4430 ms | **3527 ms** | 903 ms |
| cold | 5751 ms | **4604 ms** | **1147 ms** |

`scipy in sys.modules` after `import reachy_mini`: `True` before, **`False`** after.

The replacement module is essentially free on top of numpy:

| standalone import | warm |
|---|---|
| numpy only | 398 ms |
| numpy + `scipy.spatial.transform.Rotation` | 1582 ms |
| numpy + `reachy_mini.utils.rotation` | **409 ms** |

### Per-call performance

scipy's `Rotation` is built for batches; the SDK always passes exactly one rotation, which is its worst case (all per-call validation and array-API dispatch, nothing to amortise). `linear_pose_interpolation`, the function in the goto and recorded-move loops:

| | before | after | |
|---|---|---|---|
| `yaw_as_scalar=True` | 2442 us | **983 us** | 2.5x |
| `yaw_as_scalar=False` | 1458 us | **524 us** | 2.8x |

At 100 Hz that is ~24% of a CM4 core spent on interpolation, down to ~10%.

Counterpoint: on *batches* scipy wins by ~56x (3.1 us/rotation amortised over 2000 vs 173 us looped). No call site batches today, but anyone adding one should vectorise it or use scipy there.

### Disk

A runtime install sheds **118 MB** (scipy wheel + its share of the venv).

RemiFabre:
## Regression risk & how it's contained

- **12 tests fuzz every operation against scipy**, 500 random rotations each, absolute tolerance 1e-12 — several orders of magnitude below anything the hardware can express. Covered: `from_euler` in radians and degrees, single-axis scalar, `as_euler`, `as_rotvec`, `from_rotvec` at the awkward magnitudes (0, 1e-12, ~pi, exactly pi), composition, inverse, `from_matrix` orthonormalisation of drifted matrices, and shape/dtype parity (a `(1,3,3)` for a `(3,3)` would slip past `allclose` but break callers).
- **`linear_pose_interpolation` is compared end to end** against a scipy-backed reimplementation of itself, both branches, over the whole path — this is the function that actually runs in the goto and recorded-move loops.
- scipy remains installed *purely* as that oracle (`dev` group; `pytest.yml` already syncs `--group dev`), so an upstream behaviour change or a mistake here surfaces in CI rather than on a robot.
- Conventions matched exactly, including extrinsic lowercase sequences and scipy's gimbal-lock convention of pinning the third angle. Gimbal lock (pitch = ±90°) is unreachable in practice with pitch clamped to ±40°.
- Unsupported input raises rather than silently doing something else (intrinsic sequences, wrong angle counts, batched arrays, other Euler sequences).

CI status: full unit suite passes. The two failures on this branch (`test_discovery` mDNS flake, `test_wireless::test_daemon_wireless_client_disconnection`) both reproduce on clean `main`.

Live verification: robot installed on this branch, daemon and gpio-shutdown both active, no import errors, `create_head_pose(5, -10, 20)` round-trips through `as_euler` to `[5, -10, 20]` exactly and stays orthonormal with det 1.

FabienDanieau (approved):
All good thanks!

PR #1345, opened 2026-08-11
fix(examples): throttle minimal_demo's set_target loop to 100 Hz

### LINKED ISSUE #1344: minimal_demo.py streams set_target unthrottled, making motion laggy over WiFi
The `while True` loop has no sleep, so it emits ~13000 `set_target`/s from a laptop and the movement becomes jerky on a wireless robot

RemiFabre:
Closes #1344. Unthrottled the loop emitted ~13000 `set_target`/s from a laptop against the 50 Hz control loop, which generated a very jerky movement.


andimarafioti (approved):
LGTM!

PR #1346, opened 2026-08-12
feat(daemon+sdk-js): play_recorded_move over the data channel

tfrere:
> Stacked on #1321 (`feat/sdk-staleness-check`). Ports the `play_recorded_move` command out of the wake-up wizard PR (`feat/first-wake-up-on-main`), which carries the canonical version but never landed on main nor on the js-core chain.

## Why this exists

The mobile app's first-wake-up wizard (reachy_mini_mobile_app#80) plays its cues - the wake animation, the knock-knock speaker check - through a single daemon-side primitive: "play this named recorded move, motion and bundled sound in lockstep". That primitive only lives in the wake-up PR. Updating a robot to the js-core chain and re-vendoring the SDK from `feat/sdk-staleness-check` silently dropped both sides of it, and the wizard crashed at runtime with `robot.playRecordedMove is not a function`.

Rather than blocking the wizard on the wake-up PR's timeline, this PR ports the minimal slice onto the chain the app actually runs.

## What it adds

- **Daemon**: `PlayRecordedMoveCmd` on the WebRTC data channel (protocol + backend dispatch/handler) - the data-channel twin of `POST /api/move/play/recorded-move-dataset`. Loads the move cache-first from a HF dataset and runs it through `Backend.play_move`, so the sidecar sound plays on the robot speaker in sync. Fire-and-forget: the ack means "dispatched", not "finished". An optional `initial_goto_duration` eases into the move's first frame instead of snapping (used when replaying the wake move while already awake).
- **Daemon**: `PreloadDatasetCmd`, a generic cache warm-up primitive. A client that plans to play moves from a non-official dataset sends it right after connecting; the daemon runs the (cache-first, hence idempotent) `snapshot_download` in a worker thread and acks with the local path. Fail-open: if the preload failed, `play_recorded_move` still downloads on demand at playback time.
- **JS SDK**: `playRecordedMove(moveName, { dataset?, initialGotoDuration? })` and `preloadDataset(dataset)`, js-core style (return `false` if the channel isn't open; the daemon self-guards against concurrent moves, so a rapid double call is a no-op).

## Design note: who knows about which datasets

The wake-up PR's version had the daemon preloading two personal-namespace onboarding datasets (`Anne-Charlotte/new-emotions`, `tfrere/reachy-mini-onboarding-moves`) in `DEFAULT_DATASETS` - infrastructure hardcoding knowledge of one client feature, fragile to account renames. This PR inverts that dependency: `DEFAULT_DATASETS` is back to the official pollen-robotics libraries only, and the mobile app warms its own onboarding dataset at session start via `preload_dataset`. The daemon stays feature-agnostic; the constants live with the feature that needs them (the app). Merging the onboarding moves into `pollen-robotics/reachy-mini-emotions-library` remains desirable (it would delete the app-side constant too), but is no longer a blocker for this chain.

## Known trade-off

**Duplication with the wake-up PR.** This is a port, not the canonical change. Whichever of the two merges second will hit small, mechanical conflicts on `protocol.py` / `abstract.py` / `recorded_move.py` (this PR's `recorded_move.py` resolution - official libraries only - should win).

## Validation

- Daemon of the previous revision (`0326bcde3`) installed on a physical robot; the wizard's wake animation and speaker check run end-to-end from the mobile app. The `preload_dataset` revision (`489413092`) is pending a robot install (robot currently offline) - the command wraps the pre-existing, already-exercised `preload_dataset()` startup path.
- SDK built, packaged and vendored into the mobile app (`reachy-mini-sdk-489413092` tarball in reachy_mini_mobile_app#80); `tsc` clean, vitest 173/173.

---

*Made with [Cursor](https://cursor.com)*


FabienDanieau (approved):
Approving. Validated on hardware: flashed `feat/wakeup-moves` on the robot and the mobile first-wake-up wizard's step emotes now play over the data channel (`wake-mini-up`, `curious1`, `proud2`, `welcoming2`). Before this branch every one of them was rejected by the command union:

```
WebRTC invalid command: Input tag 'play_recorded_move' found using 'type'
does not match any of the expected tags: ...
```

Reviewed the diff itself — the two new commands, their wiring in `abstract.py`, and the `DEFAULT_DATASETS` restriction. Inverting the dependency so the daemon stops hardcoding one client's onboarding datasets is clearly right, and `preload_dataset` is the correct primitive for it.

## One field note: a stale dataset cache is unrecoverable without manual intervention

After flashing, `play_recorded_move` was accepted but then failed:

```
play_recorded_move: 'proud2' from 'pollen-robotics/reachy-mini-emotions-library'
failed to load: Move proud2 not found in recorded moves library
```

`proud2.json` / `proud2.ogg` are both present on the Hub at the repo root, where `process()` globs. The robot simply had an older cached snapshot. Fixed with:

```
hf cache rm dataset/pollen-robotics/reachy-mini-emotions-library
```

The mechanism is `RecordedMoves.__init__`:

```python
self.local_path = snapshot_download(
    self.hf_dataset_name, repo_type="dataset",
    local_files_only=True,
)
```

It falls back to a network download only on `LocalEntryNotFoundError` — i.e. when there is *no* cache. A snapshot that is merely stale or partial is used forever, so the robot stays pinned to it and reports "not found" indefinitely. `preload_default_datasets()` should paper over this at boot, but `preload_dataset()` swallows every failure into a `logger.warning` and returns `None`, so one flaky boot leaves the old snapshot in place silently.

**This is pre-existing, not introduced here** — I'm raising it because this PR is what makes it load-bearing. Named-move playback from arbitrary datasets over the DC is now a first-class client API, so "the robot has a stale snapshot of a library" goes from a corner case to something every user hits the next time a move is added to the emotions library.

Not blocking, and probably its own PR. Two cheap options:

- On `ValueError` from `get()`, retry once through a networked `snapshot_download` before giving up — turns a permanent failure into a one-off delay.
- Or surface it: have `preload_dataset` failures reach the client (the `preload_dataset` response already carries a status), so a client that warmed the cache knows it didn't work.

Either way, worth a line in the PR body that an existing robot may need its cache cleared — the failure reads as "the move doesn't exist" rather than "your copy is old", which is a slow thing to debug.

tfrere:
pushed a few more commits on top: a stop_move primitive to interrupt whatever move is currently playing (no upload_id needed, also silences the move's bundled sound), recorded-move dataset loading now runs off the event loop (on a cold cache the snapshot_download was blocking the loop and freezing the pose stream), and preloadDatasetAndWait in the js sdk which resolves once the daemon acks the preload. that last one partially addresses @FabienDanieau's note about surfacing preload results to the app - the ack is now exposed to callers. running the robot-side validation today.

PR #1347, opened 2026-08-12
feat(host): update gate, resilience UX, picker/boot refresh (re-land #1316 on main)

tfrere:
## Why

#1316 was part of a stack (`feat/sdk-js-core` <- `feat/host-shell-refresh` <- `feat/sdk-staleness-check`). When #1315 landed on `main`, the child PRs were never retargeted, so #1316 got squash-merged into `feat/sdk-js-core` instead of `main`. GitHub shows it as merged, but none of its changes reached `main`.

This PR re-lands exactly that work on `main`. `main` has been merged into the branch to resolve the single `ts/vitest.config.ts` add/add conflict (the SDK source alias from this branch is kept, since the host tests need it).

## What

Same content as #1316: host update gate, resilience UX, picker/boot refresh (31 files).

## Follow-up

#1321 is in the same situation (squash-merged into `feat/host-shell-refresh`). It will be re-landed in a follow-up PR from `feat/host-shell-refresh` once this one is merged, and #1346 will be retargeted to `main`.


FabienDanieau (approved):
LGTM

PR #1348, opened 2026-08-12
feat(host): SDK staleness self-check v2 (re-land #1321 on main)

tfrere:
## Why

Same story as #1347: #1321 was the third PR of a stack and got squash-merged into `feat/host-shell-refresh` instead of `main`, so none of its changes reached `main`. This PR re-lands it.

#1347 is now merged, so `main` has been merged back into this branch: the diff below is exactly the #1321 changes (8 files, +1264/-13), nothing from #1316 is repeated.

## What

Same content as #1321: SDK staleness self-check v2 (npm source, major-only, persisted dismissal), plus the SDK lifecycle contract doc.

## Follow-up

#1346 is based on `feat/sdk-staleness-check`, itself stacked on this branch, and should be retargeted to `main` once this lands.


FabienDanieau (approved):
LGTM

PR #1351, opened 2026-08-13
fix(update): pick the newest pre-release from PyPI, not the last key

FabienDanieau:
## Issue

With the beta channel enabled, the daemon reports "up to date" even though a newer RC is live on PyPI. `1.10.0rc5` never shows up.

Repro, on a Wireless running the `1.9.0` release:

```console
$ curl 'http://127.0.0.1:8000/update/available?pre_release=true'
{"update":{"reachy_mini":{"is_available":false,"current_version":"1.9.0","available_version":"1.9.0"}}}
```

`1.10.0rc5` is on PyPI, so this should report it as available.

## Description

`get_pypi_version` took `list(data["releases"].keys())[-1]` as the latest pre-release. PyPI orders that mapping **lexicographically**, not by version: `"1.9.0rc1"` sorts after `"1.10.0rc5"`. So the trailing key was `1.9.0rc1`, older than the stable `info.version`, the `pre_version > version` check failed, and the stable release was returned.

Fix: take the `max()` over every parseable key, which does not care about key order. Yanked and file-less releases are skipped, because `pip install --pre` refuses those. Offering one would loop the update flow forever: the check keeps reporting an update that can never install.

### On a Reachy Mini Wireless (CM4), against live PyPI

Before, with the `1.9.0` release installed in `/venvs/mini_daemon` (`1.10.0rc5` already published):

```console
get_pypi_version('reachy_mini', True)  -> 1.9.0        # wrong
is_update_available('reachy_mini', True) -> False
```

After, same robot, daemon running this branch:

```console
$ curl 'http://127.0.0.1:8000/update/available?pre_release=true'
{"update":{"reachy_mini":{"is_available":true,"current_version":"1.10.0-dev.0","available_version":"1.10.0-rc.5"}}}

$ curl 'http://127.0.0.1:8000/update/available?pre_release=false'
{"update":{"reachy_mini":{"is_available":false,"current_version":"1.10.0-dev.0","available_version":"1.9.0"}}}
```

The stable channel is unchanged. `current_version` is `1.10.0-dev.0` because the daemon ran from an editable checkout; forcing the local version to an installed `1.9.0` gives the same verdict:

```console
{"update":{"reachy_mini":{"is_available":true,"current_version":"1.9.0","available_version":"1.10.0-rc.5"}}}
```

### Unit tests

Three regression tests in `tests/unit_tests/test_wireless_version.py`, each failing on the old code:

- `test_get_pypi_version_prerelease_ignores_key_order` covers the bug itself
- `test_get_pypi_version_prerelease_skips_yanked_and_deleted`
- `test_get_pypi_version_prerelease_skips_unparseable` (`.post1`)

```console
$ uv run pytest tests/unit_tests/test_wireless_version.py tests/unit_tests/test_router_update.py -q
52 passed
```

Two existing pre-release tests passed `[]` as their file lists, which the new yanked/deleted guard filters out. They now carry a real file entry.

Full `tests/unit_tests` run: 894 passed, 16 failed. All 16 failures are the pre-existing audio/video/`test_wireless.py` ones that need a board or a running daemon, identical on `origin/main`. `pre-commit` clean; `mypy` reports no error in the touched file.

## Tested on

- [x] Reachy Mini Wireless
- [ ] Reachy Mini Lite

## AI assistance

`Assisted-by: Claude:claude-opus-5`

### DIFF
--- a/src/reachy_mini/utils/wireless_version/update_available.py
+++ b/src/reachy_mini/utils/wireless_version/update_available.py
@@ -55,10 +55,17 @@ def get_pypi_version(package_name: str, pre_release: bool) -> semver.Version:
     version = _semver_version(data["info"]["version"])
 
     if pre_release:
-        releases = list(data["releases"].keys())
-        pre_version = _semver_version(releases[-1])
-        if pre_version > version:
-            return pre_version
+        # PyPI orders ``releases`` lexicographically, so the last key is not the
+        # newest version ("1.9.0rc1" sorts after "1.10.0rc5"). Take the max.
+        candidates = [version]
+        for v, files in data["releases"].items():
+            if not files or all(f.get("yanked") for f in files):
+                continue  # deleted or yanked: pip would refuse to install it
+            try:
+                candidates.append(_semver_version(v))
+            except ValueError:
+                continue  # version scheme we can't compare (e.g. ``.post1``)
+        return max(candidates)
 
     return version
 
--- a/tests/unit_tests/test_wireless_version.py
+++ b/tests/unit_tests/test_wireless_version.py
@@ -226,12 +226,62 @@ def test_get_pypi_version_stable(monkeypatch):
     assert update_available.get_pypi_version("reachy-mini", pre_release=False) == semver.Version.parse("2.0.0")
 
 
+_FILES = [{"yanked": False}]
+
+
 def test_get_pypi_version_prerelease(monkeypatch):
-    """Pre-release lookup prefers a newer trailing release entry."""
-    payload = {"info": {"version": "2.0.0"}, "releases": {"1.9.0": [], "2.1.0rc1": []}}
+    """Pre-release lookup prefers a newer pre-release over the stable one."""
+    payload = {"info": {"version": "2.0.0"}, "releases": {"1.9.0": _FILES, "2.1.0rc1": _FILES}}
     monkeypatch.setattr(update_available.requests, "get", lambda *a, **k: _FakeResponse(payload))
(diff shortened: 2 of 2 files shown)


tfrere (approved):
Tested this as a reviewer, LGTM.

What I checked:

- **Regression test really catches the bug**: reverted `update_available.py` to `main`'s version while keeping the new tests -> `test_get_pypi_version_prerelease_ignores_key_order` fails as expected; with the fix, all 36 tests in `test_wireless_version.py` pass (52 including `test_router_update.py`).
- **Against live PyPI (dev machine)**: `get_pypi_version("reachy-mini", pre_release=False)` -> `1.9.0`, `pre_release=True` -> `1.10.0-rc.5`. On `main`'s code the same call returns `1.9.0` (rc5 invisible), confirming the bug in real conditions.
- **On a real robot (daemon venv, reachy-mini 1.9.0 installed)**: dropped the patched module in and ran it standalone -> `pypi pre = 1.10.0-rc.5`, `is_update_available(pre_release=True) = True`. This unblocks the RC channel end to end (`/update/available?pre_release=true` and the BLE `UPDATE_CHECK PRE` path from #1368).
- Yanked / file-less releases are correctly skipped (pip would refuse them), and unparseable legacy tags (`.post1`) no longer crash the lookup - both strictly better than the old behavior, where a trailing unparseable key would raise.
- `ruff check` and `mypy` clean on the touched files. (`ruff format --check` flags `test_wireless_version.py`, but `main`'s version of the file fails it too - pre-existing, not this PR.)

Note: this supersedes #1369 (closed as duplicate); this version handles yanked/deleted releases, which mine didn't.

PR #1355, opened 2026-08-14
feat(sdk-js): resolve a recorded move on the daemon's dispatch ack

tfrere:
## Issue

No issue filed; found while building a demo app that plays moves from arbitrary community datasets.

`playRecordedMove()` is fire-and-forget: its `boolean` says the command reached the data channel, nothing more. An app that wants to gate its UI on "the move actually started" has to watch `is_move_running` with a deadline of its own, and that misreads a cold dataset as a failure.

Reproduction, on a robot whose HF cache doesn't have the dataset yet:

1. `robot.playRecordedMove('some-move', { dataset: 'someone/their-moves' })`
2. Watch `is_move_running` for a couple of seconds, as any UI would.
3. Nothing is running, so the app reports the move never started - and then the robot plays it ten seconds later, once the download finishes.

The daemon is not at fault: `play_recorded_move` builds `RecordedMoves`, which `snapshot_download`s the dataset when it isn't cached, and only then starts playback. There is no deadline based on a data-channel round trip that can tell that apart from a bad move name.

## Description

The signal was already on the wire. The daemon acks the dispatch with `status: "ok"` once the move is loaded, or `status: "error"` when it couldn't load it - the SDK just dropped it on the floor.

`playRecordedMoveAndWait()` surfaces it, mirroring `preloadDatasetAndWait()` right next to it: `true` when the move was loaded and handed to the player, `false` when the daemon could not load it (unknown name, missing dataset), `null` on the fail-open timeout, and a rejection when the channel is closed. `timeoutMs` defaults to 120 s because it has to cover that download, not a round trip.

It is built on the existing public `request()` helper, so it adds no new waiting machinery - the whole method is a payload plus a `status === 'ok'` check. The payload itself is factored out so the two variants can't drift apart.

One subtlety worth a reviewer's eye: the reply is matched on the `command` echo **alone**, not on `move_name`. The daemon's error ack carries only `command`, `status` and `error` (`abstract.py`), so a matcher keyed on the move name - the obvious thing to write, and what `preloadDatasetAndWait` does with `dataset_name` - would miss every failure and report a load error as a timeout. There is a test pinning that exact case.

The fire-and-forget variant stays as-is: it is the right primitive for a warm dataset, and nothing needs to change to keep working.

The other two commits are docs: one spells out what the ack does *not* promise (see below), the other fills in `stopMove`, `preloadDataset` and `preloadDatasetAndWait`, which have been on main for a while but were missing from the primitives table.

## What this does not fix

Both are pre-existing daemon behaviours, not regressions, but a primitive named `...AndWait` invites a stronger reading than the ack supports, so they are documented on the method:

- **`true` doesn't mean the robot is moving.** The ack is sent *before* `play_move`, which returns silently when another move is already running. So a play issued mid-move is acked `ok` and then dropped. Callers that must win call `stopMove()` first (the mobile app already does). Making the daemon ack after the concurrency guard, or reply `status: "ignored"`, would be the real fix - separate PR, and it needs a daemon-version story.
- **There is no end-of-playback signal for recorded moves**, unlike `play_uploaded_move` which broadcasts both start and end. Apps still watch `is_move_running` fall, or size a timer. This PR at least lets them start that timer from the ack instead of from the JS call.

## Testing

`ts/lib/recorded-move.test.ts`, 9 cases, following the style of `request.test.ts` (fake data channel + fake timers): both variants put the same payload on the wire, an `ok` ack resolves `true`, an error ack **without `move_name`** resolves `false`, an ack landing 30 s in is still honoured, no answer fail-opens to `null`, and a closed channel rejects.

I checked the failure test actually bites: tightening the matcher to `m.move_name === moveName` makes it fail rather than pass silently.

```
$ npx tsc --noEmit && npx vitest run
TypeScript: No errors found
PASS (199) FAIL (0)
```

Exercised on a Reachy Mini Wireless, driving this branch's SDK from a browser harness over a real WebRTC session:

- **Daemon 1.10.0rc5** (knows `play_recorded_move`, error acks still anonymous - the old-daemon arm of the matcher): valid cached move -> `true` (robot played it), unknown move -> `false` via the status-only arm, and the concurrent unknown+valid case never resolves the unknown move `true` (the review's cross-match repro). The documented anonymous-error residual was observed live (a wrong `false`/`null`, never a wrong `true`) - exactly what the sibling daemon commit closes on daemons that carry it.
- **Daemon 1.9.0** (predates the command entirely; rejects it as an invalid tag and never acks): all three calls fail-open to `null` after `timeoutMs`, no crash, session stays healthy - the contract's fail-open behaviour on pre-1.10 daemons.
- The daemon commit itself was exercised directly (branch code, fake `send_response`, failing dataset load): the error ack now carries `{"command": "play_recorded_move", "status": "error", "move_name": ..., "error": ...}`.

## Tested on

- [x] Reachy Mini Wireless
- [ ] Reachy Mini Lite

## AI assistance

- [ ] No AI involvement

---

## Follow-ups after the next npm release

Both call sites below are pinned to a published SDK, so they can only move once this ships. Listing them here so they don't get lost:

- **JS demo app** - the recorded-moves panel now lets you pick any dataset tagged `reachy_mini_community_moves`, so cold datasets are the common case, and it currently shows "no move started" on a perfectly fine download. It should await the ack instead of its fixed dispatch grace.
- **Mobile app (first wake-up wizard)** - `MEET_ME_REVEAL_MS`, `CAMERA_PLAY_MS`, `SPEAKER_PLAY_MS` and `RESET_AFTER_MOVE_MS` are all hand-sized around a guessed dispatch latency, precisely because there was no completion ack (the comments say so). Starting those timers from the ack rather than from the JS call removes the guessed margin - the one that was already found too tight once at 300 ms. `sdk-types.ts` needs the method added as optional there.


FabienDanieau (changes requested):
Nice addition — the tri-state contract mirroring `preloadDatasetAndWait` is the right shape, and the `recordedMoveCmd()` extraction is the right anti-drift move. One real bug to fix before merge, then a few cuts that make the diff shorter.

**Bug: acks cross-match whenever two `play_recorded_move` commands are in flight.** Reproduced on a robot. The predicate is only the command echo, and `PendingReplies.matchBroadcast` hands an ack to the *newest* matching waiter, so two concurrent calls swap results:

```
send concurrent [unknown_move, no1]
ACK {"command":"play_recorded_move","status":"error","error":"Move unknown_move not found..."}
ACK {"command":"play_recorded_move","status":"ok","move_name":"no1"}
result: unknown_move -> true | no1 -> false   // both wrong
```

Same theft with a fire-and-forget `playRecordedMove` in flight, or a late ack from a timed-out call. It's timing-dependent (unknown-move lookup ~2s vs ~1.7s cached load on the robot), so it'll be flaky in the wild.

Fix, no new plumbing — `request()` already takes `match`:

```ts
match: (m) => m.command === 'play_recorded_move'
    && (m.status === 'error' || m.move_name === moveName),
```

Verified on a robot: with this predicate the wrong-`true` never reproduces (all orderings), and a foreign ok ack arriving first is correctly rejected by the waiter. The existing test suite passes unchanged. The inline comment saying a stricter predicate "would miss every failure" only holds for the fully-strict form — this OR-form misses none; the test header (lines 8-12) states the same claim and needs the same update. One residual stays, also observed: an anonymous error ack can still be claimed by the wrong waiter (a wrong `false`/`null`, never a wrong `true`), because the daemon omits `move_name` on error acks. Adding `"move_name": cmd.move_name` there in `_async_play_recorded_move` is additive, mirrors `preload_dataset`'s error ack, and closes it completely — worth a sibling commit here or a follow-up.

**Cuts** (~-30 lines, none blocking):

- `recorded-move.test.ts:74-85` — the "same payload on the wire" test compares `recordedMoveCmd()`'s output to itself through two non-mutating paths; it structurally can't fail. The shared helper is the guard — drop the test.
- `types.ts:533-550` — 18-line jsdoc restating the impl doc. Condense to the 8-line contract summary `preloadDatasetAndWait` uses; keep the full rationale at the impl only.
- `reachy-mini.ts:1101-1117` — destructure + 5-line inline type only to rebundle two fields for `recordedMoveCmd`. Take `opts` whole: `this.request(recordedMoveCmd(moveName, opts), { timeoutMs: opts.timeoutMs ?? 120000 })`.

**FYI, pre-existing:** a handled `false` also fires the SDK-global `error` event (`_handleRobotMessage` emits before `matchBroadcast`). Not this PR's fault, but this API makes it a routine outcome — maybe worth an issue. Same for the test-harness `makeInstance` copy, now at four suites — a shared `test-helpers.ts` would be a nice standalone cleanup.

tfrere:
Thanks for the repro, that cross-match was a real one. All addressed:

- matcher is now the OR-form you verified: command echo + (error status or move_name match). Updated the inline comment and the test header, the "would miss every failure" claim only held for the fully-strict form.
- added a concurrency regression test: two in-flight calls (unknown move acked error, good one acked ok with its move_name), each resolves from its own ack in both ack orders. The ok-first order fails on the old bare-echo matcher.
- took all three cuts: dropped the same-payload test, condensed the types.ts jsdoc to the preloadDatasetAndWait shape, opts passed whole.
- sibling daemon commit adds move_name to the play_recorded_move error ack, mirroring preload_dataset, so the anonymous-error residual is closed on new daemons. Older daemons still send anonymous errors and the OR-form covers them.

ts suite: 199 passing (was 198, -1 deleted test, +2 concurrency cases). tsc, ruff check and format clean.

PR #1360, opened 2026-08-16
fix(apps): scrub PYTHONPATH and GST_PYTHONPATH_1_0 from app subprocess env

### LINKED ISSUE #1359: App subprocesses inherit the daemon's PYTHONPATH and GST_PYTHONPATH_1_0, breaking apps_venv isolation
### Summary

`AppManager.start_app` scrubs GStreamer/XDG variables from the app subprocess environment, but not `PYTHONPATH` or `GST_PYTHONPATH_1_0`. Both are set by the daemon's own `gstreamer_bundle.pth` and point into the **daemon's** `site-packages`. Because both take precedence over the target interpreter's own `site-packages`, an app launched with `apps_venv`'s Python still imports `reachy_mini` and `gi` from the daemon's venv.

This is the same class of bug the existing scrub list was written to fix (referenced there as pollen-robotics/reachy-mini-desktop-app#185) — these two variables were just missed.

https://github.com/pollen-robotics/reachy_mini/blob/main/src/reachy_mini/apps/manager.py#L189-L200

### Impact

Silently invisible while the daemon and `apps_venv` hold the *same* SDK version, because importing from the wrong venv lands on identical code. It breaks as soon as they diverge — which is the normal case for an app pinned to a newer SDK than the installed daemon.

Concretely: `reachy_mini_conversation_app` requires `reachy-mini>=1.10.0rc2`, while the newest **stable** release is `1.9.0`. On a stable daemon the app installs fine into `apps_venv` and then cannot start.

### Reproduction

1. Create a venv with `reachy-mini[mujoco]==1.9.0` (latest stable).
2. Start the daemon so apps get an isolated venv:
   ```
   mjpython -m reachy_mini.daemon.app.main --sim --desktop-app-daemon
   ```
3. Install the conversation app via `POST /api/apps/install`. It lands in `apps_venv` with `reachy-mini==1.10.0rc5`, as intended.
4. `POST /api/apps/start-app/reachy_mini_conversation_app`.

The app exits 1. Every frame of the traceback is in `apps_venv`, except the SDK itself, which resolves to the daemon's venv:

```
File ".../apps_venv/lib/python3.12/site-packages/reachy_mini_conversation_app/main.py", line 128, in run
  from reachy_mini_conversation_app.console import LocalStream
File ".../apps_venv/lib/python3.12/site-packages/reachy_mini_conversation_app/console.py", line 18, in <module>
  from reachy_mini.io.jsonrpc import JsonRpcError
ModuleNotFoundError: No module named 'reachy_mini.io.jsonrpc'
```

`reachy_mini.io.jsonrpc` exists in 1.10.0rc5 but not in the daemon's 1.9.0.

Adding `PYTHONPATH` to the scrub tuple fixes that import and reveals the second leak:

```
File ".../apps_venv/lib/python3.12/site-packages/reachy_mini/vision/face_tracking.py", line 11, in <module>
  import gi
ModuleNotFoundError: No module named 'gi'
```

### Isolating the second variable

`gi` is present in `apps_venv`, and `GST_PYTHONPATH_1_0` alone determines whether it is importable:

```console
$ env -u PYTHONPATH GST_PYTHONPATH_1_0="<daemon_venv>/lib/python3.12/site-packages/gstreamer_python/lib/python3.12/site-packages" \
    ./apps_venv/bin/python -c "import gi"
ModuleNotFoundError: No module named 'gi'

$ env -u PYTHONPATH -u GST_PYTHONPATH_1_0 ./apps_venv/bin/python -c "import gi; print('ok')"
ok
```

### Suggested fix

```diff
                 "XDG_DATA_DIRS",
                 "XDG_CONFIG_DIRS",
+                "PYTHONPATH",
+                "GST_PYTHONPATH_1_0",
             ):
                 app_env.pop(key, None)
```

Both are re-established for the child by `apps_venv`'s own `gstreamer_bundle.pth` at interpreter startup, matching the rationale already documented for the existing entries. With both added, the app starts through `/api/apps/start-app` and reaches `state: running`.

### Workaround

Launch the app directly, bypassing the daemon launcher:

```bash
env -u PYTHONPATH -u GST_PYTHONPATH_1_0 ./apps_venv/bin/reachy-mini-conversation-app --ui
```

### Environment

- macOS 15 (Darwin 25.6.0), Apple Silicon
- Python 3.12.12
- `reachy-mini` 1.9.0 (daemon) / 1.10.0rc5 (`apps_venv`)
- Daemon: `--sim --desktop-app-daemon`
- Scrub list verified unchanged on `main` and in the 1.10.0rc5 wheel

lijunzh:
Fixes #1359.

### Problem

`AppManager.start_app` scrubs GStreamer/XDG variables from the app subprocess environment, but not `PYTHONPATH` or `GST_PYTHONPATH_1_0`. Both are set by the daemon's own `gstreamer_bundle.pth` and point into the **daemon's** `site-packages`, and both take precedence over the target interpreter's own `site-packages`. An app launched with `apps_venv`'s Python therefore imports `reachy_mini` and `gi` from the daemon's venv.

These belong to the same set the existing scrub list already handles (added for pollen-robotics/reachy-mini-desktop-app#185) — they were just missed.

### Why it matters

Invisible while the daemon and `apps_venv` hold the same SDK version, because importing from the wrong venv lands on identical code. It breaks as soon as they diverge — the normal case for an app pinned to a newer SDK than the installed daemon.

`reachy_mini_conversation_app` requires `reachy-mini>=1.10.0rc2` while the newest stable is `1.9.0`. On a stable daemon it installs into `apps_venv` correctly and then cannot start:

```
ModuleNotFoundError: No module named 'reachy_mini.io.jsonrpc'   # PYTHONPATH leak
ModuleNotFoundError: No module named 'gi'                        # GST_PYTHONPATH_1_0 leak (after the first is fixed)
```

`GST_PYTHONPATH_1_0` isolated on its own:

```console
$ env -u PYTHONPATH GST_PYTHONPATH_1_0="<daemon_venv>/.../gstreamer_python/lib/python3.12/site-packages" \
    ./apps_venv/bin/python -c "import gi"
ModuleNotFoundError: No module named 'gi'

$ env -u PYTHONPATH -u GST_PYTHONPATH_1_0 ./apps_venv/bin/python -c "import gi; print('ok')"
ok
```

### Change

Adds both names to the existing scrub tuple, with a comment explaining why. `apps_venv`'s own `gstreamer_bundle.pth` sets both afresh at interpreter startup, matching the rationale documented for the existing entries.

### Test

Adds `test_start_app_scrubs_leaking_env_vars`, which stubs `create_subprocess_exec` and asserts the leaking variables are absent from the child environment while a neutral one (`PATH`) is still forwarded. Verified it fails on `main` and passes with this change.

### Verification

With the fix applied, `POST /api/apps/start-app/reachy_mini_conversation_app` reaches `state: running` against a daemon whose SDK differs from `apps_venv`, and the robot moves. Checked on macOS 15 (Apple Silicon), Python 3.12.12, daemon in `--sim --desktop-app-daemon`.

`ruff check` passes repo-wide.

### DIFF
--- a/src/reachy_mini/apps/manager.py
+++ b/src/reachy_mini/apps/manager.py
@@ -186,6 +186,15 @@ class AppManager:
             #   * The app ends up using .venv's plugin scanner binary and registry cache,
             #     which can mask issues specific to apps_venv's own gstreamer install.
             # See pollen-robotics/reachy-mini-desktop-app#185.
+            #
+            # PYTHONPATH and GST_PYTHONPATH_1_0 belong to the same set and are the
+            # most damaging members: they point at the daemon's own site-packages and
+            # take precedence over the target interpreter's. An app launched with
+            # apps_venv's Python therefore imports `reachy_mini` and `gi` from the
+            # *daemon's* venv, so an app pinned to a different SDK than the daemon
+            # fails at import time (ModuleNotFoundError for `reachy_mini.io.jsonrpc`
+            # and `gi` respectively). apps_venv's own gstreamer_bundle.pth sets both
+            # afresh at interpreter startup.
             app_env = os.environ.copy()
             for key in (
                 "GST_PLUGIN_PATH_1_0",
@@ -196,6 +205,8 @@ class AppManager:
                 "PYGI_DLL_DIRS",
                 "XDG_DATA_DIRS",
                 "XDG_CONFIG_DIRS",
+                "PYTHONPATH",
+                "GST_PYTHONPATH_1_0",
             ):
                 app_env.pop(key, None)
 
--- a/tests/unit_tests/test_app.py
+++ b/tests/unit_tests/test_app.py
@@ -9,6 +9,7 @@ import uvicorn
 from reachy_mini import ReachyMiniApp
 from reachy_mini.apps import AppInfo, SourceKind
 from reachy_mini.apps.manager import AppManager, AppState
+from reachy_mini.apps.sources import local_common_venv
 from reachy_mini.daemon.app.main import Args, create_app
 from reachy_mini.daemon.daemon import Daemon
 from reachy_mini.reachy_mini import ReachyMini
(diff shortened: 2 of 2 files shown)


pierre-rouanet (approved):
LGTM, thanks @lijunzh !

PR #1363, opened 2026-08-17
Stop leaking provider text and credentials

### LINKED ISSUE #1361: OAuth failure text is rendered as same-origin HTML, and the browser logs the credential bundle
When Hugging Face returns an OAuth error, the daemon puts the provider's `error_description` straight into the result page it serves from its own origin, unescaped. The same raw string is stored on the session and handed back by the status poll, so a crafted callback URL reaches the browser twice: once as markup, once as json. Nearby messages are worse – one is built from the entire decoded token response, so a partial response could put token material into an HTTP reply.

Separately, the host writes the iframe URL to debug output, and that URL carries the credentials hash holding the user's bearer token and app configuration. The creds decoder also logs its raw input when decoding fails.

alozowski:
## Issue

Closes https://github.com/pollen-robotics/reachy_mini/issues/1361

## Description
This PR stops a failed Hugging Face login from leaking provider text and credentials into pages, API responses, and logs. Failures now return one of five fixed strings, and the detail stays in the daemon log as an exception type or an HTTP status. The callback ignores `error_description` and escapes what it inserts, so the provider can no longer put HTML into a page on the daemon's own origin. The access log replaces the OAuth callback query, which carried the authorization code. The browser stops logging the iframe URL that carries the user's bearer.

## Testing
Tested on a wireless unit against live Hugging Face. A real session with a bogus code made the token endpoint return 400, and the response body never reached the page, the session record, or the log, which showed only the HTTP status. A script tag in `error_description` never reached the page either. An invalid save-token returned a fixed
message, no token info appeared in the log, and the stored credential still worked afterwards.
test now covers the filter that removes it.

## Tested on

- [x] Reachy Mini Wireless
- [ ] Reachy Mini Lite

## AI assistance

- [ ] No AI involvement
tests were autogenerated


FabienDanieau:
**LGTM once the one residual leak below is closed.**

The premise holds: the daemon API has no auth (CORS only) and on wireless `/logs/ws/daemon` streams `journalctl -f` to anyone on the LAN, so the log is a public surface.

### Tested on a Wireless against live HF

Source diff applied onto `v1.10.0`, deployed, tested, restored to stock (md5-verified).

Both bugs reproduce on stock: `<p><script>MARKER</script></p>` rendered raw, and `?code=…` verbatim in the access log.

After the patch: `error_description` never reaches the page or session (`Authorization was denied.` / `Authentication failed. Please try again.`), the callback shows as `oauth/callback?<redacted>`, other routes keep their query, `relay-status` is gone from INFO, zero markers in the journal. With a real `state` and a bogus code HF answered 400; only `[HF Auth] OAuth token exchange returned HTTP 400` landed in the log, and the stored token survived. Locally: 66 tests pass, ruff (pre-commit selection), format and mypy clean.

### Before merge

`routers/hf_auth.py:135`: `detail=f"Failed to refresh relay: {e}"` is the same leak in the same router; the warning above it already keeps the detail. One-line fix.

### Optional, about -34 lines

- `hf_auth.py:52-58`: `AUTHENTICATION_FAILED` and `AUTHENTICATION_UNAVAILABLE` both tell the user "retry". Merge them.
- `main.py:528-536`: uvicorn's path is always `args[2]` (your test hardcodes it), so no need to scan all args:
  ```python
  if isinstance(record.args, tuple) and len(record.args) > 2:
      path = record.args[2]
      if isinstance(path, str) and path.startswith(_OAUTH_CALLBACK_PATH + "?"):
          record.args = (*record.args[:2], _OAUTH_CALLBACK_PATH + "?<redacted>", *record.args[3:])
  ```
- `test_apps_hf_auth.py:202-227`: `MagicMock` (already imported) does async CMs, no need for the 26-line `_Response`/`_Session`:
  ```python
  resp = MagicMock(status=status)
  resp.text = AsyncMock(return_value=body)
  sess = MagicMock()
  sess.__aenter__.return_value = sess
  sess.post.return_value.__aenter__.return_value = resp
  monkeypatch.setattr(hf_auth.aiohttp, "ClientSession", lambda *a, **k: sess)
  ```
- `hf_auth.py:283-334`: four identical set-status-and-return blocks; a small `_fail(session, message)` collapses each to one line.

### Nits

- `_OAUTH_CALLBACK_PATH` duplicates the two router prefixes and the test hardcodes the same literal, so a prefix move would silently disable redaction. An assert against `app.routes` makes it rot-proof.
- TS: dropping `err` from `console.warn` is right, dropping the whole call leaves a bad creds hash undiagnosable. Keep the warn without the second argument, and assert one argument rather than no call.
- PR description: drop the two dangling fragments (`"test now covers the filter that removes it."`, `"tests were autogenerated"`).

### Separate ticket

Unauthenticated `/logs/ws/daemon` deserves its own issue (may already be covered by the ghsa authbypass work).

<sub>Review assisted by Claude (claude-opus-5); on-robot testing and conclusions checked by me.</sub>

alozowski:
Fixed callback log matching, restored a safe browser warning, and removed duplicated auth handling and unnecessary test scaffolding. I also found and fixed provider-text leaks in Hub refresh and relay logs, unhandled background notification failures, and incorrect device-code expiry classification

FabienDanieau:
**Previous points all addressed, blocking item closed.** The filter rewrite is better than what I suggested: exact path match also fixes the old substring downgrade of any line containing `/health-check`.

Both new fixes check out against `huggingface_hub` 1.20.1: `OAuthErrorCode` is a `str` enum so `error_code == "expired_token"` works, and `_auth.py:322` does interpolate `({e})` into a WARNING.

Re-tested on a Wireless with only `main.py` deployed: callback redacted (with and without trailing slash), ordinary queries preserved, polling routes gone from INFO, zero markers in the journal. 89 relevant tests pass, ruff/format/mypy clean.

### Two new findings

- `_hub_auth_log_filter` catches every WARNING+ on that logger, including `_auth.py:390` (`Error parsing stored tokens file`), which has no provider text. Gate on the refresh path, not the level.
- `_notify_relay_of_token_change` drops the `create_task` handle; the task can be GC'd mid-flight and `_notify_relay` swallows everything, so the loss is silent. `save_token` is an async route, so this path is live. `DeviceCodeSession.task` already guards the same footgun.

### Scope

218/63 became 669/482; 100 of the 222 deleted source lines are comments, plus an `Optional` to `X | None` sweep. Inlining `_persist_device_oauth_token` also drops the note that `_save_oauth_token` is private, range-pinned and contract-tested. Per `AGENTS.md`, I would drop the comment/typing churn and cherry-pick the relay-containment and device-code fixes into their own PRs. Not blocking.

### Complexity

- `test_daemon_logging_config.py:L126-183`: a real uvicorn server, twice, to test a pure `LogRecord` transform. `log_config=None` skips uvicorn's `dictConfig`, so it proves nothing a hand-built record does not. Restore `_access_record` plus one trailing-slash case. ~-45
- `L35-73`: 36-line manual save/restore of four loggers. `monkeypatch.setattr(logger, "filters", [])` already does this. ~-20
- `test_apps_hf_auth.py:L259`: `relay_fails` parametrize duplicates the test at L229. ~-6
- `L185`: both error codes hit the same branch. ~-2
- `AUTHENTICATION_FAILED` / `AUTHENTICATION_UNAVAILABLE` still say the same thing to the user. ~-4

Keep the hub-refresh test: it is the only thing proving the logger name is right.

<sub>Review assisted by Claude (claude-fable-5-1); on-robot testing and conclusions checked by me.</sub>

PR #1365, opened 2026-08-17
Require trusted https endpoints for signaling, central, and TURN

### LINKED ISSUE #1364: Signaling, central, and TURN endpoints accept untrusted destinations for bearer tokens
`resolveSignalingUrl()` in `ts/host/src/lib/signalingUrl.ts` accepts any `?signaling_url=` value or `SIGNALING_URL` space variable that starts with `http`, and the host then attaches the browser's Hugging Face bearer to that origin. A crafted link to the Space sends authenticated traffic to a host the sender controls, plaintext `http://` passes the same check.

The robot side has the same gap in `central_signaling_relay.py`, `webrtc_utils.py`, and `central_consumer.py` (all under `src/reachy_mini/media/`): the configured central, TURN, and consumer URLs are used as given, with no scheme or host check, and bearer-carrying requests follow redirects. Either way the token can reach a third-party origin or cross the network unencrypted.

alozowski:
## Issue

Closes https://github.com/pollen-robotics/reachy_mini/issues/1364

## Description

Checking that an override URL looks valid doesn't help, because an attacker's URL can also look valid, so `?signaling_url=` is now used only when the page itself runs on localhost. Local development still works, and on a deployed Space the parameter is ignored. That leaves two trusted sources: 
- the `SIGNALING_URL` set by the space author
- and the built-in default.

Both are checked for scheme, credentials, query, and fragment.

The robot needs the same protection. A new validator in `utils/network.py` applies the same rules, and also normalises the base path. It is used everywhere a configured endpoint receives the daemon token: the relay, the TURN refresher, the central consumer, and central-robot-status. Every request that carries the token now refuses redirects.

A rejected URL should degrade without breaking things, so a bad TURN URL turns the relay off instead of breaking the media server, and the route checks the URL on each request, so a bad `REACHY_CENTRAL_URL` cannot stop the daemon from starting.

The only behaviour removed is that query parameter on a deployed space, but neither client uses it.

## Testing

Tested on a wireless against live Hugging Face. Nothing was refused with the default endpoints, the central proxy returned 200 with the real robot payload, and `/send` succeeded, so disabling redirects doesn't break the live central Space. After a fresh login the relay reached "Remote access enabled" without a restart. With plaintext `REACHY_CENTRAL_URL` and `REACHY_TURN_URL` the relay refused to start, `TURN` was disabled, the daemon stayed up, and central-robot-status returned `invalid_configuration`.

## Tested on

- [x] Reachy Mini Wireless
- [ ] Reachy Mini Lite

## AI assistance

- [ ] No AI involvement


FabienDanieau (approved):
LGTM, tested on a Wireless (1.10.0 + these files):

- Production central over HTTPS: relay connects, robot shows up in `central-robot-status`, no regressions from `allow_redirects=False`.
- `REACHY_CENTRAL_URL=http://...`: one clean warning, relay skipped, daemon stays up (`NRestarts=0`), route returns `invalid_configuration`.

Optional, none blocking:

1. `hf_auth.py`: `CENTRAL_SIGNALING_SERVER` is a module constant, but it's re-validated on every poll of `central-robot-status`, and the desktop app polls it continuously. With a bad URL that's a warning every few seconds for the daemon's lifetime (confirmed: 3 requests, 3 warnings). Validating once where the constant is defined (try/except → `str | None`) and having the route check `is None` fixes both.
2. `webrtc_utils.py:175` and `central_consumer.py:709`: `raise_for_status()` doesn't raise on 3xx, so a redirect now falls through to `.json()` and surfaces as a decode error, then takes the fast-retry path. A `300 <= status < 400` branch would name the real cause.
3. `central_consumer.py`: `__init__` `Raises:` and the `from_env()` docstring still say the only failure is an empty token / `None` return; callers checking `is None` now get a `ValueError` on a plaintext URL. Fail-closed is right, docs just need to say so.
4. `signalingUrl.ts`: `?signaling_url=` on a hosted page is dropped silently; a `console.warn` would save someone with a staging bookmark some head-scratching. Also `new URL(window.location.origin).hostname` in a try/catch is just `window.location.hostname`, and `URLSearchParams` never throws on a string, so both try/catches can go.
5. `network.py:31`: `urlsplit().hostname` is already lowercase, `.lower()` is redundant.

alozowski:
Addressed all comments, should be fine

PR #1367, opened 2026-08-17
Daemon credential store

### LINKED ISSUE #1366: The daemon stores its Hugging Face credentials in the user's own credential store
The robot has no credential store of its own. Manual entry, device-code login, and redirect OAuth each write the token a different way, and every read goes through the library's ambient lookup, so the robot's credentials are being written in the same store as the user's personal tokens, with no way to tell which entry is ours and nowhere to keep refresh data we own (`src/reachy_mini/apps/sources/hf_auth.py`).

Consequences:
- signing the robot out logs out every saved token on the machine, so users lose credentials they never meant to touch
- cancelling a login isn't final: a flow already in progress still saves the token and marks the cancelled session authorized
- the relay re-reads the token only when it reconnects, so a token that expires mid-session is never recovered and the robot stays registered with central but unreachable until a restart
- every Hub call the daemon makes authenticates as whoever set HF_TOKEN or ran hf auth login, not as the robot
- redirect OAuth asks for write-repos, manage-repos, and inference-api when the robot only reads at runtime

alozowski:
## Issue

Closes https://github.com/pollen-robotics/reachy_mini/issues/1366

## Description
- first of all, all the dead code in `hf_auth.py` was removed;
- the daemon now stores its HF credentials in a dedicated file with `0600` permissions (`mkstemp()` in `_write_store()` function). Manual token entry, redirect OAuth, and device-code OAuth all write to this store, and every daemon credential read uses it;
- sign-out writes a signed-out record and advances a lifecycle generation instead of calling `logout()`. This prevents an in-flight login from restoring credentials after sign-out while leaving the user’s CLI credentials untouched. Cancelling an OAuth session is also final: cancellation and credential persistence are coordinated so a late token response cannot be saved;
- all HF calls now choose their credentials explicitly. Private resources use the daemon-owned token, while the public emotions, dances, and face-detector repositories explicitly disable authentication. The daemon therefore never inherits credentials from `HF_TOKEN` or `hf auth login`;
- the central relay uses the same daemon-owned credentials and refreshes them after a `401`, allowing the robot to reconnect without restarting;
- there is intentionally no migration from the Hugging Face CLI token file or `HF_TOKEN`. 

Note, that after upgrading, an already linked robot appears signed out once and must be linked again.

## Testing

Tested on a wireless against live Hugging Face. Starting from signed out, a desktop-app login reached "Remote access enabled". The new file held a refresh token with a 30-day expiry at mode 600. The unit's existing token file wasn't read and stayed unchanged.

## Tested on

- [x] Reachy Mini Wireless
- [ ] Reachy Mini Lite

## AI assistance

- [ ] No AI involvement

### DIFF
--- a/src/reachy_mini/apps/sources/hf_auth.py
+++ b/src/reachy_mini/apps/sources/hf_auth.py
@@ -367,6 +367,7 @@ async def exchange_code_for_token(
             session.status = "error"
             session.error_message = AUTHENTICATION_FAILED_MESSAGE
             return {"status": "error", "message": session.error_message}
+        token_data["access_token"] = access_token
 
     except Exception as error:
         logger.warning(
@@ -377,9 +378,13 @@ async def exchange_code_for_token(
         return {"status": "error", "message": session.error_message}
 
     try:
-        landed = _persist_login(
-            _token_fields(token_data, None), session.lifecycle_generation
-        )
+        with _store_lock:
+            if _oauth_sessions.get(session.session_id) is not session:
+                landed = False
+            else:
+                landed = _persist_login(
+                    _token_fields(token_data, None), session.lifecycle_generation
+                )
     except OSError as error:
         logger.warning(
             "[HF Auth] Could not save OAuth credentials (%s)", type(error).__name__
@@ -451,10 +456,8 @@ def get_oauth_session_status(session_id: str) -> dict[str, Any]:
 
 def cancel_oauth_session(session_id: str) -> bool:
     """Cancel an OAuth session."""
-    if session_id in _oauth_sessions:
-        del _oauth_sessions[session_id]
-        return True
-    return False
+    with _store_lock:
+        return _oauth_sessions.pop(session_id, None) is not None
 
 
 def is_oauth_configured() -> bool:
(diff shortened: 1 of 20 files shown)


src/reachy_mini/apps/sources/hf_auth.py
```diff
@@ -1,164 +1,176 @@
...
-#   - http://reachy-mini.local:8000/api/hf-auth/oauth/callback  (wireless)
-#   - http://localhost:8000/api/hf-auth/oauth/callback          (lite)
-#
-# Then set HF_OAUTH_CLIENT_ID on all robots (same value for all).
-#
-# Environment variables:
-#   HF_OAUTH_CLIENT_ID     - Required for OAuth login
-#   HF_OAUTH_CLIENT_SECRET - Optional (for confidential clients)
-#
-# Pollen's HuggingFace OAuth app - works for all Reachy Mini robots
```
jUDASmILE, line 35: Hi, I would suggest better to keep the comments for better understanding what it does of following section. If any behavior changed then just sync the comments as well. thanks.

FabienDanieau (changes requested):
Tested on a Wireless (1.10.0 + live HF) and the design holds up: store lands at 0600 with a 30-day refresh token, sign-out leaves the user's `token`/`stored_tokens` byte-identical while the daemon refuses to read them (verified with a poisoned `HF_TOKEN` too), and the 401 refresh really does reconnect where `main` never does. Nice work, this file needed the rewrite.

Requesting changes for the rebase fallout, not the design.

**Rebase needed.** Branch is `CONFLICTING` with `main`; conflicts are test-only (`test_hf_auth.py`, `test_router_hf_auth.py`, `test_webrtc_utils.py`), all keep-both except `test_hf_auth.py` where `main`'s `token_path.read_text()` / `replacement_observation` asserts have to go, since that's exactly the behaviour this PR removes.

**28 tests fail once rebased** (baseline `main` has 1 pre-existing failure in `test_wireless.py`). Two causes, both mechanical:
- `test_trusted_endpoints.py:28` patches `huggingface_hub.get_token`, but `webrtc_utils._refresh_once` now reads `hf_auth.get_hf_token`. The patch lands nowhere, the daemon sees signed-out, TURN never fetches, and 24 TURN/relay tests time out. Point the fixture at `hf_auth.get_hf_token` and they pass.
- Your new `test_turn_credentials_refresh_once_populates_cache` stub `_Resp` has no `status_code`; `main`'s `_refresh_once` reads it since #1365. Add `status_code = 200`.

**`ruff format`** rewrites 4 spots in `hf_auth.py` (three collapsed `logger.debug(...)` calls and `_complete_device_login`'s signature). It's a pre-commit hook, so lint CI goes red.

**Worth a look, not blocking:** `get_hf_credential` holds `_store_lock` across the `refresh_access_token()` network call, and `get_hf_token()` is called synchronously from `async def` routes, including `/api/hf-auth/status` which the apps poll. Inside the 5-minute refresh margin a status poll stalls the event loop on an HTTPS round-trip and parks the relay and TURN threads on the same lock. `get_token()` used to be a file read. The generation counter you already have is the CAS you need: snapshot under the lock, release, refresh, re-acquire and only write if `lifecycle_generation` hasn't moved.

Minor: the immediate retry after a recovered 401 is wasted (`_token_updated` stays set into the next cycle, watcher tears it down at +20ms, reconnect lands after a full backoff ~5s). Bumping `_connection_attempts` and returning instead of `.set()` gets it to ~50ms. Measured on the robot; happy to share the harness.

PR #1368, opened 2026-08-18
feat(wireless): optional PRE argument for BLE update commands

tfrere:
## Issue

The BLE update proxy pins both `UPDATE_CHECK` and `UPDATE_START` to the stable channel (`pre_release=false`), so a client provisioning a robot over BLE can neither see nor install a release candidate. The WebRTC path already exposes this choice (`start_update` carries a `pre_release` flag); the BLE path has no equivalent, which blocks RC testing of the mobile app's over-BLE forced-update flow while `v1.10.0` only exists as an RC.

## Description

Both commands accept an optional `PRE` argument (`UPDATE_CHECK PRE`, `UPDATE_START PRE`) forwarded as `pre_release=true` to `/update/available` and `/update/start`. Bare commands keep the exact stable-channel behaviour, so existing clients are unaffected. Daemons that predate the argument ECHO the argumented form back (the standard unknown-command reply), which lets clients fall back to the bare command.

A separate `style(wireless)` commit runs `ruff-format` on `bluetooth_service.py` first: the file predates the formatting hook, which reflows it wholesale on any edit; formatting it in isolation keeps the functional diff readable.

## Testing

Two unit tests added to `test_ble_update_proxy.py` assert the `pre_release` param actually sent for the default and `PRE` variants of both commands (the suite is Linux-only, like the service).

Exercised on a Reachy Mini Wireless (daemon 1.9.0) by loading this branch's `bluetooth_service.py` on the robot and driving the real code paths against the live daemon HTTP API:

- `_handle_command(b"UPDATE_CHECK PRE")` -> `OK: working`, and the dispatched job hits `GET /update/available` with `pre_release=true`; the bare command keeps `pre_release=false`.
- Unauthenticated `UPDATE_START PRE` -> `ERROR: Not connected. Please authenticate first.` (no job scheduled).
- `UPDATE_START` was deliberately NOT triggered (it would actually install), and the GATT transport layer itself is untouched by this diff.

Note: on that 1.9.0 daemon the PRE check still reports `latest 1.9.0` because of the releases-dict key-order bug - end-to-end RC resolution needs #1351, as flagged below.

## Tested on

- [x] Reachy Mini Wireless
- [ ] Reachy Mini Lite

## AI assistance

- [ ] No AI involvement

### DIFF
--- a/src/reachy_mini/daemon/app/services/bluetooth/bluetooth_service.py
+++ b/src/reachy_mini/daemon/app/services/bluetooth/bluetooth_service.py
@@ -67,6 +67,7 @@ def get_hardware_id() -> str | None:
         return None
     return hashlib.sha256(raw.encode("ascii")).hexdigest()[:16]
 
+
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
@@ -352,7 +353,7 @@ class ResponseCharacteristic(Characteristic):
         self.notifying = False
         logger.info("Response notifications disabled")
         # Stop journal streaming if running (client disconnected without JOURNAL_STOP)
-        if hasattr(self.service, '_bt_service') and self.service._bt_service:
+        if hasattr(self.service, "_bt_service") and self.service._bt_service:
             self.service._bt_service._stop_journal()
 
     def send_notification(self, text: str):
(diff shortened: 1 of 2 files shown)


tfrere:
Note for review: the RC channel this PR exposes over BLE only resolves correctly once #1351 lands - today get_pypi_version(pre_release=True) returns a stale version because of the releases-dict key order. The PRE argument itself is independent, but end-to-end RC installs over BLE need both.

cdussieux (approved):
looks good and test and tested on wireless

PR #1379, opened 2026-08-26
fix(net): honour HTTP_PROXY/HTTPS_PROXY on our aiohttp call sites

### LINKED ISSUE #1377: Hugging Face requests do not consistently respect HTTP/HTTPS proxy configuration
## Ticket Type

🐛 Bug Report

## Robot Type

Wireless

## Operating System

Linux (WSL2 on Windows)

## Troubleshooting

* [x] I have read https://huggingface.co/docs/reachy_mini/troubleshooting

## Reachy Mini version

1.9.0

## Description

Reachy Mini's Hugging Face related network operations do not consistently respect an HTTP/HTTPS proxy configured on the host environment.

This makes it difficult or impossible to use Reachy Mini in environments where direct access to Hugging Face is unavailable but HTTPS access through a local HTTP proxy works correctly.

In my environment, the proxy itself is working: `curl` can successfully connect to Hugging Face when explicitly configured to use the proxy. However, Reachy Mini's Hugging Face-related requests do not consistently use the same proxy configuration and can fail to connect.

The issue appears to affect Hugging Face authentication and/or Hub-related requests performed by Reachy Mini rather than the underlying network connectivity itself.

I would like to improve Reachy Mini so that its Hugging Face network operations consistently respect standard HTTP/HTTPS proxy configuration, while preserving direct connections when no proxy is configured.

## Context & Reproduction

Environment:

* Windows 11 host
* WSL2 / Linux
* Python 3.12
* `reachy-mini==1.9.0`
* Hugging Face access through a local HTTP proxy
* Example proxy: `http://192.168.1.195:7897`

The proxy is known to work independently of Reachy Mini.

For example, from the WSL environment:

```bash
curl -4 -x http://192.168.1.195:7897 -I https://huggingface.co
```

successfully establishes an HTTPS connection through the proxy.

However, Hugging Face-related operations performed by Reachy Mini do not consistently inherit or use the proxy configuration.

A similar problem was observed during Hugging Face authentication/API access, where the same environment could access Hugging Face successfully with an explicitly configured proxy, while the Reachy Mini operation failed.

### Expected behavior

When `HTTP_PROXY` / `HTTPS_PROXY` (or equivalent standard proxy configuration) is configured, Reachy Mini's Hugging Face-related HTTP(S) requests should use that proxy.

When no proxy is configured, requests should continue to use the normal direct connection.

The proxy behavior should ideally be consistent across:

* Hugging Face OAuth/authentication requests
* Hugging Face Hub API requests
* Hugging Face Space/app metadata requests
* Hugging Face downloads or other Hub operations performed by Reachy Mini

### Actual behavior

The underlying proxy works correctly, but Reachy Mini's Hugging Face-related network operations can fail because the proxy configuration is not consistently respected.

This is particularly problematic in environments where direct access to `huggingface.co` is unavailable.

## Relevant logs or stack trace

The exact error depends on which Hugging Face operation is being performed.

The important diagnostic result is that direct access and proxy access behave differently:

```text
Direct:
https://huggingface.co
→ connection fails / times out

Through proxy:
curl -4 -x http://192.168.1.195:7897 -I https://huggingface.co
→ HTTP/2 200
```

This indicates that the proxy and the external network path are functional, while the failing Reachy Mini operation is not using the working proxy path consistently.

## Additional Info / Workarounds

A workaround is to configure a VPN or otherwise route Hugging Face traffic through a network path that does not require Reachy Mini to explicitly configure an HTTP proxy.

The current Reachy Mini troubleshooting documentation recommends VPN-based routing for environments where Hugging Face is inaccessible:

https://huggingface.co/docs/reachy_mini/troubleshooting

However, a standard HTTP/HTTPS proxy is often preferable in development environments because it can provide selective routing without changing the routing behavior of the entire host or robot.

### Proposed fix

I would like to investigate and implement a consistent proxy handling strategy for Reachy Mini's Hugging Face network operations.

The preferred approach would be to:

1. Identify all Hugging Face HTTP clients/request paths used by Reachy Mini.
2. Ensure they consistently respect standard `HTTP_PROXY` / `HTTPS_PROXY` environment variables where supported.
3. Avoid introducing a Reachy Mini-specific proxy implementation if the underlying Hugging Face client already provides standard proxy support.
4. Add regression tests covering requests with and without proxy configuration.
5. Document the supported proxy configuration and any limitations, particularly for authentication and Hub downloads.

The implementation should preserve the existing behavior for users who do not configure a proxy.

## AI assistance

AI helped write this report, a human verified it

pierre-rouanet:
Fixes #1377

## Why

`aiohttp.ClientSession` defaults to `trust_env=False`, which means it does **not** read `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` from the environment. `requests` — and therefore `huggingface_hub` — reads them by default.

We use both stacks, so Hugging Face traffic was split in two:

| Path | Client | Proxy honoured before this PR |
| --- | --- | --- |
| `snapshot_download`, `hf_hub_download`, `HfApi`, `whoami` | `huggingface_hub` → `requests` | ✅ |
| OAuth token exchange, app store listing, app update check | raw `aiohttp` | ❌ |
| `git clone`, `uv pip install git+https://…` | subprocess (inherits env) | ✅ |

That matches the report exactly: logging in and browsing the app store fail behind a proxy, while installing an app (which goes through `huggingface_hub`) works.

Nothing was ever using plain HTTP to reach Hugging Face — every HF URL in the package is `https://`. The failure was purely the ignored proxy configuration.

## What

`trust_env=True` on every `aiohttp.ClientSession` we open:

- `apps/sources/hf_auth.py` — OAuth token exchange (`https://huggingface.co/oauth/token`)
- `apps/sources/hf_space.py` — authorized app list + `/api/spaces`
- `apps/sources/app_update_checker.py` — `/api/spaces/{id}` update check
- `daemon/app/routers/hf_auth.py` — central robot-status proxy
- `media/central_signaling_relay.py`, `media/central_consumer.py` — central signaling (not HF, but the same blind spot)

Per the issue's point 3, no Reachy-Mini-specific proxy handling is introduced — this just stops opting out of aiohttp's own support. `NO_PROXY` comes along for free on the same code path.

## Tests

`tests/unit_tests/test_proxy_env_support.py`:

- End-to-end tests running a minimal recording HTTP server that acts as both an origin server (origin-form request line) and a forward proxy (absolute-form request line), so they assert on the request line actually received:
  - listing and update-check requests to an **unresolvable** host succeed via `HTTP_PROXY` — they cannot pass by accident
  - with no proxy configured, the request goes out directly (existing behaviour preserved)
  - a host in `NO_PROXY` bypasses the proxy
- An AST guard over `src/reachy_mini/` that fails if any `aiohttp.ClientSession(...)` is constructed without `trust_env=True`, so a new call site can't silently regress.

Verified all three proxy tests fail on `main` and pass with the fix. Full unit suite: 820 passed, 27 skipped (hardware-marked tests deselected).

`tests/unit_tests/test_hf_auth.py`'s stub session took no arguments; it now records its kwargs and asserts `trust_env=True` reached it.

## Docs

New troubleshooting entry documenting the supported configuration, including how to expose the variables to the daemon on Wireless (`systemctl edit reachy-mini-daemon`), and the limitations: authenticated proxies are passed through but not covered by CI, SOCKS is unsupported, and apps making their own network calls are responsible for their own proxy handling.

## Note for reviewers

`trust_env=True` also makes aiohttp read `~/.netrc` for HTTP auth. Our sessions send explicit `Authorization` headers where they need one (which takes precedence), so this shouldn't change behaviour — but it's the one non-proxy side effect of the flag and worth a second pair of eyes.

### DIFF
--- a/docs/source/troubleshooting.md
+++ b/docs/source/troubleshooting.md
@@ -414,7 +414,15 @@ Known limitations:
 - Only unauthenticated HTTP proxies are covered by our tests. A proxy needing
   credentials (`http://user:password@host:port`) is passed straight through to
   the underlying client and is not exercised in CI.
-- SOCKS proxies are not supported.
+- SOCKS proxies are not supported, and proxy URLs must be plain
+  `http://` (HTTPS is tunnelled through them); an `https://` proxy URL is
+  ignored with a warning.
+- `ALL_PROXY` is not honoured — set `HTTP_PROXY` and `HTTPS_PROXY`
+  explicitly. (`requests`-based calls would fall back to `ALL_PROXY`, the
+  daemon's aiohttp calls never have.)
+- `~/.netrc` is never consulted for these calls: the proxy is resolved
+  explicitly per request, so a developer netrc cannot break or alter the
+  daemon's authenticated Hugging Face / central requests.
 - Apps run in their own processes and inherit the daemon environment, but an
   app making its own network calls is responsible for honouring the proxy
   itself.
--- a/docs/source/troubleshooting.md
+++ b/docs/source/troubleshooting.md
@@ -374,6 +374,54 @@ Note that you may also need to use mirrors to reach services like PyPI and GitHu
 </details>
 
 <details>
+<summary><strong>How to use Reachy Mini behind an HTTP proxy?</strong></summary>
+
+Reachy Mini honours the standard proxy environment variables for all of its
+outgoing HTTP(S) traffic — Hugging Face authentication, the app store listing,
+app installs and updates, and the WebRTC signaling server:
+
+```bash
+export HTTP_PROXY=http://proxy.example.com:3128
+export HTTPS_PROXY=http://proxy.example.com:3128
+export NO_PROXY=localhost,127.0.0.1,reachy-mini.local
+```
+
+Keep local addresses in `NO_PROXY` so the daemon, mDNS discovery and the
+dashboard stay reachable without going through the proxy.
+
+For the Wireless version the variables must be visible to the daemon service,
+not just to your shell. Add them to the unit and restart it:
+
+```bash
+sudo systemctl edit reachy-mini-daemon
+```
+
+Add the following, then save:
+
+```
+[Service]
+Environment="HTTP_PROXY=http://proxy.example.com:3128"
+Environment="HTTPS_PROXY=http://proxy.example.com:3128"
+Environment="NO_PROXY=localhost,127.0.0.1,reachy-mini.local"
+```
+
+```bash
+sudo systemctl restart reachy-mini-daemon
+```
+
+Known limitations:
+
+- Only unauthenticated HTTP proxies are covered by our tests. A proxy needing
+  credentials (`http://user:password@host:port`) is passed straight through to
+  the underlying client and is not exercised in CI.
+- SOCKS proxies are not supported.
+- Apps run in their own processes and inherit the daemon environment, but an
+  app making its own network calls is responsible for honouring the proxy
+  itself.
+
+</details>
+
+<details>
 <summary><strong>How to make the conversation app work in China?</strong></summary>
 
 Reachy Mini conversation app relies on OpenAI gpt-realtime API, which might be inaccessible from China.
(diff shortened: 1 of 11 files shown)


cdussieux (approved):
added explicit resolve

PR #933, opened 2026-03-05
Deprecate web dashboard.

pierre-rouanet:
- [x] Remove web dashboard assets
- [x] Keep only a static page to redirect towards desktop app
- [x] Update the doc
   - [x] README.md
   - [x] docs/source/SDK/quickstart.md
   - [x] docs/source/platforms/simulation/get_started.md
   - [x] docs/source/platforms/reachy_mini/usage.md
   - [x] docs/source/platforms/reachy_mini/get_started.md
   - [x] docs/source/platforms/reachy_mini/development_workflow.md
   - [x] docs/source/platforms/reachy_mini/install_daemon_from_branch.md
   - [x] docs/source/troubleshooting.md
   - [x] docs/source/troubleshooting/motors_diagnosis.md
   - [x] docs/notebooks/README.md

cc @CarolinePascal (to double check what's should be updated in the doc)

PR #1391, opened 2026-08-31
Lift the starlette<1.0.0 ceiling

### LINKED ISSUE #1317: starlette<1.0.0 pin holds downstream apps on a version with five open advisories

BrinShadewater:
Follows up #1317.

The ceiling's recorded reason — `# TODO: update TemplateResponse(request, name, context) before allowing Starlette 1.x` — describes code that no longer exists: no `TemplateResponse`, `Jinja2Templates`, or `starlette.templating`/`fastapi.templating` import remains anywhere in the package (detail in #1317). Meanwhile the pin is transitive, so applications built on the SDK cannot resolve the five open starlette advisories (two high, including GHSA-wqp7-x3pw-xc5r — SSRF/NTLM credential theft via UNC paths in `StaticFiles` on Windows) on their own.

**Change:** floor at `starlette>=1.1.0` (the version fixing both high-severity advisories); the lock resolves starlette **1.3.1**, which carries all five fixes, and moves fastapi 0.129.0 → 0.141.1 for starlette-1.x compatibility.

**Verification** — clean `python:3.12` container, the CI marker set (`not audio and not video and not wireless and not webrtc`, `MUJOCO_GL=disable`), baseline first:

| Checkout | starlette | Result |
|---|---|---|
| `main`, frozen lock | 0.52.1 | 869 passed, 1 failed |
| this branch | 1.3.1 | 869 passed, **1 failed — the same test** |

The one failure (`test_startup_check_gpio_shutdown.py::test_target_same_is_untouched`) fails identically on an unmodified checkout of `main`, so it is environmental, not from this change. The audio/video and lint jobs were not run locally — CI here will cover them.

Closes #1317.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,7 @@ dependencies = [
     "rich",
     "rustypot>=1.4.2",
     "scipy>=1.15.3, <2.0.0",
-    "starlette<1.0.0", # TODO: update TemplateResponse(request, name, context) before allowing Starlette 1.x
+    "starlette>=1.1.0",
     "toml",
     "uvicorn[standard]",
     "websockets>=12,<16",
--- a/uv.lock
+++ b/uv.lock
@@ -3,7 +3,7 @@ revision = 3
 requires-python = ">=3.11"
 
 [options]
-exclude-newer = "2026-08-04T08:50:02.277943339Z"
+exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.
 exclude-newer-span = "P7D"
 
 [options.exclude-newer-package]
@@ -896,7 +896,7 @@ wheels = [
 
 [[package]]
 name = "fastapi"
-version = "0.129.0"
+version = "0.141.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "annotated-doc" },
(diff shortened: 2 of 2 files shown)


FabienDanieau (changes requested):
Superseded by #1326, which makes the same change with the correct floor and a clean lock diff. Closing this in favour of it.

For the record, the reasoning here was sound and I verified all three claims:

- The TODO is genuinely obsolete: the dashboard routes that used `Jinja2Templates` / `TemplateResponse` were deleted in #933 ("Deprecate web dashboard"), not migrated. The remaining jinja2 usage is `apps/templates/*.j2` rendered through `Environment`/`FileSystemLoader`, which starlette's signature change does not touch.
- starlette is a real direct dependency now, so keeping it declared is right: `daemon/app/middleware.py:7` imports `starlette.types` (added with #1209).
- The fastapi bump is required, not incidental: fastapi 0.129.0's metadata caps `starlette<1.0.0`.

The one substantive difference is the floor. The five advisories open against 0.52.1 do not all clear at 1.1.0:

| GHSA | sev | fixed in |
|---|---|---|
| GHSA-82w8-qh3p-5jfq, form limits ignored on urlencoded, DoS | **high** | **1.3.1** |
| GHSA-jp82-jpqv-5vv3, authority poisoning | low | 1.3.0 |
| GHSA-wqp7-x3pw-xc5r, StaticFiles UNC SSRF/NTLM on Windows | high | 1.1.0 |
| GHSA-x746-7m8f-x49c, HTTPEndpoint getattr dispatch | medium | 1.1.0 |
| GHSA-86qp-5c8j-p5mr, Host header path poisoning | medium | 1.0.1 |

`>=1.1.0` would let a downstream resolver land 1.1.x or 1.2.x and stay exposed to the second high, which is the situation the issue is about. #1326 declares `starlette>=1.3.1,<2.0.0`, and its lock diff moves only the starlette and fastapi entries, whereas this branch's lock also reverts `exclude-newer` to the `0001-01-01` placeholder and mangles the ipython markers (`colorama` to `python_version < '0'`, `pexpect` losing its `sys_platform != 'win32'` guard).

I have tested starlette 1.3.1 on a real robot and against the full unit suite on current main; results are in #1326. Thanks for the write-up and the verification table, they made this quick to confirm.

FabienDanieau:
Closing in favour of #1326, which lands the same change with the correct `starlette>=1.3.1,<2.0.0` floor (see the review above for why 1.1.0 is not enough) and a lock diff limited to starlette and fastapi. Tested 1.3.1 on a real robot and against the full unit suite there.

PR #1395, opened 2026-08-31
Cast JSON floats before packing int-typed XVF3800 registers

### LINKED ISSUE #1394: Audio config API cannot write int-typed XVF3800 parameters
## Problem

`POST /api/audio/config/apply` (and the JS SDK's `applyAudioConfig` over the WebRTC data channel — same code path) answers `{"applied": false}` for every int-typed XVF3800 parameter (int32/uint32/uint8), e.g. `PP_ECHOONOFF`. Only float-typed parameters work.

## Repro

```bash
curl -X POST http://reachy-mini.local:8000/api/audio/config/apply \
  -H 'Content-Type: application/json' \
  -d '{"config":[{"name":"PP_ECHOONOFF","values":[0]}],"verify":true}'
# -> {"applied": false}, register unchanged
```

## Cause

The request schema is `values: list[float]`, so pydantic coerces `0` to `0.0`, and `ReSpeaker.write` packs int registers with `struct.pack("i", value)` / `value.to_bytes(...)` — both raise on floats. The exception is swallowed per-parameter by `apply_audio_config`, which just reports failure.

## Solution

Cast to `int` in the integer packing branches, exactly like the float branch already casts to `float`.

FabienDanieau:
## Problem

`POST /api/audio/config/apply` — and the JS SDK's `applyAudioConfig` over the WebRTC data channel, which shares the code path — returns `{"applied": false}` for every int-typed XVF3800 parameter (e.g. `PP_ECHOONOFF`). Only float-typed parameters were writable. Fixes #1394.

## Repro

```bash
curl -X POST http://reachy-mini.local:8000/api/audio/config/apply \
  -H 'Content-Type: application/json' \
  -d '{"config":[{"name":"PP_ECHOONOFF","values":[0]}],"verify":true}'
# -> {"applied": false}, register unchanged
```

## Solution

The request schema coerces values to `float` (pydantic `list[float]`), and `ReSpeaker.write` packed them raw — `struct.pack("i", 0.0)` raises, `float.to_bytes` doesn't exist. Cast to `int` in the integer packing branches, mirroring the existing `float(...)` cast in the float branch. Verified on a real Wireless (the apply now returns `{"applied": true}` and the register reads back changed); regression test on the fake board writes floats into int32/uint8/uint32 registers.

### DIFF
--- a/src/reachy_mini/media/audio_control_utils.py
+++ b/src/reachy_mini/media/audio_control_utils.py
@@ -216,18 +216,22 @@ class ReSpeaker:
                 if isinstance(data_list, str)
                 else bytearray(data_list)
             )
+        # Integer types cast explicitly, like the float branch does: values
+        # arriving through JSON (REST apply endpoint, JS applyAudioConfig over
+        # the data channel) are floats after pydantic's list[float] coercion,
+        # and struct.pack("i", 0.0) / float.to_bytes raise on them.
         elif data_type == "uint8":
             for i in range(data_cnt):
-                payload += data_list[i].to_bytes(1, byteorder="little")
+                payload += int(data_list[i]).to_bytes(1, byteorder="little")
         elif data_type == "uint32" or data_type == "int32":
             for i in range(data_cnt):
                 payload += struct.pack(
-                    b"I" if data_type == "uint32" else b"i", data_list[i]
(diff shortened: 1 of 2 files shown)


cdussieux (approved):
lgtm

PR #1397, opened 2026-09-01
Add over-the-air BLE test suite runnable from an external host

cdussieux:
New tests/bluetooth_tests suite: read-only smoke tests for the BLE provisioning service, run from an external Linux host with a Bluetooth adapter (e.g. a Raspberry Pi) against a robot in radio range. The suite discovers the robot on its own (advertised name / status-service UUID), imports protocol constants from the firmware source so it fails loudly on protocol drift, and covers GATT layout, the Device Information service, status characteristics, PING/ECHO round-trips, the journal streaming lifecycle, the public WiFi provisioning surface (including the saved-networks-must-not-leak invariant), and the auth gates on privileged commands. No PIN is ever sent (the robot-global wrong-PIN lockout is never armed) and nothing on the robot is mutated.

Scanning goes through BlueZ (bleak), but the connection uses a minimal ATT client over a raw LE L2CAP socket (att_client.py): the robot's dual-mode advertisement makes BlueZ's Device1.Connect pick the classic BR/EDR bearer and fail with br-connection-profile-unavailable, while phones and Web Bluetooth are LE-only and never hit this. The raw socket behaves like them, works unprivileged, and tests the robot exactly as shipped.

Kept outside testpaths with a new 'ble' marker and a 'ble-test' dependency group (bleak); run_on_pi.sh rsyncs the repo to a Pi, bootstraps a minimal venv (handles hosts without python3-venv/pip), and runs the suite. Validated end-to-end: 15/15 passing from a Raspberry Pi against a live robot.

## Issue

Closes #

## Tested on

- [ ] Reachy Mini Wireless
- [ ] Reachy Mini Lite

## AI assistance

- [ ] No AI involvement

### DIFF
--- a/src/reachy_mini_ble_test/__init__.py
+++ b/src/reachy_mini_ble_test/__init__.py
@@ -1,0 +1,13 @@
+"""Over-the-air BLE smoke tests for the Reachy Mini provisioning service.
+
+Run from any external Linux host with BlueZ and a Bluetooth adapter in radio
+range of a powered robot — NOT on the robot itself:
+
+    pip install reachy-mini pytest        # or: pip install --no-deps reachy-mini pytest
+    pytest --pyargs reachy_mini_ble_test -v
+
+This package deliberately depends on nothing but the standard library (plus
+pytest to run), so a minimal ``--no-deps`` install is enough on a bare test
+host like a Raspberry Pi. See README.md in this directory for scope,
+configuration, and gotchas.
+"""
--- a/src/reachy_mini_ble_test/att_client.py
+++ b/src/reachy_mini_ble_test/att_client.py
@@ -1,0 +1,324 @@
+"""Minimal GATT client over a raw LE L2CAP socket (the ATT channel).
+
+Why not connect through BlueZ's D-Bus API? The robot's controller is
+dual-mode and its advertisement does not carry the "BR/EDR Not Supported"
+flag, so ``org.bluez.Device1.Connect`` picks the classic BR/EDR bearer —
+where the robot has nothing listening — and fails with
+``br-connection-profile-unavailable``. Phones and Web Bluetooth are LE-only
+by construction, which is why they never hit this. This client does the
+same thing they do: it opens an LE-bearer L2CAP connection on the ATT
+channel (CID 4) directly, exactly like the classic ``gatttool``. It works
+unprivileged, needs no host or robot configuration, and exercises the
+robot's GATT service exactly as shipped.
+
+Python 3.12's socket module cannot express the LE address type in a
+BTPROTO_L2CAP sockaddr (3.13 added it), so bind/connect go through libc
+with a hand-packed ``struct sockaddr_l2``.
+
+Implements the small ATT subset the tests need: MTU exchange, service and
+characteristic discovery, reads (with blob continuation), writes, CCCD
+subscription, and notifications.
+"""
+
+import ctypes
+import os
+import queue
+import socket
+import struct
+import time
+from uuid import UUID
+
+BDADDR_LE_PUBLIC = 1
+BDADDR_LE_RANDOM = 2
+ATT_CID = 4
+
+# Linux-only socket constants, via getattr so type checking stays
+# platform-independent (the suite itself only runs on Linux/BlueZ hosts).
+_AF_BLUETOOTH: int = getattr(socket, "AF_BLUETOOTH", 31)
+_BTPROTO_L2CAP: int = getattr(socket, "BTPROTO_L2CAP", 0)
+
+# ATT opcodes
+_ERROR_RSP = 0x01
+_EXCHANGE_MTU_REQ, _EXCHANGE_MTU_RSP = 0x02, 0x03
+_FIND_INFO_REQ, _FIND_INFO_RSP = 0x04, 0x05
+_READ_BY_TYPE_REQ, _READ_BY_TYPE_RSP = 0x08, 0x09
+_READ_REQ, _READ_RSP = 0x0A, 0x0B
+_READ_BLOB_REQ, _READ_BLOB_RSP = 0x0C, 0x0D
+_WRITE_REQ, _WRITE_RSP = 0x12, 0x13
+_NOTIFICATION = 0x1B
+
+_ATT_ERR_ATTRIBUTE_NOT_FOUND = 0x0A
+_ATT_ERR_ATTRIBUTE_NOT_LONG = 0x0B
+
+_CHARACTERISTIC = 0x2803
+CCCD_UUID16 = 0x2902
+
+_PROPERTY_BITS = {
+    0x01: "broadcast",
+    0x02: "read",
+    0x04: "write-without-response",
+    0x08: "write",
+    0x10: "notify",
+    0x20: "indicate",
+}
+
+
+def _uuid128_to_str(le_bytes: bytes) -> str:
+    return str(UUID(bytes=bytes(reversed(le_bytes))))
+
+
+def _uuid16_to_str(value: int) -> str:
+    return f"{value:08x}-0000-1000-8000-00805f9b34fb"
+
+
+def _uuid_bytes_to_str(uuid_bytes: bytes) -> str:
+    if len(uuid_bytes) == 2:
+        return _uuid16_to_str(struct.unpack("<H", uuid_bytes)[0])
+    return _uuid128_to_str(uuid_bytes)
+
+
+def _bdaddr(mac: str) -> bytes:
+    return bytes(reversed(bytes.fromhex(mac.replace(":", ""))))
+
+
+def _sockaddr_l2(mac_le: bytes, addr_type: int) -> bytes:
+    # struct sockaddr_l2: family, psm, bdaddr[6], cid, bdaddr_type (+pad)
+    return struct.pack("<HH6sHBx", _AF_BLUETOOTH, 0, mac_le, ATT_CID, addr_type)
+
+
+class AttError(Exception):
+    """ATT Error Response from the server."""
+
+    def __init__(self, request_opcode: int, handle: int, code: int) -> None:
+        """Record the failed request opcode, target handle, and error code."""
+        self.request_opcode = request_opcode
+        self.handle = handle
+        self.code = code
+        super().__init__(
+            f"ATT error 0x{code:02x} for request 0x{request_opcode:02x} "
+            f"on handle 0x{handle:04x}"
+        )
+
+
+class Characteristic:
+    """A discovered characteristic: handles, UUID, decoded properties."""
+
+    def __init__(
+        self, decl_handle: int, properties_mask: int, value_handle: int, uuid: str
+    ) -> None:
+        """Store the discovery result for one characteristic declaration."""
+        self.decl_handle = decl_handle
+        self.value_handle = value_handle
+        self.uuid = uuid
+        self.properties = [
+            name for bit, name in _PROPERTY_BITS.items() if properties_mask & bit
+        ]
+
+
+class AttClient:
+    """Synchronous GATT client over a raw LE L2CAP (ATT) socket."""
+
+    def __init__(
+        self, address: str, address_type: str = "public", timeout: float = 10.0
+    ) -> None:
+        """Prepare a client for the peer at ``address`` (no I/O yet)."""
+        self.address = address
+        self.address_type = address_type
+        self.timeout = timeout
+        self.mtu = 23
+        self.characteristics: list[Characteristic] = []  # in handle order
+        self.notifications: queue.Queue[tuple[int, bytes]] = queue.Queue()
+        self._sock: socket.socket | None = None
+
+    # -- connection ----------------------------------------------------
+
+    def connect(self) -> None:
+        """Open the LE connection and discover the GATT database."""
+        peer_type = (
+            BDADDR_LE_RANDOM if self.address_type == "random" else BDADDR_LE_PUBLIC
+        )
+        libc = ctypes.CDLL(None, use_errno=True)
+        sock = socket.socket(_AF_BLUETOOTH, socket.SOCK_SEQPACKET, _BTPROTO_L2CAP)
+        try:
+            local = _sockaddr_l2(b"\x00" * 6, BDADDR_LE_PUBLIC)
+            if libc.bind(sock.fileno(), local, len(local)) != 0:
+                raise OSError(ctypes.get_errno(), os.strerror(ctypes.get_errno()))
+            peer = _sockaddr_l2(_bdaddr(self.address), peer_type)
+            if libc.connect(sock.fileno(), peer, len(peer)) != 0:
+                raise OSError(ctypes.get_errno(), os.strerror(ctypes.get_errno()))
+        except BaseException:
+            sock.close()
+            raise
+        sock.settimeout(self.timeout)
+        self._sock = sock
+        self._exchange_mtu()
+        self._discover()
+
+    def close(self) -> None:
+        """Close the LE connection."""
+        if self._sock is not None:
+            self._sock.close()
+            self._sock = None
+
+    # -- ATT plumbing --------------------------------------------------
+
+    def _socket(self) -> socket.socket:
+        if self._sock is None:
+            raise ConnectionError("not connected")
+        return self._sock
+
+    def _recv(self, deadline: float) -> bytes:
+        sock = self._socket()
+        remaining = deadline - time.monotonic()
+        if remaining <= 0:
+            raise TimeoutError("timed out waiting for ATT PDU")
+        sock.settimeout(remaining)
+        pdu = sock.recv(1024)
+        if not pdu:
+            raise ConnectionError("LE connection closed by peer")
+        return pdu
+
+    def _request(self, pdu: bytes, expected_opcode: int) -> bytes:
+        """Send one ATT request and return the matching response PDU.
+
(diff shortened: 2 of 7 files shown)


src/reachy_mini_ble_test/att_client.py
```diff
@@ -0,0 +1,324 @@
...
+        )
+        libc = ctypes.CDLL(None, use_errno=True)
+        sock = socket.socket(_AF_BLUETOOTH, socket.SOCK_SEQPACKET, _BTPROTO_L2CAP)
+        try:
+            local = _sockaddr_l2(b"\x00" * 6, BDADDR_LE_PUBLIC)
+            if libc.bind(sock.fileno(), local, len(local)) != 0:
+                raise OSError(ctypes.get_errno(), os.strerror(ctypes.get_errno()))
+            peer = _sockaddr_l2(_bdaddr(self.address), peer_type)
+            if libc.connect(sock.fileno(), peer, len(peer)) != 0:
+                raise OSError(ctypes.get_errno(), os.strerror(ctypes.get_errno()))
```
FabienDanieau, line 148: Any non-zero return is treated as fatal, but on a clean Pi 4 (BlueZ 5.82) `connect()` returns `EINPROGRESS` (115), which means "in progress", not failure. Standard handling: on `EINPROGRESS`, `select()` for writability, then read `SO_ERROR` and raise `OSError(so_err, ...)` if non-zero. As written, a perfectly valid connection is thrown away, and the caller's retry hides the real error (see `ble_link.py:163`).

src/reachy_mini_ble_test/att_client.py
```diff
@@ -0,0 +1,324 @@
...
+                handle = int(struct.unpack_from("<H", rsp, 1)[0])
+                self.notifications.put((handle, rsp[3:]))
+            elif opcode == expected_opcode:
+                return rsp
+            elif opcode == _ERROR_RSP and rsp[1] == pdu[0]:
+                _, req_op, handle, code = struct.unpack("<BBHB", rsp)
+                raise AttError(req_op, handle, code)
+            # anything else (e.g. server-initiated exchange): ignore
+
+    def wait_notification(self, timeout: float) -> bytes:
```
FabienDanieau, line 201: `wait_notification` ignores the source handle. Fine today because only the response characteristic notifies, but silently wrong the day network status gets `notify`. One-line check against `self._response_char.value_handle` (or take the expected handle as a parameter).

src/reachy_mini_ble_test/att_client.py
```diff
@@ -0,0 +1,324 @@
...
+            entry_len, offset = rsp[1], 2
+            while offset + entry_len <= len(rsp):
+                decl_handle = int(struct.unpack_from("<H", rsp, offset)[0])
+                props, value_handle = struct.unpack_from("<BH", rsp, offset + 2)
+                uuid = _uuid_bytes_to_str(rsp[offset + 5 : offset + entry_len])
+                self.characteristics.append(
+                    Characteristic(decl_handle, int(props), int(value_handle), uuid)
+                )
+                offset += entry_len
+            start = decl_handle + 1
```
FabienDanieau, line 257: If a Read By Type response yields no parsable entry, `decl_handle` is unchanged and `start` advances by one handle, so this loops up to ~65k round trips before giving up. Break when the inner loop appended nothing.

src/reachy_mini_ble_test/att_client.py
```diff
@@ -0,0 +1,324 @@
...
+        rsp = self._request(struct.pack("<BH", _READ_REQ, handle), _READ_RSP)
+        value = rsp[1:]
+        while len(value) % (self.mtu - 1) == 0 and value:
+            try:
+                rsp = self._request(
+                    struct.pack("<BHH", _READ_BLOB_REQ, handle, len(value)),
+                    _READ_BLOB_RSP,
+                )
+            except AttError as e:
+                if e.code == _ATT_ERR_ATTRIBUTE_NOT_LONG:
```
FabienDanieau, line 307: BlueZ answers a Read Blob at `offset == length` with `Invalid Offset (0x07)`, not only `Attribute Not Long (0x0B)`. A value whose length is an exact multiple of `mtu - 1` raises here instead of returning. Accept both codes.

src/reachy_mini_ble_test/ble_link.py
```diff
@@ -0,0 +1,227 @@
...
+        self.advertised_uuids: set[str] = set()
+        self._att: AttClient | None = None
+        self._command_char: Characteristic | None = None
+        self._response_char: Characteristic | None = None
+
+    # -- discovery / connection ----------------------------------------
+
+    @staticmethod
+    def _bluetoothctl(*args: str, timeout: float) -> str:
+        return subprocess.run(
```
FabienDanieau, line 115: Return code and stderr are ignored. A missing `bluetoothctl` surfaces as a raw `FileNotFoundError`, a scan overrun as `TimeoutExpired`, both out of a session fixture with no hint. Wrap them in `BleLinkError` with a message.

src/reachy_mini_ble_test/ble_link.py
```diff
@@ -0,0 +1,227 @@
...
+        """
+        self._bluetoothctl(
+            "--timeout",
+            str(int(self.scan_timeout)),
+            "scan",
+            "on",
+            timeout=self.scan_timeout + 10,
+        )
+        wanted = _normalize_name(self.name)
+        for line in self._bluetoothctl("devices", timeout=10).splitlines():
```
FabienDanieau, line 134: `bluetoothctl devices` lists BlueZ's *cache*, not what the scan just saw. On my laptop it returned two headsets that emitted nothing during a 15 s scan. A robot powered off an hour ago still "discovers", then `connect()` fails with a bare `OSError` instead of the fixture's helpful message. The scan call above already captures stdout and discards it; parse the `[NEW] Device` lines from it instead.

src/reachy_mini_ble_test/ble_link.py
```diff
@@ -0,0 +1,227 @@
...
+            match = re.fullmatch(r"Device ((?:[0-9A-F]{2}:){5}[0-9A-F]{2}) (.*)", line)
+            if not match:
+                continue
+            mac, name = match.groups()
+            if self.address and mac.lower() != self.address.lower():
+                continue
+            if self.address or _normalize_name(name) == wanted:
+                self.device_address, self.device_name = mac, name
+                info = self._bluetoothctl("info", mac, timeout=10)
+                self.advertised_uuids = set(
```
FabienDanieau, line 144: This is not the advertisement. `bluetoothctl info` reports BlueZ's merged SDP + GATT cache. From a host that has ever been bonded to the robot, this set contains A2DP Audio Source, Handsfree AG and SIM Access (impossible in a 31-byte LE advert) and `...def0`, which the firmware never advertises (`bluetooth_service.py:1064` only puts `REACHY_STATUS_SERVICE_UUID` in). So `test_advertises_status_service_uuid` passes from a stale bond even if advertising is completely broken. Parse the UUID lines from the scan output instead.

src/reachy_mini_ble_test/ble_link.py
```diff
@@ -0,0 +1,227 @@
...
+        a random address type in case a robot ever uses one.
+        """
+        if self.device_address is None:
+            raise BleLinkError("discover() must find the robot before connect()")
+        try:
+            self._att = AttClient(
+                self.device_address, address_type="public", timeout=self.scan_timeout
+            )
+            self._att.connect()
+        except OSError:
```
FabienDanieau, line 163: Catching every `OSError` and retrying with `address_type="random"` means the exception the user finally sees is from the retry, not the real failure (on the Pi: a bare `BlockingIOError` hiding an `ECONNREFUSED`). It also doubles the wait when the robot is simply busy with another central, which the README lists as the first gotcha. Narrow this to the case where a random address can help, and let the original error propagate otherwise.

src/reachy_mini_ble_test/test_ble_smoke.py
```diff
@@ -0,0 +1,172 @@
...
+    SYSTEM_STATUS_UUID,
+    ReachyBleLink,
+)
+
+AUTH_ERROR = "ERROR: Not connected. Please authenticate first."
+
+
+def _skip_if_unsupported(response: str, command: str) -> None:
+    if response.startswith("ECHO:"):
+        pytest.skip(f"robot firmware predates {command} (command echoed back)")
```
FabienDanieau, line 42: This is the mechanism that turned a live regression green (see review body, point 1): the robot lost `WIFI_STATUS`, `WIFI_KEYEX` and `WIFI_SCAN` and the suite exited 0. Please invert it: fail by default, skip only when `REACHY_BLE_ALLOW_OLD_FIRMWARE=1` is set.

src/reachy_mini_ble_test/test_ble_smoke.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+# -- discovery / GATT layout -------------------------------------------
+
+
+def test_robot_is_discoverable(link: ReachyBleLink) -> None:
```
FabienDanieau, line 48: Cannot fail: the `link` fixture already `pytest.fail`s when `discover()` returns `None`, and the name check is what discovery matched on. Either drop it or make it assert something discovery did not already guarantee (e.g. the robot was seen in *this* scan, once discovery parses scan output).

src/reachy_mini_ble_test/test_ble_smoke.py
```diff
@@ -0,0 +1,172 @@
...
+    assert link.read_char(MODEL_NUMBER_UUID) == "Reachy Mini"
+    assert link.read_char(FIRMWARE_REVISION_UUID).startswith("[HOTSPOT]:")
+
+
+# -- status characteristics --------------------------------------------
+
+
+def test_system_status_is_online(link: ReachyBleLink) -> None:
+    """System status characteristic reads as Online."""
+    assert link.read_char(SYSTEM_STATUS_UUID) == "Online"
```
FabienDanieau, line 80: Vacuous: `bluetooth_service.py:560` is `StaticCharacteristic(..., "Online", ...)`, a hardcoded literal. It cannot read anything else while the service is alive, and the read path is already covered by `test_device_information_service`. Suggest dropping it.

src/reachy_mini_ble_test/test_ble_smoke.py
```diff
@@ -0,0 +1,172 @@
...
+def test_wifi_status_public_omits_known_networks(link: ReachyBleLink) -> None:
+    """Unauthenticated WIFI_STATUS must NOT leak the saved-network list."""
+    response = link.command("WIFI_STATUS")
+    _skip_if_unsupported(response, "WIFI_STATUS")
+    status = json.loads(response)
+    assert {"mode", "connected"} <= set(status), status
+    assert "known" not in status, "saved networks leaked without authentication"
+
+
+def test_wifi_keyex_returns_fresh_x25519_key(link: ReachyBleLink) -> None:
```
FabienDanieau, line 150: Docstring says "fresh" but the test never calls `WIFI_KEYEX` twice or compares `kid` / `pk`. Either compare two calls or rename to `..._returns_valid_x25519_key`.

src/reachy_mini_ble_test/test_ble_smoke.py
```diff
@@ -0,0 +1,172 @@
...
+def test_wifi_scan_requires_auth(link: ReachyBleLink) -> None:
+    """WIFI_SCAN is refused without a PIN session."""
+    response = link.command("WIFI_SCAN")
+    _skip_if_unsupported(response, "WIFI_SCAN")
+    assert response == AUTH_ERROR
+
+
+def test_cmd_scripts_require_auth(link: ReachyBleLink) -> None:
+    """CMD_* scripts are refused without a session (nothing is executed)."""
+    assert link.command("CMD_RESTART_DAEMON") == AUTH_ERROR
```
FabienDanieau, line 172: Only auth test without `_skip_if_unsupported`. Intentional? (Fine if so, `CMD_*` predates everything else; just checking it isn't an oversight, especially once the guard flips to fail-by-default.)

src/reachy_mini_ble_test/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+
+Run from any external Linux host with BlueZ and a Bluetooth adapter in radio
+range of a powered robot — NOT on the robot itself:
+
+    pip install reachy-mini pytest        # or: pip install --no-deps reachy-mini pytest
+    pytest --pyargs reachy_mini_ble_test -v
+
+This package deliberately depends on nothing but the standard library (plus
+pytest to run), so a minimal ``--no-deps`` install is enough on a bare test
+host like a Raspberry Pi. See README.md in this directory for scope,
```
FabienDanieau, line 11: `README.md` is not in the wheel (`package-data` only covers `reachy_mini`, `MANIFEST.in` only assets/templates/descriptions), so for anyone who `pip install`s this, the pointer dangles. Moot if the suite moves to `tests/hardware/` as suggested in the review body; otherwise add it to `[tool.setuptools.package-data]`.

FabienDanieau (changes requested):
Thanks for this, the raw-ATT approach is the right call and the read-only/no-PIN discipline is exactly what a smoke suite against a shared robot needs. I ran it from two hosts against a live robot before reviewing; results below. Requesting changes for three things the runs surfaced, plus a layout ask. Line-level points are inline.

## 1. Skips hide real drift (blocking)

Two consecutive runs from the same laptop, same code, same robot, minutes apart:

```
run 1:  15 passed
run 2:  11 passed, 4 skipped
        SKIPPED robot firmware predates the Hardware ID characteristic
        SKIPPED robot firmware predates WIFI_STATUS (command echoed back)
        SKIPPED robot firmware predates WIFI_KEYEX (command echoed back)
        SKIPPED robot firmware predates WIFI_SCAN (command echoed back)
```

The robot's BLE service had regressed to an older build underneath us (its daemon was down, so `check_and_update_bluetooth_service` couldn't resync). The suite lost a characteristic and the whole WiFi surface, including `test_wifi_status_public_omits_known_networks`, and still exited 0. For a suite whose purpose is catching protocol drift, `_skip_if_unsupported` inverts the signal. Please make it fail by default and gate the skip behind an explicit opt-in (`REACHY_BLE_ALLOW_OLD_FIRMWARE=1`).

Related: three tests can't fail as written (`test_system_status_is_online`, `test_robot_is_discoverable`, `test_advertises_status_service_uuid`), see inline. The last one matters most because it measures BlueZ's cache rather than the advertisement.

## 2. Clean Raspberry Pi: 15 errors (blocking until re-validated)

Pi 4, Debian trixie, BlueZ 5.82, Python 3.13, never paired or bonded with the robot (verified empty cache). Discovery works; the LE connect is refused:

```
connect rc=-1 errno=115 (EINPROGRESS) -> poll -> SO_ERROR=111 (Connection refused)
15 errors in 98s
```

Ruled out: competing central (other host's BT off), active discovery, adapter state (power-cycled). A bonded laptop connects to the same robot fine. The remaining systematic difference is the bond, but the robot was running the old service build at the time (see 1), so I can't separate "bonding is an undocumented prerequisite" from "old firmware required pairing". Could you re-run on a never-bonded host against a robot on this branch's `bluetooth_service.py` and paste the output? The "15/15 from a Pi" in the description reads as the earlier bleak / `tests/bluetooth_tests` / `run_on_pi.sh` version, none of which is in this diff, so I'm treating it as unverified for the current code.

## 3. `att_client.connect()` mishandles EINPROGRESS (blocking)

Inline at `att_client.py:148` and `ble_link.py:163`. Short version: `EINPROGRESS` is treated as fatal, then the random-address retry reports *its* error instead of the real one. That is why the Pi failure surfaced as a bare `BlockingIOError`.

## 4. Layout: put it with the other hardware tests

Please move this under `tests/hardware/ble/` following #1393 (`tests/hardware/`, per-test `@pytest.mark.ble`, `-m ble` to select, `not ble` added to the marker exclusions in `pytest.yml` / `allure.yml`), rather than shipping it as a top-level package in the wheel. Reasons:

- Every `pip install reachy-mini` currently gets `reachy_mini_ble_test/` (verified in the built wheel), and `README.md` is not in it while `__init__.py` points users at it.
- We want to run this from the physical CI runner and publish results. Being pytest already gets us `--alluredir` for free; what's missing is an artifact hook like #1393's `REACHY_TEST_ARTIFACTS=<dir>`: dump what the robot reported (GATT layout, DIS values, network status, available commands, hardware id, per-command round-trip times) to a JSON so the report has numbers, not just pass/fail.
- The "install on a bare Pi" story becomes `git clone` + `uv sync` + `uv run pytest tests/hardware/ble -m ble`, same runbook as #1393. `find_spec("reachy_mini")` keeps working for locating the firmware source without importing it.

Prerequisite worth stating in the README: the CI runner needs a BT adapter in radio range of `reachy-mini-ci`.

## Smaller

- The "why raw ATT" paragraph is in three places (README, `att_client.py`, `ble_link.py`); keep it in `att_client.py` and link from the other two.
- The PR description describes the previous implementation (`tests/bluetooth_tests`, `run_on_pi.sh`, bleak, a `ble-test` dependency group) and the template is unfilled.

## Separately

The reason this client has to exist is that the robot's advert lacks the BR/EDR-Not-Supported flag, so every BlueZ central (any Linux desktop) picks the classic bearer and fails. That's a robot-side interop bug affecting users, not just tests. Is there an issue for it? If it gets fixed, `att_client.py` collapses to a few lines of bleak, so a pointer to that issue in the module docstring would be good.

PR #1408, opened 2026-09-10
media(turn): point the default TURN credentials URL at the live proxy

### LINKED ISSUE #1356: Default TURN credentials endpoint (turn.fastrtc.org) is dead upstream
## Problem

Since #1182 the daemon fetches TURN credentials from `https://turn.fastrtc.org/credentials` by default. That domain has been dead since at least June 2026: the `.org` registry still delegates `fastrtc.org` to four Route53 nameservers, but the hosted zone behind them is gone, so they answer `REFUSED` and every public resolver returns `SERVFAIL` (gradio-app/fastrtc#429, unanswered; the fastrtc repo is dormant). The endpoint was already dead when #1182 merged.

Consequences:
- No relay candidates are ever offered, so remote WebRTC has no fallback behind symmetric NAT / CGNAT / UDP-blocking firewalls. Direct and STUN paths still work, which is why demos pass.
- Every logged-in robot logs `Failed to fetch TURN credentials` every 30 s forever.
- Security: the dangling Route53 delegation is a takeover vector (cycle hosted-zone creation until one of the delegated nameservers is assigned). Whoever takes the domain can obtain a valid cert via DNS validation and receive the HF bearer token every robot sends with the credentials request.

## Repro

`dig turn.fastrtc.org @1.1.1.1` → SERVFAIL; `dig SOA fastrtc.org @ns-79.awsdns-09.com +norecurse` → REFUSED.

## Solution

Drop the dead default so the relay is opt-in via `REACHY_TURN_URL` (branch `no-default-turn-url`). Follow-ups, separate issues if pursued: a Pollen-controlled credentials proxy to restore relay coverage, and a relay-only connectivity check so a dead endpoint can't merge silently again.

FabienDanieau:
## Problem

`turn.fastrtc.org`, the default `REACHY_TURN_URL`, has had dead DNS since June 2026. The `.org` registry still delegates the zone to four Route53 nameservers, but the hosted zone behind them is gone, so they answer `REFUSED` and resolvers return SERVFAIL (gradio-app/fastrtc#429, unanswered).

Every logged-in robot logs `Failed to fetch TURN credentials` every 30 s and offers host/srflx candidates only, so remote WebRTC has no fallback behind symmetric NAT, CGNAT or a UDP-blocking firewall. Direct and STUN paths still work, which is why demos passed and nobody noticed.

## Repro

```
$ dig +short turn.fastrtc.org        # SERVFAIL
$ journalctl -u reachy-mini-daemon | grep TURN
... WARNING - Failed to fetch TURN credentials: ... Failed to resolve 'turn.fastrtc.org'
```

## Solution

The service behind that alias, the `fastrtc/turn-service` Space, is still up. Address it directly.

Note this resolves #1356 differently from the issue's own proposal (drop the default, make the relay opt-in). A working default gives every robot relay coverage with no configuration, which seems worth more than opting out of an endpoint that now works. Happy to switch to the opt-in version if you disagree.

Verified on a Wireless: the daemon logs `Refreshed 5 TURN server(s)` on a 5 min cadence, and a session to an HF Space backend selected the relay pair and pushed ~2.2 Mbps of video through `turn.cloudflare.com`.

No new test: a default the validator rejects already fails four tests that construct `TurnCredentials()` with no url (checked with `REACHY_TURN_URL=http://evil.example`). A well-formed but dead host is what bit us, and catching that needs live DNS, which belongs in the physical CI workflow.

Fixes #1356

### DIFF
--- a/docs/source/SDK/cloud-backend-consumer.md
+++ b/docs/source/SDK/cloud-backend-consumer.md
@@ -122,14 +122,16 @@ candidate of its own (see `_apply_turn_servers` in
 If you need the consumer to allocate its own relay (e.g. when neither
 side can hole-punch), pass an `ice_servers_provider` returning
 `aiortc.RTCIceServer` objects with TURN credentials. The HF-hosted
-Cloudflare TURN proxy at `https://turn.fastrtc.org/credentials` is a
-reasonable default for HF Spaces:
+Cloudflare TURN proxy at `https://fastrtc-turn-service.hf.space/credentials`
+is a reasonable default for HF Spaces. Use a token the visitor owns where
+you can: the free allowance is metered per Hugging Face account, so a
+shared Space secret spends one quota for every visitor.
 
 ```python
 async def _ice():
     async with aiohttp.ClientSession() as s:
         r = await s.get(
-            "https://turn.fastrtc.org/credentials",
+            "https://fastrtc-turn-service.hf.space/credentials",
             headers={"Authorization": f"Bearer {space_secret_hf_token}"},
             params={"ttl": 600},
         )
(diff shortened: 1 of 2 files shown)


pierre-rouanet (approved):
LGTM

PR #1422, opened 2026-09-16
fix(daemon): keep the JSON-RPC relay off a throwaway event loop

### LINKED ISSUE #1421: JSON-RPC relay dies with 'Event loop is closed' after the daemon is started over the HTTP API
## Problem

After the daemon is started or restarted through the HTTP API, every JSON-RPC frame relayed from the WebRTC DataChannel is dropped. The client only sees its `rpcCall` time out and reads the robot as unreachable. The journal shows:

```
RuntimeError: Event loop is closed
```

It stays broken until the daemon process itself restarts.

## Reproduce

1. Run the daemon without `--autostart`, then start it over the API (the desktop app, the mobile app and the startup-app flow all do this).
2. Connect a client over WebRTC and call `apps.status` through the JSON-RPC relay.
3. The call times out; the daemon logs the error above on each frame.

Booting with `--autostart` is fine.

## Cause

`_setup_jsonrpc_relay` captures the loop it is wired on with `asyncio.get_running_loop()`. The API path runs `start()` through `bg_job_register.run_command`, which is `asyncio.run` in a thread, so that loop is closed as soon as the job finishes, and `run_coroutine_threadsafe` then fails on every frame.

Reproduced on a Wireless robot on 1.10.0 while driving the conversation app from the mobile app.

FabienDanieau:
Fixes #1421

## Problem

After the daemon is started or restarted over the HTTP API, every JSON-RPC frame relayed from the DataChannel is dropped with `RuntimeError: Event loop is closed`. Clients only see their `rpcCall` time out and read the robot as unreachable, until the daemon process restarts. Booting with `--autostart` is fine.

## Reproduce

Run the daemon without `--autostart`, start it over the API (what the desktop app, the mobile app and the startup-app flow do), connect over WebRTC and call `apps.status` through the relay. It times out; the journal shows the error on each frame.

## Solution

`_setup_jsonrpc_relay` captured the loop it was wired on, which on the API path is `asyncio.run` in a thread and closes when the job ends. The FastAPI lifespan now hands the daemon its long-lived loop and the relay uses that. If the loop is closed anyway (process shutdown), the handler answers a JSON-RPC error carrying `relay_unavailable` instead of dropping the frame silently.

Two unit tests pin the relayed call on the lifespan loop and the error on a closed one. Reproduced and verified on a Wireless robot.


FabienDanieau:
Needed by pollen-robotics/reachy_mini_mobile_app#96, which drives the on-robot conversation app through this relay.

PR #1423, opened 2026-09-16
fix(security): block DNS rebinding and cross-site writes on the daemo…

cdussieux:
…n API

The unauthenticated HTTP API relies on network locality, but two browser vectors let an internet page reach it anyway (CAN-2026-2032024):

- DNS rebinding: the attacker's domain re-resolves to the robot's LAN IP, making the page same-origin with the API.
- Preflight-free CSRF: CORS only stops the page from reading responses; a "simple" cross-origin POST (e.g. text/plain) still executes server-side, and FastAPI parses the body as JSON regardless of Content-Type. This still allowed e.g. POST /api/hf-auth/save-token to replace the stored HF token from a drive-by page.

Add LocalNetworkGuardMiddleware: the Host header must be an IP literal, localhost, or a .local mDNS name (all requests), and state-changing requests carrying an Origin must come from the robot itself, a local/private-network origin, or a native webview scheme (Tauri/Capacitor). Requests without an Origin header (curl, the Python SDK, native apps) are unaffected.

Direct LAN clients remain trusted; closing that requires a pairing secret and is tracked separately.

## Issue

partially closes VULN04

## Tested on

- [ ] Reachy Mini Wireless
- [ ] Reachy Mini Lite

## AI assistance

- [ ] No AI involvement


FabienDanieau:
Approach is right and the HTTP half is solid. One gap blocks it.

## WebSockets are entirely unguarded

`__call__` returns early on non-`http` scopes, so the nine WS endpoints — including `/api/move/ws/set_target`, `/api/move/ws/raw/write`, `/ws/sdk` and `/api/logs/ws/daemon` — get neither check.

This matters more than the POST vector the PR closes: browsers apply no CORS to WebSockets, so a drive-by `http://` page gets the write *and* the read that CORS was blocking for `fetch`. It also means the commit message's "the `Host` header must be an IP literal, `localhost`, or a `.local` mDNS name (all requests)" isn't accurate as written.

Reproduced on this branch — both of these pass today:

```python
app.add_middleware(LocalNetworkGuardMiddleware)

# Host: reachy-mini.local:8000, Origin: http://evil.example.com
with client.websocket_connect("/ws/set_target", headers=...) as w:
    assert w.receive_text() == "moved"

# same with Host: rebind.evil.example.com:8000 -> also accepted
```

Fix, verified locally (22 passed including your 20; Origin-less SDK clients and `tauri://localhost` still connect):

```diff
-        if scope["type"] != "http":
+        if scope["type"] not in ("http", "websocket"):

-        method = str(scope.get("method", "GET")).upper()
+        # A WebSocket handshake is never "safe": it is readable cross-origin.
+        method = ("POST" if scope["type"] == "websocket"
+                  else str(scope.get("method", "GET")).upper())

+    async def _deny(self, scope, send, status, detail):
+        if scope["type"] == "websocket":
+            await send({"type": "websocket.close", "code": 1008, "reason": detail})
+            return
+        await self._reject(send, status, detail)
```

Gotcha when you write the tests: `TestClient` hardcodes `host: testserver` on WS handshakes regardless of `base_url`, so every WS case needs an explicit `host` header.

## Two smaller things on the trust logic

- `is_trusted_origin`'s `host == request_host` branch lets a public-IP `Origin` through when the robot is reached at that same public IP. Dropping the branch changes 2 verdicts out of a 114-case Host×Origin matrix, both of exactly that shape — every legitimate host already matches the loopback / private / `.local` branches.
- `_MDNS_HOSTNAME_RE` rejects the trailing-dot FQDN form `reachy-mini.local.`, which browsers accept. `.local` is never delegated in the public DNS root (RFC 6762), so `host.lower().rstrip(".").endswith(".local")` is equivalent for the threat model and fixes that case. Note the `.lower()` — the current regex is `IGNORECASE`.

Also worth a line in the PR body as a known behaviour change: bare LAN hostnames (`http://reachy-mini:8000` via LLMNR/NetBIOS, Tailscale `*.ts.net`) now get a 400. Only `.local` is documented, so this is probably fine, just not obvious.

## Housekeeping

The commit carries `Co-Authored-By:` rather than `Assisted-by: Claude:<model-id>` (`docs/contributing.md:91`), and the Tested-on / AI-assistance boxes are unticked.

PR #1424, opened 2026-09-16
docs(sdk): stop hardcoding the JS SDK version, resolve latest from npm

FabienDanieau:
## Problem

The JS app docs pinned `@pollen-robotics/reachy-mini-sdk@1.8.0` verbatim in six files. Every npm release makes them stale.

## Solution

Replace the hardcoded version with `npm install @pollen-robotics/reachy-mini-sdk@latest` (which still writes an exact, reproducible pin into `package.json`) and point to the [npm package page](https://www.npmjs.com/package/@pollen-robotics/reachy-mini-sdk) as the source of truth. CDN examples use a `<version>` placeholder. The exact-pin guidance (no `^` ranges) is preserved, and the guide now states that `>= 1.8.0` is the minimum supported version (host shell + protocol v1 contract).

## Also: a small fix on the media side

Second commit, while we were in these files.

Every video reference in the docs pointed app authors at `reachy.attachVideo()`, and `handle.media` appeared in none of them. That call cannot work in an embed: `connectToHost()` completes the WebRTC handshake before the app mounts, so the SDK's one-shot `videoTrack` event has already fired and a listener registered afterwards never runs. The `<video>` element stays black with no error anywhere. `host/src/embed/index.ts` explains this at length, but that is a source comment no app author reads.

Repro: build a host-shell app, call `reachy.attachVideo(el)`, get a black element and no diagnostic.

- `APP_CREATION_GUIDE` §5 listed `ConnectedHandle` without its `media` field, then said "The API is intentionally minimal", so readers concluded the handle had no video API. Added the field plus a short subsection on the race and `handle.media.attachVideo()`.
- `APP_CREATION_GUIDE` §11.6 step 4, `docs/source/SDK/javascript-sdk.md` and `skills/create-js-app.md` all recommended the broken call. Corrected.

`reachy.attachVideo()` stays documented as correct for standalone `mountHost` apps, where it is.

Found while debugging [FabienDanieau/reachy-mini-detector](https://huggingface.co/spaces/FabienDanieau/reachy-mini-detector): with no video frames the detection loop never ran inference at all, so it presented as "the detector doesn't work".


ts/APP_CREATION_GUIDE.md
```diff
@@ -385,6 +390,41 @@ The API is intentionally minimal. If you need a custom channel
...
+
+const handle = await connectToHost();
+const detach = handle.media.attachVideo(videoEl);
+handle.onLeave(() => detach());
+```
+
+There is no equivalent race for the data channel, mute toggles, motor
+commands or state updates: the bridge resolves only once ICE **and**
+the data channel are connected, and state events stream continuously
+at 50 Hz. Keep calling `reachy.setHeadRpyDeg(...)`,
```
darkseed: "At 50 Hz" isn't right for the JS SDK: `state` is a 500 ms `requestState()` poll (`ts/lib/reachy-mini.ts:1863`), or ~30 Hz after `subscribePose()` (`POSE_PUSH_INTERVAL_MS = 33`)  which is what `javascript-sdk.md` and `AGENTS.md` already say. 

The no-race argument still holds, only the number needs fixing:

```suggestion
the data channel are connected, and `state` events keep arriving
(polled every ~500 ms, or ~30 Hz after `subscribePose()`), so a late
listener just picks up the next one. Keep calling `reachy.setHeadRpyDeg(...)`,
```

AGENTS.md
darkseed: Line 173 ("Media flow") isn't in the diff, so commenting at file level: it still says to pass `<video>` to `reachy.attachVideo()`, in a section that opens with "Once `connectToHost()` resolves"  the embed case this PR fixes everywhere else. Should be `handle.media.attachVideo()`.

darkseed:
Tested it with the https://huggingface.co/spaces/FabienDanieau/reachy-mini-detector and it also is able to detect the RM with an orange facemask on.
 
 Docs read well. Two small things as inline comments.

